PR #14, opened 2023-01-03
Attribute Editor Features

Innders:
### Brief Description

Increase the amount of fields that can be edited.

### Description

- All `types` have default fields.

```javascript 
['title', 'description', 'example', 'default', 'regex', 'type']
```
- Each `type` of field has different data fields that can be set.

```javascript
  string: { fields: ['minLength', 'maxLength'] },
  integer: { fields: ['ge', 'gt', 'le', 'lt'] },
  float: { fields: ['ge', 'gt', 'le', 'lt'] },
  list_of_strings: { fields: ['enum', 'minItems', 'maxItems'] },
  boolean: { fields: [] },
```
- When adding a new Attribute changing the `type` will show/hide relevant fields.
- Custom `EmuEditor` component created for `list_of_strings`.
- Bug fix error when multiple attributes selected.

### Screenshots

image
image

### Testing

1. Go to `/settings/attributes`
2. Double click on an attribute
3. Fill out fields and click `Update Attribute`


iLLiCiTiT:
What is `minLength` and maxLength` for?

Innders:
> 

They should be for the number of options you can select from `enum`

iLLiCiTiT:
Are we considering other options then "singleselection" (1 min and 1 max) and "multiselection" (0 min and len(enum) max)?

martastain:
This actually comes from pydantic - it support this attributes so I thought it is a good idea just to propagate them. For single selection, we have `type=string` with `enum` set, for multiselections we have `type=list_of_strings`. you can set minItems to 1 to essentially make at least one item required.

martastain:
`list_of_strings` with `minItems == maxItems == 1` would still be a list with exactly one item.

iLLiCiTiT:
Ok "sugar" to make it work :)

PR #22, opened 2023-01-12
Op 4734 project roots overrides

martastain:
Adds a project manager tab for setting user/machine-based overrides of project roots.


jakubjezek001:
Task linked: [OP-4734 Project roots overrides per user/machine](https://app.clickup.com/t/6658547/OP-4734)

PR #24, opened 2023-01-16
Feature: Search Filter Tables

Innders:
### Brief Description

Search filter rows in `Browser`, `Editor`, `Users`, `Attributes`.

### `Browser`, `Users`, `Attributes`

- Search looks for matches across all fields.
- You can filter down the results further using a second search separated by a comma `,`.
- The search "app, render, light" would need to have a field that matched each one of those keywords to be matched.

subsets_filter_search_v003

### `Editor` (OLD V001)

- When searching a flattened view of the hierarchy is shown.
- When searching finishes the hierarchy view is shown again, except filtered down to the search.
- No multiple comma searches for this yet.

### `Editor` Issues

- Could be better optimised as big arrays (9000+) can be generated for large projects.
- Data fetching could be done in RTK Query.
- Weird bug that requires ` const context = useSelector((state) => ({ ...state.context }))` to be used in order for `Editor` field to work correctly.

OLD V001 Screenshot
filter_editor_v002

### Testing

1. Find an asset field to search for.
2. Trying typing in that value into the search bar.
3. The results should filter and show matching results.


jakubjezek001:
Task linked: [OP-4725 search filter in tables](https://app.clickup.com/t/6658547/OP-4725)

Innders:
# (LATEST V002)

### Brief Description

Innders:
### Editor V002

- Complete redesign and rewrite.
- Uses a search with drop down suggestions.
- New `SearchDropdown` component.
- Height animations on the drop down.
- Search filtering only occurs `onSubmit` to improve performance.
- Can show a set number of suggestions (default 5).
- Suggestions are now more useful and match the search better.
- You can now select on a specific suggestion.
- Full keyboard support, `ArrowUp`/`ArrowDown` to select suggestions, `Enter` to submit and `Escape` to close.

filter_editor_v004

PR #31, opened 2023-01-24
Enhancement: User Page Overview

Innders:
### Brief Description

When no user is a selected a panel with helpful stats and shortcut buttons.

### Description

- Show total users, admins, services, managers.
- Clicking between projects changes these values based on the number of users in assigned to the project.
- Clicking on **Total** selects all users shown in list.
- Clicking **Admin, Service, Manager** prefills the search bar with the user type to filter through the list and only show that type.
- Search filter can now do inverse searches and remove results that match the search with adding `!` to the beginning. For example `!admin` will show everything but matches for "admin". 
- Clicking **Users** does the opposite by filling the search bar with `!admin, !service, !manager`. This shows all users that aren't one of those.
- Show last updated user on the project with a quick link to show the user details.
- Show **me** user with quick link.
- Show buttons to **Add New User** and **Add New User Attribute**.

users_overview_v001_small

### DIFF
--- a/src/hooks/useSearchFilter.js
+++ b/src/hooks/useSearchFilter.js
@@ -40,11 +40,26 @@ const useSearchFilter = (fields, data) => {
     if (searchArray.length && dataWithKeywords) {
       return dataWithKeywords.filter((user) => {
         const matchingKeys = []
-        user.keywords?.some((key) =>
+        const inverseMatchingKeys = []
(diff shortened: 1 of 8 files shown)


jakubjezek001:
Task linked: [OP-4730 improve users page](https://app.clickup.com/t/6658547/OP-4730)

PR #32, opened 2023-01-24
Fix Reformat User Detail Data Infinite Loop

Innders:
### Description

Fixes an issue where the users page could go into an infinite loop and crash the browser.


jakubjezek001:
Task linked: [OP-4730 improve users page](https://app.clickup.com/t/6658547/OP-4730)

PR #58, opened 2023-03-03
fix: merge project name into context

Innders:
### Brief Description

Access to project name was lost in Addon when context was split out into project and context.

Found by @Minkiu


Minkiu (approved):

martastain (commented):
Context from useSelecto is immutable, so I've merged it with the project name later, in postMessage. Now it works with the example addon. We should consider, what other data from project might need - but I guess addon could request the data itself when needed - with the project name and access token, it can access both the active project and active user via rest.

PR #59, opened 2023-03-05
Enhancement/op 4976 new entity dialog

Innders:
### Description

Fix lots of bugs and tweaks to feedback.

When creating a new folder or task, a dialog will appear to quickly input the name and type.

In the future this dialog can be build out further to support more advanced creation.


Innders:
### Bugs

1. Error on deleting folder still deletes folder in UI (requires refresh).
2. Creating root folder doesn't let you add more folders (after commit) unless you refresh.

PR #63, opened 2023-03-15
Fix: disable event viewer cache

Innders:
- Event viewer always fetches new data on mount.
- Fixes and issue where old caches were being used.
- Moved pubsub to RTK Query


src/services/events/getEvents.js
```diff
@@ -1,4 +1,5 @@
 import { ayonApi } from '../ayon'
+import PubSub from '/src/pubsub'
```
Minkiu, line 2: Wouldn't here `import PubSub from '../pubsub'` work too?

PR #66, opened 2023-03-22
Attribute Fields Enhancements

Innders:
### Description

- Enum dropdown editor
- Use title for field name
- Use date formatting
- Description shows as a tooltip when hovering over field name

image


Innders:
The extra fields on version causes representations to go off screen.

image

An idea is to move subset / version / representations details panel to a tab view making sure height is always 100% with overflowing scrolling.

Innders:
After speaking with Milan the representations has stayed stacked with version. But a tabs approach of some sort could be explored in the future.

Changes that were made to folder/task/version details.

- Rewrite to use **one** component for all entity types.
- Use `EntityDetailsHeader` for title and icon. Keeps the UI consistent across to Editor page.
- Enable multiple selection of entities.
- Better handling of overflowing scroll.

PR #105, opened 2023-06-02
Fix: Remove Root Folder

Innders:
### Description

- Having two "add folders" button was unnecessary and confusing.
- Now you can create "root" folders by using the "add folder" button when nothing is selected.
- Or by right clicking in the table (but not on a specific table row).
- Ability to deselect rows.

root_folder_context_v001


martastain:
works like a charm

PR #106, opened 2023-06-03
Redesign Manage Projects Buttons

Innders:
### Description

This is a bit of an experiment but I think it's an improvement.

- Open, Delete buttons have been completely removed.
- Other actions have been added to the context menu.
- This has also been done for the teams page.
- If these changes are made, we might need to rethink the button layout in settings as well.

Only problem might be tablet (touch screen) users not being able to reach "delete project". But then maybe deleting a project shouldn't be done on a touch screen device.

image
image
image


Innders:
Also moved the "Manage Projects" button back to left hand side projects menu and removed from the right hand side user menu.

1. I kept going to the projects menu to look for the button
2. I quite like the idea of having "left is for project scope" "right is for global scope".

PR #109, opened 2023-06-15
fix: filter by folder type maintains hierarchy

Innders:
This actually wasn't a bug and was the intended behaviour but I can see it would be very confusing to the user.

I have changed the behaviour to maintain the hierarchical structure (not flattened)


jakubjezek001:
Task linked: [OP-6170 Hierarchy Filtering Duplicates Folders](https://app.clickup.com/t/6658547/OP-6170)

PR #110, opened 2023-06-15
fix: Show Columns matches column labels

Innders:
image


jakubjezek001:
Task linked: [OP-6167 Show Columns Filter Correct Names](https://app.clickup.com/t/6658547/OP-6167)

PR #111, opened 2023-06-15
fix: workfile detail path use overflow field

Innders:
image


jakubjezek001:
Task linked: [OP-6165 Workfile name cutoff](https://app.clickup.com/t/6658547/OP-6165)

PR #122, opened 2023-06-25
fix: make editor changes more obvious

Innders:
image


Innders:
close merged into #122

PR #124, opened 2023-06-26
Fix: duplicating teams design fix

Innders:
Based on user research from @LiborBatek and @BigRoy, thank you!

Duplicating teams now shows a dialog for naming your newly duplicated team. 

duplicate_team_dialog


BigRoy:
For completeness sake, can the dialog show which Team you are duplicating? (e.g. a header "Duplicate team X")
And can we uppercase the field's name on the left to "Name" or maybe even better as "Team name"?

Innders:
Because... I realised those exact same points you just made but forgot to update the gif. XD 

This is how the it looks since b3052c6 

image

LiborBatek:
Nice! Seems much more clear and better UX! Also agree with @BigRoy proposals....they make perfect sense!

Thx for including me during testing...it was fun!

PR #171, opened 2023-08-24
Fix/onboaring fixes

Innders:
- fix: toast error always coming up.
- fix: release select overflow, it now scrolls.
- fix: remove addons from release expansion as it was it was too overwhelming.
- fix: closed release bio overflow now truncates text.
- fix: timestamp now says "Today" if it was released "hours" ago.

### DIFF
--- a/src/components/Release/ReleasePreset.jsx
+++ b/src/components/Release/ReleasePreset.jsx
@@ -5,7 +5,6 @@ import { format, formatDistanceToNow } from 'date-fns'
 import Type from '/src/theme/typography.module.css'
 
 const ReleasePreset = ({
-  addons,
   label,
   bio,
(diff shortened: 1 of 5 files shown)


src/components/Release/ReleasePreset.jsx
```diff
@@ -55,16 +55,16 @@ const ReleasePreset = ({
         <Icon icon={icon} />
         <div>
```
Minkiu, line 56: I know this is merged, it's out of curiosity, is this `<div>` really necessary here?

Innders: Yes, there's CSS that targets that div to create a flexbox 😊 probably should have labelled it better though.

PR #180, opened 2023-09-02
Fix: breadcrumbs show scope name

Innders:
- Fix bug where project name wasn't getting set in uri
- Now show page name as a default
- Settings page shows "Studio Settings"
- Manage Projects page shows "Projects Manager"

- Make font size bigger and give secondary blue background to make breadcrumbs more prominent
- Fix auto selection bug not always selecting whole text

Thanks @m-u-r-p-h-y for the feedback

image

### DIFF
--- a/src/containers/Breadcrumbs/Breadcrumbs.jsx
+++ b/src/containers/Breadcrumbs/Breadcrumbs.jsx
@@ -1,7 +1,7 @@
 import axios from 'axios'
 import { useEffect, useState, useRef } from 'react'
 import { useSelector, useDispatch } from 'react-redux'
-import { useNavigate } from 'react-router-dom'
+import { useLocation, useNavigate } from 'react-router-dom'
 import { toast } from 'react-toastify'
 import { InputText } from '@ynput/ayon-react-components'
 import * as Styled from './Breadcrumbs.styled'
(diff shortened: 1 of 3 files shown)


src/containers/Breadcrumbs/Breadcrumbs.jsx
```diff
@@ -17,14 +17,35 @@ import {
...
-
+const uri2crumbs = (uri = '', pathname) => {
   // parse uri to path and query params
-
-  const [path, query] = uri.split('://')[1].split('?')
+  const [scope, pathAndQuery = ''] = uri.split('://')
+  const [path, query] = pathAndQuery.split('?')
   const crumbs = path.split('/').filter((crumb) => crumb)
+
+  if (scope?.includes('ayon+settings')) {
```
Minkiu, line 28: I guess this could also be a switch?

```js
switch (scope) {
  case "ayon+settings":
    let settingsScope = query?.includes('project') ? 'Projects Manager' : 'Studio Settings'
    crumbs.unshift(settingsScope)
    break;
  case "ayon+entity":
    crumbs.unshift("Project")
    break;
  default:
    let pageTitle = pathname.split("/")[1]
    if (pageTitle.includes('settings')) pageTitle = 'Studio Settings'
    else if (pageTitle.includes('manageProjects')) pageTitle = 'Projects Manager'
    // just a regular url
    crumbs.unshift(upperFirst(pageTitle));
}
```

Regardless,  `pathname.split("/")[1]` might fail if `pathname` contains no slashes (unsure if possible)

PR #184, opened 2023-09-06
Fix: propagating new complex items of arrays in settings

### LINKED ISSUE #185: server: deep nested settings resolving
At the moment we are not able to save settings in case those are list with deep nested attributes. 

Here is the example of what will saving result into:

Image

And here is the example of settings where the issue is occuring:

Image

martastain:
At the moment we are not able to save settings in case those are list with deep nested attributes. 

Here is the example of what will saving result into:

Image

And here is the example of settings where the issue is occuring:

Image

closes #185


jakubjezek001 (approved):
testing was successfull

PR #187, opened 2023-09-11
Fix: Community Feedback Bug

Innders:
- Upgrade a-r-c to fix tags select crash
- header font size 16px to 14px
- editor columns width default 140
- workfiles showing old items when switching

Thanks @BigRoy


BigRoy:
⚡ Lightning quick fixes here, thanks!

PR #200, opened 2023-09-22
OP-6657 show warning when name is wrong in create project dialog

martastain:
Project creation dialog: draw a red outline when project name or project code is wrong. Display actual error message as a tooltip.

image

Red outline is var(--color-hl-error) which should be used there, but it is not a very nice color. We should find a better one and include it to ayon-react-components.


martastain:
Updated: use '.error' instead of explicit style. I tried to use `invalid` property, which is styled in ARC, but it does not propagate.

PR #222, opened 2023-10-13
Artist Dashboard: No projects button opens project dialog

Innders:
## Changelog Description

- When there are no projects the dashboard will show a button that opens the new project dialog form.

## Testing

1. Have no projects or inside `UserDashboardPage:57` set if state to always `true`.
2. Click "Create first project" button.
3. New project dialog should open.
4. Creating new project should direct user to newly created project in project settings.


martastain (approved):
Works

PR #232, opened 2023-11-02
Use REST endpoint instead of GraphQL to load entity detail

martastain:
Revert the original behavior of entity detail dialog to use useGetEntityQuery instead of GraphQL.

Reason: This is the only place in the frontend, where a user can see the actual payload as available in REST, which is useful for debugging. It contains a complete entity payload as stored in the database along with its dynamic attributes.

Drawback: Only one entity is loaded at time, but i don't believe that is a major problem


Innders (approved):
👍

PR #236, opened 2023-11-10
Thumbnails: upload thumbnails from Browser, Editor and Dashboard

Innders:
## Changelog Description

- Click to upload or drag and drop, to add thumbnails to folders, tasks and versions.
- Upload multiple images: if multiple entities are selected and multiple files dropped, matching file to entity names will be uploaded. For example `sh010.png -> sh010, shot020.png -> sh020`.

## Additional Information

thumbnail_upload_kanban_v001
thumbnail_upload_browser_v001

#### Uploading multiple images to multiple folders
thumbnail_upload_browser_multiple_v001

### DIFF
--- a/src/components/Details/EntityDetailsHeader.jsx
+++ b/src/components/Details/EntityDetailsHeader.jsx
@@ -84,6 +84,7 @@ const EntityDetailsHeader = ({
           thumbnails={thumbnails}
           isLoading={isLoading}
           onUpload={onThumbnailUpload}
+          portalId={'editor-entity-details-container'}
         />
       )}
       {isLoading ? (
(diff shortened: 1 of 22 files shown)


mkolar:
Looking good for a quick first take. I have a few comments that should probably be addresses now even though it si just a PoC.

1. The drop are should be much smaller I think. it is really nice to drop to a full sidebar, but we should reserve that for something else. Probably attachments for comments later on. I'd limit thumbnails to either just the thumbnail itself, or possible to the upper section only. 
image
image


2. Clicking the full image to replace seems a bit too dangerous considering most thumbnails are published through different means. I'd add a hover button to upload and remove thumbnail instead.

Functionally though, it's great. The bulk works as expected too

martastain:
We should merge this at the same time we merge https://github.com/ynput/ayon-backend/pull/66 , but since that modifies the DB schema, it is probably a good idea to do it for Ayon 0.6

Innders:
Fixes made.

thumbnail_upload_browser_v002

PR #241, opened 2023-11-27
Auto-expand project sidebar if project name is too long

martastain:


martastain:
It is still work in progress - hovering a project, which hides project code and displays pin/settings icons instead changes the width, project code and icons should probably be in a container with a fixed width.

jakubjezek001 (approved):
LGTM! ;) thanks! 
ayon

PR #253, opened 2023-12-14
Editor: New Entity Independent Label and Name Editing

Innders:
## Changelog Description

- Enhancement: Independent editing of `label` and `name` for new uncommitted entities.
- Enhancement: Add shortcut tooltip to new entity buttons.
- Fix: Incorrect breadcrumbs for new entities.
- Fix: Path clipboard copying incorrectly included spaces between slashes.

## Additional Information

image
image

## Testing

1. New entity dialog: hovering over the "Add" and "Add and Close" buttons should display shortcut.
2. Create new folder/task: selecting task should update URI (breadcrumbs) correctly.
new folder path
3. The `name` and `label` should be independently editable before committing.

### DIFF
--- a/src/components/Details/EntityDetailsHeader.jsx
+++ b/src/components/Details/EntityDetailsHeader.jsx
@@ -42,13 +42,15 @@ const EntityDetailsHeader = ({
 
   const isMultiple = values.length > 1
 
-  let subTitle = ''
+  let subTitle = '',
+    breadcrumbs = []
   if (isMultiple) {
     subTitle = values.map((v) => v?.name).join(', ')
   } else if (uri) {
     const [path, qs] = uri.split('://')[1].split('?')
     //eslint-disable-next-line
-    const [_, ...breadcrumbs] = path.split('/').filter((p) => p)
+    const [_, ...bc] = path.split('/').filter((p) => p)
+    breadcrumbs = bc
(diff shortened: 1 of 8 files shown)


BigRoy:
> New entity dialog: hovering over the "Add" and "Add and Close" buttons should display shortcut.

Out of curiosity:

Regarding this - I believe the Add and Close shows "CTRL+ENTER" but clicking regular enter also 'presses' that button for me. I wonder whether the shortcut CTRL+ENTER then still makes sense?

Whilst editing the text I was unable to have regular enter trigger the other option - it always picked the Add & Close.

BigRoy:
> Fix: Incorrect breadcrumbs for new entities.

In my tests the issue here wasn't only for new assets, but just selecting one folder, and _then a sibling of that folder._ it wouldn't update.

It works fine in Project Browser, but incorrectly in Project Editor.

Innders:
> I wonder whether the shortcut CTRL+ENTER then still makes sense?

CTRL+ENTER makes more sense when using the add folder sequence tool. Hitting regular enter will only submit the form if the focus is on the very last input but CTRL+ENTER will allow you submit the form from wherever.

Regular enter is preserved in this way as it is the generally expected behaviour. 

Screenshot 2023-12-14 at 10 02 35

PR #255, opened 2023-12-18
Addons: Delete Addons MVP

Innders:
## Changelog description

- See which addons are in which bundles.
- Archive bundles in order to delete addons.
- Delete addons.

## Additional Information

https://github.com/ynput/ayon-frontend/assets/49156310/07172597-1336-4b17-b540-01390436779d

## Todo

- Delete dep packages and installers


LiborBatek:
Looks great and super useful!

Just curious about the top header labels...are those actionable btns or just headers? If just headers I propose to name em differently...so they communicate whats present in the list and why.

image

Not sure but for ie instead of ```Delete Addons``` something like ```Installed Addons``` and for the other column instead of

```Archive Bundles``` something like ```Bundle / In Use```

You get the idea.

martastain:
Merging. We could iterate & fix cosmetics later.

LiborBatek:
> Merging. We could iterate & fix cosmetics later.

True...but usually never happens :D

PR #276, opened 2024-03-07
resolve paths in workfiles page

martastain:


src/services/customRoots.js
```diff
@@ -1,7 +1,25 @@
...
+
 const customRoots = ayonApi.injectEndpoints({
   endpoints: (build) => ({
+    getSiteRoots: build.query({
+      query: ({ projectName, siteId, platform }) => ({
+        url: `/api/projects/${projectName}/siteRoots${getSiteRootsQuery(siteId, platform)}`,
+        method: 'GET',
+      }),
+
+      providesTags: ['siteRoots'],
```
Innders, line 18: `siteRoots` needs to be added to `src/services/ayon.js` tagsTypes list

Innders (changes requested):
Add siteRoots to tagTypes

src/pages/WorkfilesPage/WorkfileDetail.jsx
```diff
@@ -2,12 +2,42 @@ import { Section, Panel } from '@ynput/ayon-react-components'
...
+    if (replacements.hasOwnProperty(p1)) {
+      return replacements[p1]
+    }
+    return match
+  })
+}
 
 const WorkfileDetail = ({ workfileId, style }) => {
   const projectName = useSelector((state) => state.project.name)
+  const [selectedSite, setSelectedSite] = useState(null)
```
Innders, line 40: Could we use useLocalStorage here so the user doesn't have to set it everytime? In the future we could use redux global state if siteId is useful to know in other parts of the app.

src/services/customRoots.js
Innders, line 18: siteRoots needs to be added to src/services/ayon.js tagsTypes list

Innders (changes requested):

PR #281, opened 2024-03-08
enhancement: deactivate project from context menu

Innders:
## Changelog Description

- Activate and deactivate projects from the context menu when on the projects manager page.
- Only show "Delete Project" once a project is inactive.

activate_project_from_context_v001


BigRoy:
I wonder - would it be possible to "show the Delete Project" button but have it be deactivated when a project is active, and have e.g. a tooltip state that you need to deactivate a project first to enable the delete project button? I just wonder if anyone will just "know" they need to first deactivate the project or understand why it might show for some and not for other projects?)

PR #286, opened 2024-03-12
Force delete folders with children

Innders:
## Changelog Description

- When deleting a folder with dependencies show a confirmation dialog to force save.

image

## Testing

1. Delete a folder that has children.
2. Change and delete some other folders without children.
3. Commit changes.
4. Accept "Force Delete" confirmation.
5. Start from 1 again but reject confirmation.

### DIFF
--- a/src/pages/EditorPage/EditorPage.jsx
+++ b/src/pages/EditorPage/EditorPage.jsx
@@ -876,15 +876,10 @@ const EditorPage = () => {
               deleteLabel: 'Force Delete',
               showToasts: false,
               accept: async () => {
-                // filter out the 409s
-                const failed = res.operations.filter((op) => op.status === 409).map((op) => op.id)
-                // filter out updates that aren't in failed array with correct id
-                const forcedUpdates = updates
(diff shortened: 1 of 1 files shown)


BigRoy:
Question - is the Force Delete 'autoselected' or the cancel button? With something as 'dangerous' as this I'd expect "cancel" to be auto-selected (e.g pressing enter)

Also, in OpenPype old project manager when deleting e.g. the project you had to manually type the project name or DELETE or whatever to confirm you actually really really want to delete so that you don't accidentally remove it by clicking things too quickly. Should we consider cases like that?

Additionally, from the wording it is a bit unclear whether this will remove it ONLY from the backend or will actually also physically start removing files from disk. I suspect it's the former - but someone might expect the latter from the wording.

Innders:
> Additionally, from the wording it is a bit unclear whether this will remove it ONLY from the backend or will actually also physically start removing files from disk. I suspect it's the former - but someone might expect the latter from the wording.

@martastain I'm actually not sure on this?

martastain:
The server does not touch files at all. Yet.

PR #287, opened 2024-03-12
Tools Field Dropdown

Innders:
## Changelog Description

- Tools field in editor now shows a readonly dropdown for better readability.
- It also shows the number of tools at the start in brackets.

editor_tools_dropdown_v001


BigRoy:
Ah, yes - so much better. Thank you! You're like a 🌪️ **and** on 🔥.

PR #292, opened 2024-03-15
Projects menu improvements

Innders:
## Changelog Description

- Enhancement: Full keyboard support using arrows.
- Enhancement: Start typing straight after opening the menu to start searching, there's no need to click the search icon.

- Fix: Search state now clears on close.
- Fix: Block other shortcuts when open, like `s+s`.
- Fix: Ensure the project is always focused when opening. Sometimes it would get stuck on other items.
- Fix: You can now pin **all** projects, if you are a mad man. _[when you pin all the projects](https://www.youtube.com/watch?v=danzsQTDTx4&t=72s)_

projects_menu_keyboard_v001

## Notes

I have been thinking about the shortcut to open the projects menu... `1` isn't exactly obvious and maybe something like `m` would be better? This would also have the advantage of freeing up the numbers to be used to select the projects quicker.

You can press `m` then `1` or `2` or `3` etc to pick one of your top pinned projects, which would make project switching extremely quick (if you learnt the combo).

### DIFF
--- a/src/components/Menu/MenuComponents/MenuList.jsx
+++ b/src/components/Menu/MenuComponents/MenuList.jsx
@@ -74,7 +74,7 @@ const MenuList = ({
         {items.map((item, i) => {
           // if item is a node, return it
           if (item.node) {
-            return <div key={i}>{item.node}</div>
+            return item.node
           }
 
           if (item?.id === 'divider') return <hr key={i} />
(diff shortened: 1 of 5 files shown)


martastain:
I love it! I think `1` or maybe `~` is a good choice. It is a key you can build a muscle memory for.

BigRoy:
Thank you very much! Loving the press `1`, type to search, enter to access the project quickly.

martastain:
BTW: would it be possible to use (not now) physical keys mapping instead of characters? On czech keyboard `1` is `+` and `~` is `;` .... don't ask me why. i don't use czech keyboard, but i guess other layouts may also differ - on french keyboard `1` is `&` IIRC.

PR #297, opened 2024-03-20
Assignee select mixed not working temp fix

Innders:
## Changelog Description

- Editor task `AssigneeSelect` would prevent you from adding new users when it was `mixed`.
- This was caused by the shortcuts hook re-rendering the whole page every time the mouse moved.

## Additional Information

`/projects/{project}/editor`

Temporary solution was to move the shortcuts into an empty component to prevent the whole editor page component from re-rendering.

I will need to look into the `useShortcut` hook more to optimise this better.

Maybe we can have a `<Shortcut />` comp that just wraps the `useShortcut` hook.

### DIFF
--- a/src/pages/EditorPage/EditorPanel.jsx
+++ b/src/pages/EditorPage/EditorPanel.jsx
@@ -196,6 +196,8 @@ const EditorPanel = ({
       },
     }
 
+    console.log('create form')
+
     const type = nodes[nodeIds[0]]?.data?.__entityType
 
     if (type) {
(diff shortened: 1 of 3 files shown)


src/pages/EditorPage/EditorPanel.jsx
```diff
@@ -196,6 +196,8 @@ const EditorPanel = ({
       },
     }
 
+    console.log('create form')
+
```
BigRoy, line 200: Likely this was purely for debugging?

PR #299, opened 2024-03-21
Enhancement/ay 3883 add tags to editor panel

LudoHolbik:
Added Tags input into the Editor Panel. Mixed values are shown when selecting multiple items with different tags (as is expected). 

Minor fix needs to be done in ayon-react-components if we also want to show text information about tags being "mixed" (just like in Type or Assignees field). Separate ticked was created for this issue.

image

### DIFF
--- a/src/pages/EditorPage/EditorPanel.jsx
+++ b/src/pages/EditorPage/EditorPanel.jsx
@@ -641,6 +641,7 @@ const EditorPanel = ({
                       onChange={(v) => handleLocalChange(v, changeKey, field)}
                       align="right"
                       width={200}
+                      buttonStyle={{ border: '1px solid var(--md-sys-color-outline-variant)'}}
                       isChanged={isChanged}
                     />
                       )
--- a/src/pages/EditorPage/EditorPanel.jsx
+++ b/src/pages/EditorPage/EditorPanel.jsx
@@ -187,7 +187,7 @@ const EditorPanel = ({
         changeKey: '_status',
         label: 'Status',
         field: 'status',
-        placeholder: `Mixed (${statusValues.isMultiple && statusValues.isMultiple.join(', ')})`,
+        placeholder: `Mixed (${statusValues.multipleValues && statusValues.multipleValues.join(', ')})`,
         ...statusValues,
       },
       _assignees: {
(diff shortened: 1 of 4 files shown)


src/pages/EditorPage/EditorPanel.jsx
```diff
@@ -617,6 +627,20 @@ const EditorPanel = ({
...
+                      input = (
+                    <TagsSelect
+                      value={isMultiple ? union(...isMultiple) : value || []}
+                      tags={projectTagsObject}
+                      tagsOrder={projectTagsOrder}
+                      isMultiple={!!isMultiple}
+                      onChange={(v) => handleLocalChange(v, changeKey, field)}
+                      align="right"
+                      styleDropdown={{ overflow: 'hidden' }}
+                      width={200}
```
Innders, line 640: Can we please add a border like the rest of the fields.

```
style={{
  border: isChanged
    ? '3px solid var(--md-sys-color-primary)'
    : '1px solid var(--md-sys-color-outline-variant)',
}}
```

src/pages/EditorPage/EditorPanel.jsx
```diff
@@ -617,6 +627,20 @@ const EditorPanel = ({
                       widthExpand
                     />
                   )
+                    } else if (field === 'tags') {
+                      input = (
+                    <TagsSelect
+                      value={isMultiple ? union(...isMultiple) : value || []}
```
Innders, line 633: Lets go ahead and change this confusing variable name to something else like `mixedValue`.

Innders (changes requested):

src/pages/EditorPage/EditorPanel.jsx
Innders: Sorry this piece of code seems to be wrong, my bad. The border shouldn't be changing here but the background should be going blue.

This is broken for status select and tags select. Look at how the **Type** select dropdown background goes to blue `var(--md-sys-color-primary)` and the inside text goes to dark blue `var(--md-sys-color-on-primary)`. 

image

We need to do the same for both **Status** and **Tags**. **Tags**

image

There is a chance there is something broken with the ARC component but lets just use `style` for now and fix ARC later.

Innders (changes requested):

src/pages/EditorPage/EditorPanel.jsx
LudoHolbik: I have changed styles as you suggested.

After looking into ayon-react-components I thought that It would be better to put styles into `buttonStyle` property instead of `style` + I added the isChanged property to make it work.

src/pages/EditorPage/EditorPanel.jsx
LudoHolbik: I changed `isMultiple` to `multipleValues`, I was thinking about `mixedValues` but technically they are not mixed, it is the union function that mixes them.

LudoHolbik:
I made changes to Tags and Status based on what we discussed and how **Type** looks like, this is current visual of what happens when you change them:

image

Innders (changes requested):
changed bg colours work well. small fix the tags have lost their default border.

image

LudoHolbik:
I have added the default border without dynamic style.

Innders (approved):

Innders:
Last thing, can you please add ## Changelog Description to the top of the PR and add the "enhancement" label please.

This helps with the auto generated changelog.

Screenshot 2024-03-25 at 12 14 24

PR #301, opened 2024-03-25
Fix/ay 4015 fix status select chevron

LudoHolbik:
I have made some smaller adjustments for Status field.

1. Chevron spacing is changed (Text is on left side, chevron on very right) in Both editor and Browser section.
2. If multiple statuses are selected, they are highlighted in dropdown (but no chevron is shown in dropdown menu in this case).

image

image

image


Innders (approved):
Looks good to me!

Just one small thing... could you please update the PR description to include ## Changelog Description at the top and add the "bug" tag.

Screenshot 2024-03-25 at 12 14 24

PR #302, opened 2024-03-26
Use project/folders instead of hierarchy in the hierarchy component

martastain:


src/containers/hierarchy.jsx
```diff
@@ -74,6 +76,44 @@ const filterHierarchy = (text, folder, folders) => {
...
+            icon={folders[child.folderType]?.icon}
+            text={child.label}
+            name={child.name}
+          />
+        ),
+      }
+     
+    }
+    if (child.id in parents) {
+      nchild.children = buildTree(parents, child.id, folders)
```
Innders, line 106: We will probably want to optimise the build tree. Calling the same function inside of itself can get super slow. There are some other tricks we can do to speed it up!

martastain:
The backend part is now available in ayon:dev

martastain:
I've fixed the issue with failing request during the first load. I think this could be merged at any point if that's useful.

Innders:
> I've fixed the issue with failing request during the first load. I think this could be merged at any point if that's useful.

I agree. The FE isn't ready yet because filtering by folder type is tricky and the editor page is just a beast. Plus also it works pretty well at the moment and I don't want to break anything right now.

When the editor gets a rewrite /folders will be used!

Issue #303, opened 2024-03-27
Crashing the event Viewer


stephenscollay:
When syncing from Shotgrid to Ayon I have noticed that it is possible to max out the memory of the Event Viewer Page which requires a reload to fix.

![Capture](https://github.com/ynput/ayon-frontend/assets/50131715/64b08ce7-9518-4b63-ad41-0749647eb135)

Innders:
Interesting, are you syncing a large amount of data?

stephenscollay:
I should have put this above sorry but this is using the Shotgrid > Ayon sync button after having removed the shots in Ayon. I would not say a large amount, 4 shots each with 4 tasks. I'll be honest, I am not sure what the limitation is but it has occurred twice so I thought I may as well note it.

PR #305, opened 2024-03-27
fix: throttle websocket overload

Innders:
## Changelog Description

If lots of changes were made at a different site then the websocket messages could overload the client.

- Fix: When there are too many messages to handle they will be ignored and you will be prompted to reload the page.
- Hopefully it solves: #303 

image

## Testing

1. Open events viewer page.
2. In a new browser, make loads of changes, like creating 1000 folders.
3. The events viewing will probably show the first 10ish and then the toast should appear.

## Additional Information

We might need to play with the threshold amount, it's currently pretty low, about 15 changes in one go.

### DIFF
--- a/src/context/websocketContext.jsx
+++ b/src/context/websocketContext.jsx
@@ -88,11 +88,12 @@ export const SocketProvider = (props) => {
       }
 
       const data = JSON.parse(message.data)
-      if (data.topic === 'heartbeat') return
+      const { topic, sender, summary } = data || {}
+      if (topic === 'heartbeat') return
 
-      if (data.topic === 'server.restart_requested') setServerRestartingVisible(true)
+      if (topic === 'server.restart_requested') setServerRestartingVisible(true)
 
-      if (data.sender === window.senderId) {
+      if (sender === window.senderId) {
         return // my own message. ignore
       }
 
@@ -105,10 +106,10 @@ export const SocketProvider = (props) => {
 
       lastCall = now
 
-      if (data.topic === 'shout' && data?.summary?.text) toast.info(data.summary.text)
+      if (topic === 'shout' && data?.summary?.text) toast.info(summary.text)
 
       console.log('Event RX', data)
-      PubSub.publish(data.topic, data)
+      PubSub.publish(topic, data)
     }
   })()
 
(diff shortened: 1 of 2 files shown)


src/context/websocketContext.jsx
```diff
@@ -54,20 +55,62 @@ export const SocketProvider = (props) => {
...
+        callCount += 1
+      } else {
+        callCount = 0
+      }
+
+      lastCall = now
+
+      if (data.topic === 'shout' && data?.summary?.text) toast.info(data.summary.text)
+
+      console.log('Event RX', data)
```
LudoHolbik, line 110: Should console.log stay here ?

martastain: I use this Event RX everytime i debug events :)

Innders: Historically yes, makes it a lot easier for us to debug. Maybe at some point we can remove it. Probably a bigger discussion if any logs should be visible when built.

src/context/websocketContext.jsx
```diff
@@ -54,20 +55,62 @@ export const SocketProvider = (props) => {
...
+    let lastCall = Date.now()
+
+    return (message) => {
+      // If the function is called more than 10 times per second, return early.
+      if (callCount > 100) {
+        setOverloaded(true)
+        return console.log('WS OVERLOAD!!!!')
+      }
+
+      const data = JSON.parse(message.data)
```
LudoHolbik, line 90: What do you think about destructuring ? Say you would write something like 
`const { topic, sender, summary } = data;`

So you do not have to extract it from data everywhere. You would get something like 

```
  if (topic === 'heartbeat') return;
  if (topic === 'server.restart_requested') {
    setServerRestartingVisible(true);
    return;
  }
  if (sender === window.senderId) return;
```
Not pushing, just asking.

src/context/websocketContext.jsx
```diff
@@ -54,20 +55,63 @@ export const SocketProvider = (props) => {
...
+  // it is a closure that keeps track of the number of calls and the last call time
+  let onMessage = (() => {
+    let callCount = 0
+    let lastCall = Date.now()
+
+    return (message) => {
+      // If the function is called more than 10 times per second, return early.
+      if (callCount > 100) {
+        setOverloaded(true)
+        return console.log('WS OVERLOAD!!!!')
```
BigRoy, line 87: Should a log like this be more informational? Like e.g. `Overload: Websocket messages exceeded 100 calls.`

LudoHolbik (approved):

martastain (approved):

PR #308, opened 2024-03-28
Save context when refreshing page

Innders:
## Changelog Description

- When the user refreshes the page the current state of the app should remain.
- Only works on editor, browser and workfiles.
- One negative is it makes the context slice extremely complicated, but it's very easy to now add states that sync with local storage.
- Fixes lots of issues with version select and improves how it fetches and caches version data.
- Adds a context menu on hierarchy that allows you to reset all products to latest version.

https://github.com/ynput/ayon-frontend/assets/49156310/14bc12ab-8de0-4e1a-8939-cd7040c1dee0

## Testing

1. Open some folders, select loads of stuff.
2. Refresh the page and see if everything's the same.


BigRoy:
Although I do like this - what is now the way to e.g. really "reset" my view so that e.g. my version selection, etc. get lost on purpose? Like, now it persists - but what if I don't want that. Which is quite the usual case if someone maybe intentionally refreshes?

I also wonder how well this "preserving" works if after refresh maybe certain folders or statuses were removed or changed by someone else. Does it still work? or does it error over the missing entries?

Innders:
> what is now the way to e.g. really "reset" my view 

I think if a user is refreshing their page to "reset" or "fix" something then we haven't done a good enough job of providing the right navigation/tools/UX.

>  but what if I don't want that. Which is quite the usual case if someone maybe intentionally refreshes?

I guess we will find out during testing what should and shouldn't be persisted. This has gone to the extreme, but now the infrastructure is in place to dial it back or expanded it in different areas.

> I also wonder how well this "preserving" works if after refresh maybe certain folders or statuses were removed or changed by someone else. Does it still work? or does it error over the missing entries?

If you are viewing something that is deleted by somebody else then real time updates should show that, you shouldn't have to refresh to see that. Selections of deleted items might be left over but it shouldn't error and will be cleared once the user clicks anything else.

BigRoy:
> I think if a user is refreshing their page to "reset" or "fix" something then we haven't done a good enough job of providing the right navigation/tools/UX.

I understand the thought process - just wanted to remark that I do that all the time, across many apps. Like a 'bring me back to basics" step. Anyway, looking forward to seeing this deployed and seeing how much I like it.

Is this only on "refresh" within one tab or is this somehow _cookie_ based? Like what happens if I have two tabs open of a project which I'm using for different 'views' to the same data?

Innders:
>  just wanted to remark

Definitely good remark, I think we will find out quite quickly if it's super annoying. 

> somehow cookie based

It's localstorage so a new tab will also contain that data. It's not like if you open a folder it will open on the other tab as well. But if you were to refresh that second tab it would mimic the first tab, which is a small side effect.

> 'bring me back to basics"

Maybe the URI bar could be used to "reset" the path with a reset button or something.

image
image
image

BigRoy:
> Maybe the URI bar could be used to "reset" the path with a reset button or something.

That could be interesting - anyway, don't let me hold you back. Let's see how this "feels" in production down the line. I tend to be hesitant about stuff that in the end doesn't really matter.

martastain (approved):
Everything works, but i kinda agree with @BigRoy that we should have a way to reset the context without going to devtools. Not sure if there's anything blocking and it could be added later... just nice to have

mkolar:
Very good!

Regarding the reset, what exactly should reset? the selection? To what? Nothing selected? There is not "default" selection as far as I'm aware. 

I quite like the idea of reseting selection on the URI bar, but not sure if people will realize that. Worth a shot maybe. I just tried a whoole bunch of PM systems. Jira, GH, clickup... and they all try to keep the selection and sometimes fail, so the fact things reset is more of a byproduct of buggy implementation usually. 

Does is even make sense to not have context selected? At minimum you tend to have a project right?

BigRoy:
> I quite like the idea of reseting selection on the URI bar, but not sure if people will realize that. Worth a shot maybe. I just tried a whoole bunch of PM systems. Jira, GH, clickup... and they all try to keep the selection and sometimes fail, so the fact things reset is more of a byproduct of buggy implementation usually.
> 
> Does is even make sense to not have context selected? At minimum you tend to have a project right?

I think there's a difference between having your "context" selected (like what the AYON tools do inside DCCs as well) as opposed to persisting the user selection. If it's just about the current context that the user is in - then yes, definitely auto-select it. (Usually, in many PM systems I think that context is even part of the URL - and as such, it makes total sense that _that_ is getting selected). But if I have expanded/unexpanded many things, selected five assets and then refresh - I believe many PMs might just reset to the current context. No?

Innders:
> But if I have expanded/unexpanded many things, selected five assets and then refresh

TLDR; if you are refreshing to perform an action then the UX is broken, you should never have to refresh, ever.

I think this comes down to historically broken web experiences. The refresh button has been used to "fix" something that was broken and we have all just got used to it. But now the web app experience should be more like an... **native app** and we should be pushing towards that goal.

If you expand a load of folders in Maya, how do you collapse them all, you can't refresh there.

Really the end goal is to completely remove any need for the refresh button at all and this would actually make this PR redundant.

But I would argue for now it's better to preserve state for those times where a version hasn't updated and you need to refresh to get the most up to date information. When doing that you don't want to lose the selection on that version.

Overtime we will fix these bugs that require refreshes and maybe even to the point this code could be removed, because no refresh would ever be needed.

BigRoy:
> If you expand a load of folders in Maya, how do you collapse them all, you can't refresh there.

Just to answer this question: I close and reopen the tool/app/ui. 
So if opening a new tab in the browser works for that, then we're good?

PR #311, opened 2024-04-01
Version NONE button fix

Innders:
## Changelog Description

**Version NONE** button was fixed in bundles.

## Additional Information

`/settings/bundles`

The bundle was being set to `undefined` and then would not be included in the addons object for the PATCH request. Setting the addon version to `null` ensured it was included in the object.

image

## Testing

1. Select an addon/s using the mouse or side bar buttons.
2. Click `Version NONE` to set the addon/s version to `NONE`
3. Click save and the version `NONE` should persist.

### DIFF
--- a/src/pages/SettingsPage/Bundles/NewBundle.jsx
+++ b/src/pages/SettingsPage/Bundles/NewBundle.jsx
@@ -170,7 +170,7 @@ const NewBundle = ({ initBundle, onSave, addons, installers, isLoading, isDev, d
       const newAddons = { ...newFormData.addons }
       for (const addon of selectedAddons) {
         if (!latest) {
-          newAddons[addon.name] = undefined
+          newAddons[addon.name] = null
           continue
         }
         const versionList = Object.keys(addon.versions || {})


LudoHolbik:
So it work and `null` value is included in addons object. However, I am not sure why the change was implemented in the first place since when I tested it by creating new bundle, even `undefined` values were being saved and displayed value was NONE (for both `undefined` and `null`). I have tried both with and without development mode. We can discuss this on call if you want.

image

martastain:
when set to `undefined`, backend SHOULD keep the existing value, while `null` should clear the version. "should" is the important word here. could you confirm?

Innders:
> I have tried both with and without development mode.

The key difference is how the server handles missing add-on versions during updates.

### Creating a new bundle (POST request):
- If an add-on version is missing (set to undefined), it's excluded from the data sent to the server.
- Since it's a new add-on, the server automatically sets the version to null.

### Updating an existing bundle (PATCH request):
- Excluding the add-on version from the request (like in creating a new bundle) won't update the version on the server.
- To remove the version or set it to null, you need to explicitly send addon: null in the request.

### In simpler terms:

When creating a new bundle, forget to specify a version? No problem, the server will set it to null for you.
Updating an existing bundle and want to remove the version? Don't just skip it, tell the server explicitly with addon: null.

martastain:
To provide more context:
This allows to selectively change the version of a single addon without sending the rest of the array. Now - with server addons, which you can change without creating a new bundle - it makes sense. Originally this was not possible and changing an addon version required a new bundle. This is now required only when you need to change a pipeline addon.

LudoHolbik:
Ok it makes sense now, I was creating instead of updating which was the difference that raised questions. All clear for me , you can merge.

LudoHolbik (approved):

PR #312, opened 2024-04-01
Dashboard renamed to Home

Innders:
## Changelog Description

- **Dashboard** renamed to **Home** to better reflect pages description.
- **Home** button moved to far left to better align with web standards.
- Use classnames for button booleans instead of styled components variables.

## Additional Information

image

### DIFF
--- a/src/containers/header/AppHeader.jsx
+++ b/src/containers/header/AppHeader.jsx
@@ -1,6 +1,6 @@
 import { useEffect, useRef } from 'react'
 import { Link, useLocation, useNavigate } from 'react-router-dom'
-import { Spacer, Toolbar, UserImage, InputSwitch } from '@ynput/ayon-react-components'
+import { Spacer, UserImage, InputSwitch } from '@ynput/ayon-react-components'
 
 import Breadcrumbs from '../Breadcrumbs/Breadcrumbs'
 import HeaderButton from './HeaderButton'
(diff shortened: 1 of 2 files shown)


martastain:
I agree it is better from the semantic point of view, but:

- projects button opens the leftmost sidebar, which i would expect left most button would do
- you're destroying current user muscle memory :-(

BigRoy:
If the "Home" is not project-specific but a studio wide home where you see multiple projects then to me it being more to the left makes more sense.

I agree @martastain that it may be slightly confusing that the Project Sidebar is to the left and you tend to then expect the buttons to be close to the left. Mostly that now with the button _not_ being on the left I would not expect it to open a sidebar at all but to somehow go to another tab or whatever.

---

Do with that info whatever you want - I'm fine with both the old and new.

BigRoy:
Also, side note - I keep getting lost to where the "Create new project" button appears. (It only appears in Dashboard > Overview?)
With "Home" now being the more left button maybe there's a higher chance I get there after this PR. :)

However, is it maybe also a good idea to add a Create new project button at the bottom of the projects sidebar?

project_sidebar_create_new

Innders:
> Create new project button at the bottom of the projects sidebar

I like the idea and it seems like an obvious one but... we have tried this before and found it was unnecessary to have a button so prominent when it will only be used on average once a month.

We also extremely strict about what we put in this side bar and want to keep it as clean as possible.

Also we must remember that testing requires creating more projects but this isn't normal.

Innders:
But I also agree it's difficult to find.

@mkolar what do you think about this?

image

BigRoy:
> image

This might be prone to misclicks - anyway, if we do add it there - I'd swap them around at least.

---

Also, it might also already help if the "Tasks" page also shows the Create Project button in Home. And that instead maybe the "Select all projects" button becomes an extra button at the top of the 'collapse projects sidebar' in tasks view.

E.g. from this:
image


To:
image

Hmm - yea, no. Not sure if that's nice either.

mkolar:
> Also, it might also already help if the "Tasks" page also shows the Create Project button in Home. 

I'll have to give that a hard no. Tasks dashboard will 90% of the time used by artists, creating a new project could not be less relevant there. 

> @mkolar what do you think about this?

I feel that it was actually all clearer way before, when we allowed to got to "manage projects" page from the project sidebar. Even the fact that we now call it "projects settings" makes it slightly harder to understand that it's place where i also happen to create new ones. My humble opinion is that "Project settings" should be called "Manage projects" as it used to. And that same button could totally be in the apps menu as it is now and also on the bottom of the Project sidebar. 

> projects button opens the leftmost sidebar, which i would expect left most button would do

Very good point. I expected projects sidebar to be changed to the same style as the apps menu and user menu. Faster and kinda not a sidebar really.  In that case it feels like a quick access project menu and we can totally add things like button to create a new projects, jump to settings an so on. 


Sidebar would be no more then, however we could reintroduce it as quick navigation to anywhere for user stored shortcuts.

BigRoy:
> I'll have to give that a hard no. Tasks dashboard will 90% of the time used by artists, creating a new project could not be less relevant there.

Fair enough. 👍 

> I feel that it was actually all clearer way before, when we allowed to got to "manage projects" page from the project sidebar. Even the fact that we now call it "projects settings" makes it slightly harder to understand that it's place where i also happen to create new ones. My humble opinion is that "Project settings" should be called "Manage projects" as it used to. And that same button could totally be in the apps menu as it is now and also on the bottom of the Project sidebar.

Haha, I had no idea there was a "Create Project" button there. I was using it from the "Dashboard > Overview" instead.

mkolar:
> Haha, I had no idea there was a "Create Project" button there. I was using it from the "Dashboard > Overview" instead.

Case in point I guess

Innders:
Okay so @mkolar and @BigRoy it has been a month since switching the buttons... and I still go to the wrong place every single time. Maybe it's years of muscle memory for this one action, or maybe I was wrong.

I think this should be reviewed with some user testing with users that have never seen the website before.

PR #327, opened 2024-04-06
Redesign adding users to projects

Innders:
#### 🚀  Features

- User access groups table for easily assigning users to projects from a user perspective.
- Multi selection support.
- Click and drag selection support (like in photoshop hide/reveal layers).
- Select multiple users to set their access groups at one time.

image

#### 💅 Enhancements

- Create new user modal dialog with options to create and close or create and continue.
- Create new user shortcuts `shift+enter` to create and continue and `ctrl/cmd+enter` to create and close.
        
 image

- New `n` shortcut to create a new user.
- Original access groups is now always "New projects access" no matter what project is selected.
- When a project is selected "Selected projects access" shows to add access to specific projects.
- `( default )` removed from project list and moved to button above called `All users`.
- `All users | Selected projects` switch in favour of `Assigned only` switch.
image

#### 🐛 Fixes

- Tooltip to explain what "New projects access" is with text: "When a new project is created, the user will be added to these access groups by default."
- Updating access groups wasn't invalidating the correct user tags.
- Multisave confirmation warning popup removed as saving now supports mixed values in access groups.
- Access group dropdowns now show when the value is "mixed" on multi-selected users.
- Users with no access to a project now won't show on the teams page.

image


mkolar:
Great changes, They should make the experience much clearer and better. Small semantic things....Are we adding users to access groups, or assigning access groups to users? We should be consistent in that wording

martastain:
I'd say adding. We're assigning user to tasks and it might be slightly confusing if we reuse the term.

Innders:
I have fixed the wording to always use "add" when referencing access groups.

I have two design tweak ideas to make the sidebar clearer.

1. Button label tweaks
image

2. The title of the project list becomes the switch itself
image

martastain (approved):

PR #351, opened 2024-04-17
fix: remove other roles if user is admin/manager

LudoHolbik:
## Changelog Description

- I have added a simple function that removes any other roles, if the new role is `manager` or `admin`
- there was small caveat - if user is admin he is also a manager. I believe this might be a feature, not a bug so I have adjusted conditions accordingly.

image

https://github.com/ynput/ayon-frontend/assets/16327908/332f72f0-ebd0-4742-aca3-5c57bf8ad386

### DIFF
--- a/src/pages/SettingsPage/UsersSettings/userDetail.jsx
+++ b/src/pages/SettingsPage/UsersSettings/userDetail.jsx
@@ -313,10 +313,16 @@ const UserDetail = ({
       )
       .flat()
 
+    // if user is admin or manager, clear other roles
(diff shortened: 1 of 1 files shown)


Innders (changes requested):
Initial issue fixed but I just noticed this when a user is a service. It should only say service.

image

LudoHolbik:
I have added one condition for service.


https://github.com/ynput/ayon-frontend/assets/16327908/e7bf0bec-9409-41a8-ab75-9c545f9d3817

Innders (approved):

PR #353, opened 2024-04-17
enhancement: created check for current version in BundlesAddonList

LudoHolbik:
## Changelog Description
- Added check for version in BundlesAddonList
- If addon has not latest version, id displays icon
- When hovering over icon, latest version gets displayed as tooltip

https://github.com/ynput/ayon-frontend/assets/16327908/7eb5efa3-4200-4106-a8f0-161d74ff71d1


Innders:
Do you think it would make sense to do the version check in the map on line 129 and add a `isLatest` field?
We are already mapping so there's no extra loops and it might clean up the `AddonItem` component as well.

```js
const addonsTable = useMemo(() => {
  return addons.map((addon) => {
    return {
      ...addon,
      version: formData?.addons?.[addon.name] || 'NONE',
      dev: formData?.addonDevelopment?.[addon.name],
      isLatest: checkLatestVersion(currentVersion,allVersions)
    }
  })
}, [addons, formData])
```

LudoHolbik:
Since checking for latest version is in the end just simple comparison I would maybe leave it to AddonItem ?

`const isCurrentLatest = currentVersion === latestVersion`

Innders (approved):

PR #366, opened 2024-04-22
Delete access groups and clear overrides

LudoHolbik:
## Changelog description

1. added option for deleting whole group (Button - Delete access group)
2. added option for deleting local project settings of the group (Button - Delete project group settings)
3. Minor change - ( default ) Changed to "Global [All Projects]" - it reflects better to what id does
4. Minor change - switch order of buttons 

image

https://github.com/ynput/ayon-frontend/assets/16327908/db903a7c-779f-4440-94cd-dc033048653e

### DIFF
--- a/src/pages/SettingsPage/AccessGroups/AccessGroupDetail.jsx
+++ b/src/pages/SettingsPage/AccessGroups/AccessGroupDetail.jsx
@@ -4,6 +4,7 @@ import {
   Button,
   Section,
   Toolbar,
+  Spacer,
   ScrollPanel,
   SaveButton,
 } from '@ynput/ayon-react-components'
@@ -93,12 +94,7 @@ const AccessGroupDetail = ({ projectName, accessGroupName }) => {
   return (
     <Section style={{ flex: 2 }}>
       <Toolbar>
-      <SaveButton
-          onClick={onSave}
-          label="Save Changes"
-          active={isChanged}
-          saving={saving}
-        />
+        <Spacer />
         { isLocalProject &&
           <Button
             onClick={onDeleteLocalGroupSettings}
@@ -107,6 +103,12 @@ const AccessGroupDetail = ({ projectName, accessGroupName }) => {
             icon="delete"
           />
         }
+        <SaveButton
+          onClick={onSave}
+          label="Save Changes"
+          active={isChanged}
+          saving={saving}
+        />
       </Toolbar>
       <ScrollPanel
         className="nopad transparent"


Innders:
Just going to preemptively say that I think that there should be two buttons. One for deleting globally and one for deleting project level.

LudoHolbik:
Did you mean :
1. two buttons visually
2. two buttons in code
3. both 
?

image

image

Innders:
There will need to be two different buttons in the UI.

Green: **Delete access group**

Pink: **Remove project overrides**

martastain:
nice drawing :)
but since only one of these two buttons can be active at the same time, couldn't it just change the context?

LudoHolbik:
I think actually both can be active. Deleting project access group just deletes **project settings** of the group, not the group per se.

Innders:
The buttons do different things and switching a delete button is dangerous in my opinion. 

The main delete button will always be active.

The project overrides button will only be disabled when no project is selected.

LudoHolbik:
You can see final version in description. After consultation with Martin I removed "Delete settings" button for Global project, since the logic for it would not be future-proof. 

Result is:
1. On Global level, you can create/delete groups and change their global properties by saving them
2. On Local (project) level, you can create/delete groups and change their local properties by saving them _or just deleting (removing) all of them_ . Deleting has same effect as setting them to disabled (same thing can be achieved by just individually turning them off and saving them).

We agreed with Martin that there should be some visual differentiation between local(project) settings and global settings, since this can cause some confusion for user.

Innders (changes requested):
Awesome! Looks like it's functional but I have a couple of UI suggestions.

- Save button on the right side to stay consistent with the rest of the app. 
- The save button should always say "Save access group".

- "Delete project group settings" button:
    - rename to "Clear project overrides"
    - remove red danger colour
    - move to the left side of the settings section
    - dialog: title "Clear project overrides". Description: "Are you sure you want to delete all project override settings for this access group?

- All delete/clear buttons use trash icon
- Delete modal: title remove "global" word. Description: remove the last two lines (keep only the first line).
- Global -> "Default (all projects)"

LudoHolbik:
I have made the UX changes. This is current state:


https://github.com/ynput/ayon-frontend/assets/16327908/71044774-fe5b-4ecb-8f1b-282065f7a69b

Innders:
**Changes**

- Save button on the right side to stay consistent with the rest of the app.
- dialog: title "Clear project overrides".

- Clear overrides dialog confirmation button "delete" -> "clear".
- Remove red danger colour for the other button. It looks weird now and I'm not sure we have it anywhere else.

**Enhancements**

We try to implement right clicks across the app that clone toolbar buttons and this location would benefit from it. 

Right click on an access group:
- "Clear overrides" (disabled when not project selected)
- "Delete" with "danger" colour.

LudoHolbik:
I made those visual changes . I changed the text in save button as I was not sure what do you mean by "Save button on the right side to stay consistent with the rest of the app."

Regarding enhancements - do you mean menu that appears after right click like the one at the very end of the video ?

https://github.com/ynput/ayon-frontend/assets/16327908/665c6566-3923-4c4b-b745-6a3ae6966eea

Innders:
The first commit the buttons switched sides. Now the buttons are all on the left. The save button should be positioned on the far right of the toolbar. Look at editor/settings/users etc.. for where the save button is positioned.

LudoHolbik:
I made following changes:
- both buttons are now on the right side (like in Editor section) - i wanted just save button on the right but two "delete" buttons next to each other looked a bit weird, plus again - this is similar layout like in Editor
- three columns now do not change their width when clicking between global and local project (issue reported by Martin)
- I have added right-click menu for clearing settings and deleting group

Demo:

https://github.com/ynput/ayon-frontend/assets/16327908/b4c7bb96-e07a-46f2-8b17-b6bc43aa22d4

martastain (approved):

PR #368, opened 2024-04-23
fix: Added new 'Status' column in EditorPage

LudoHolbik:
## Changelog Description

- Added new Status column
- Values in column change color when they are updated in Editor Panel on the right
- Status column can be filtered out by "Show columns" dropdown at the top

Demo:

https://github.com/ynput/ayon-frontend/assets/16327908/8112602a-9189-4615-963b-f16f8ec118bf

### DIFF
--- a/src/pages/EditorPage/EditorPage.jsx
+++ b/src/pages/EditorPage/EditorPage.jsx
@@ -18,7 +18,7 @@ import sortByKey from '/src/helpers/sortByKey'
 
 import { editorSelectionChanged, setUri, setExpandedFolders } from '/src/features/context'
 
-import { getColumns, formatType, formatAttribute, formatAssignees } from './utils'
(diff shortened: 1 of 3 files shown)


Innders (changes requested):
A good start. The status field needs to look the same as the browser.

- status icon
- status colour

- The "changed" styles don't match the other fields?

LudoHolbik:
I have:
- added status icon and color
- changed update styles so that they look more like in other fields like FPS, Width, height, etc. ( Previously they had changed in same way like in "Type" column, which I found out is maybe only one that changes color of the text instead of highlighting background)



https://github.com/ynput/ayon-frontend/assets/16327908/c13cf855-59b5-458f-bac3-ef36fa54bad0

martastain:
Any chance of change display mode


- icon + full name
- icon + short name
- just icon

based in the column width similarly to browser/version status column?

martastain (approved):
Works like a charm!

PR #369, opened 2024-04-25
Bundle cross-checking errors and warnings

martastain:
Original issue: https://github.com/ynput/ayon-backend/issues/123

https://github.com/ynput/ayon-backend/pull/127 adds a way of validating a bundle before it is saved. 
It takes in account addon cross-compatibility and returns a list of issues (warnings and errors with the bundle)

While errors should disallow bundle creation, warnings should be just displayed to let the user know something may not work properly. If the bundle is a dev bundle then it can be created or updated with errors.

image


martastain:
Current implementation of patching existing bundles (changing server addons versions) may be problematic. Frontend now sends every change directly to the server, so in case you want to update two addons dependent on each other to the newest version, you can't do that.

PR #386, opened 2024-05-05
Fix: move access groups to under access level

Innders:
#### 1. Move **Guest** and **Developer** above **Access level** in order to couple **Access level** and **New project access** together.

image

#### 2. Add a small note about admins, managers and services having full project access.
image

#### 3. Disable (instead of hide) guest switch when a user is an admin and add a tooltip.
image


LudoHolbik:
Looks good. 👍

LudoHolbik (approved):

PR #387, opened 2024-05-05
Enhancement: 362 updating addons from marketplace does not handle errors

Innders:
### Changelog Description

When an addons fails to download from the market it shows an error message.

https://github.com/ynput/ayon-frontend/assets/49156310/e34f4e5b-722e-4bb2-9567-a0df6388d829


martastain (approved):
Tested. Works perfectly.

PR #394, opened 2024-05-09
fix: replaced user data source to server for showing developer mode

### LINKED ISSUE #335: Developer mode button not showing/disappearing after saving user
**Describe the bug**
When switching user field **Developer** on/off a refresh is required to show **Developer Mode** in the nav bar.

**To Reproduce**
1. Go to /settings/users
2. Click on a user.
3. Switch **Developer** field.
4. Click save.
5. Nothing happens in the navbar.

**Expected behavior**
**Developer Mode** button should appear/disappear without having to refresh after saving.
image

**Screenshots**
image

**Additional context**
Probably something to do with updateUsers query not invalidating some tags that would refresh the data for the button to appear.

LudoHolbik:
## Description of changes

Issue link: https://github.com/ynput/ayon-frontend/issues/335
URL: http://localhost:3000/settings/users

- replaced source of information about user - server instead of redux
- developer mode now appears/disappears without need for refresh

### Additional context

https://github.com/ynput/ayon-frontend/assets/16327908/2b1142e6-9025-4cb2-909f-67f4b7b53c35

close #335

### DIFF
--- a/src/containers/header/AppHeader.jsx
+++ b/src/containers/header/AppHeader.jsx
@@ -13,7 +13,6 @@ import { toggleMenuOpen, setMenuOpen } from '/src/features/context'
 import { HelpMenu, UserMenu } from '/src/components/Menu'
 import MenuContainer from '/src/components/Menu/MenuComponents/MenuContainer'
 import { useUpdateUserMutation } from '/src/services/user/updateUser'
-import { useGetMeQuery } from '/src/services/user/getUsers'
 import { toast } from 'react-toastify'
 import { onProfileUpdate } from '/src/features/user'
 import styled from 'styled-components'
@@ -70,8 +69,8 @@ const Header = () => {
   const handleSetMenu = (menu) => dispatch(setMenuOpen(menu))
   const location = useLocation()
   const navigate = useNavigate()
-  // get user info from BE instead of redux to have up to date information
-  const { data: user } = useGetMeQuery()
+  // get user from redux store
+  const user = useSelector((state) => state.user)
 
   // restart server notification
   const { isSnoozing } = useRestart()
--- a/src/containers/header/AppHeader.jsx
+++ b/src/containers/header/AppHeader.jsx
@@ -13,6 +13,7 @@ import { toggleMenuOpen, setMenuOpen } from '/src/features/context'
 import { HelpMenu, UserMenu } from '/src/components/Menu'
 import MenuContainer from '/src/components/Menu/MenuComponents/MenuContainer'
 import { useUpdateUserMutation } from '/src/services/user/updateUser'
+import { useGetMeQuery } from '/src/services/user/getUsers'
 import { toast } from 'react-toastify'
 import { onProfileUpdate } from '/src/features/user'
 import styled from 'styled-components'
@@ -69,8 +70,8 @@ const Header = () => {
   const handleSetMenu = (menu) => dispatch(setMenuOpen(menu))
   const location = useLocation()
   const navigate = useNavigate()
-  // get user from redux store
-  const user = useSelector((state) => state.user)
+  // get user info from BE instead of redux to have up to date information
+  const { data: user } = useGetMeQuery()
 
   // restart server notification
   const { isSnoozing } = useRestart()
--- a/src/containers/header/AppHeader.jsx
+++ b/src/containers/header/AppHeader.jsx
@@ -114,18 +114,23 @@ const Header = () => {
   const [updateUser] = useUpdateUserMutation()
   const handleDeveloperMode = async () => {
     try {
+      const newDeveloperMode = !developerMode
+      // optimistic update the switch
+      dispatch(toggleDevMode(newDeveloperMode))
+
       await updateUser({
         name: user.name,
         patch: {
-          attrib: { developerMode: !developerMode },
+          attrib: { developerMode: newDeveloperMode },
         },
       }).unwrap()
 
-      // update redux state with new data
-      dispatch(toggleDevMode(!developerMode))
+      // if the request fails, revert the switch
     } catch (error) {
       console.error(error)
       toast.error('Unable to update developer mode: ' + error.details)
+      // reset switch on error
+      dispatch(toggleDevMode(developerMode))
     }
   }
 
--- a/src/containers/header/AppHeader.jsx
+++ b/src/containers/header/AppHeader.jsx
@@ -14,7 +14,7 @@ import { HelpMenu, UserMenu } from '/src/components/Menu'
 import MenuContainer from '/src/components/Menu/MenuComponents/MenuContainer'
 import { useUpdateUserMutation } from '/src/services/user/updateUser'
 import { toast } from 'react-toastify'
-import { onProfileUpdate } from '/src/features/user'
+import { updateUserAttribs } from '/src/features/user'
 import styled from 'styled-components'
 import { useRestart } from '/src/context/restartContext'
 import { classNames } from 'primereact/utils'
@@ -122,7 +122,7 @@ const Header = () => {
       }).unwrap()
 
       // update redux state with new data
-      dispatch(onProfileUpdate({ developerMode: !developerMode }))
+      dispatch(updateUserAttribs({ developerMode: !developerMode }))
     } catch (error) {
       console.error(error)
       toast.error('Unable to update developer mode: ' + error.details)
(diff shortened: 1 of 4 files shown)


Innders (changes requested):
Sorry to be annoying... yes this technically fixes this isolated circumstance, but it doesn't fix the underlying bug, which is that the redux store isn't in sync.

This same bug could happen again if the real bug isn't fixed.

Innders:
Actually I've found another bug related to this already, so we can actually solve a lot stuff.


https://github.com/ynput/ayon-frontend/assets/49156310/846b2583-d838-485f-b17e-14a9320de06a

Innders:
I was thinking about the architecture of data fetching state and thought it would be useful to create a diagram of how it currently works.

redux-rtk-query-architecture-getUser-v002

There a couple of benefits to syncing the logged in users data with a slice, instead of always calling useGetUser().

1. `use{query}` is a fairly heavy compared to `useSelector()`.
2. `use{query}` is used for getting new data, we don't need new data as the user data is fairly static.
3. `useSelector((state) => state.user.name)` subscribes to only changes to `name`, whereas `use{query}` will re-render the component when any part of `user` changes. We can be more specific about our re-renders.
4. For authentication we never need to use `useGetUser()` because `/api/auth/login` returns the user data for us. 1 vs 2 api calls.
5. This is the recommended way in the [redux docs](https://redux-toolkit.js.org/rtk-query/usage/examples#authentication).

I'll admit that we need to properly sort out auth and I've created a task for that -> #406 

So in summary, this doesn't change how this issue should be resolved, the redux state needs updating. 

In fact this already being done in profile just no in users (when the user is me).

[onProfileUpdate: (state, action) => {](https://github.com/ynput/ayon-frontend/blob/fc563048acd29a9b5eadd71b777a62ec45501c56/src/features/user.js#L31-L34)

LudoHolbik:
Thanks for diagram and additional context. Regarding your points - agree with redux solution. 

I have added reducer so developer mode is now up to date in redux state. All user data is now being updated in redux store after sucessful save. Its even a bit faster. 

https://github.com/ynput/ayon-frontend/assets/16327908/82891ff6-89d3-4bb7-9600-416d7c3555dd

src/features/user.js
```diff
@@ -32,8 +32,12 @@ const userSlice = createSlice({
       if (!state.attrib) return
       state.attrib = { ...state.attrib, ...action.payload }
     },
+    setUserData: (state, action) => {
```
Innders, line 35: Do we need a new action here? Could we not just have used `onProfileUpdate` as it's exactly the same? Maybe change to name to make it more generic, but probably fine.

https://redux.js.org/style-guide/#model-actions-as-events-not-setters

src/pages/SettingsPage/UsersSettings/userDetail.jsx
```diff
@@ -280,6 +283,7 @@ const UserDetail = ({
         }).unwrap()
 
         toast.update(toastId.current, { render: `Updated user: ${user.name} ` })
+        dispatch(setUserData(data))
```
Innders, line 286: Save a user that isn't you and it will update your user state. Only set when `user.self` is true.

Innders, line 286: This only updates the user's data and misses the attrib data like `fullName`. Look at ProfilePage.jsx

Innders (changes requested):

src/pages/SettingsPage/UsersSettings/userDetail.jsx
LudoHolbik: My mistake, fixed.

src/features/user.js
LudoHolbik: AppHeader where developer mode switch is located uses two different containers for user data - `data` object and `attribs` object. I thought `onProfileUpdate` reducer is different to reflect this structure as it updates `attribs`, not `data` and that we want to have them separated like this (including reducers)...

I have changed the name to updateUserData...

image

src/pages/SettingsPage/UsersSettings/userDetail.jsx
LudoHolbik: Ok so this relates to first comment. I have updated user reducer names in `user.js` and their imports.
I have also added update for user attributes for userDetail. Do you agree with this approach ?

image

src/features/user.js
LudoHolbik: looking at the guide from redux, maybe `userDataUpdate` and `userAttribsUpdate` would be better ? But maybe Im overthinking it...

src/pages/SettingsPage/UsersSettings/userDetail.jsx
LudoHolbik: looking at the guide from redux, maybe `userDataUpdate` and `userAttribsUpdate` would be better ? But maybe Im overthinking it...

src/features/user.js
Innders: It's not about what data we are updating, it's about what event is causing the update. If onProfileUpdate and onMyUserUpdate update different data then it is correct to seperate them out into two different actions because they are different events that update different data.

But let's stay consistent with the naming convention and rename `updateUserData` to something like `onProfileUpdate`.

> we recommend trying to treat actions more as "describing events that occurred", rather than "setters"

(I know we have setSomeData naming all over actions, but that's bad and we should try to avoid it if we can)

src/containers/header/AppHeader.jsx
```diff
@@ -122,7 +122,7 @@ const Header = () => {
       }).unwrap()
 
       // update redux state with new data
-      dispatch(onProfileUpdate({ developerMode: !developerMode }))
+      dispatch(updateUserAttribs({ developerMode: !developerMode }))
```
Innders, line 125: Because we are trying to use event based naming, maybe this should be `onDevModeChange` or `toggleDevMode` with a new action that only toggles the developerMode.

src/pages/AccountPage/ProfilePage.jsx
```diff
@@ -124,6 +124,8 @@ const ProfilePage = ({ user = {}, isLoading }) => {
     }
   }
 
+  console.log(user, 'USERZZZ')
```
Innders, line 127: Remove console log

src/features/user.js
```diff
@@ -28,12 +28,16 @@ const userSlice = createSlice({
       state = {}
       return state
     },
-    onProfileUpdate: (state, action) => {
+    updateUserAttribs: (state, action) => {
```
Innders, line 31: revert back to `onProfileUpdate`.

src/pages/SettingsPage/UsersSettings/userDetail.jsx
```diff
@@ -280,6 +283,7 @@ const UserDetail = ({
         }).unwrap()
 
         toast.update(toastId.current, { render: `Updated user: ${user.name} ` })
+        user.self && dispatch(updateUserData(data)) && dispatch(updateUserAttribs(data))
```
Innders, line 286: Only `dispatch(updateUserAttribs(data))` gets dispatched as ` user.self && dispatch(updateUserData(data))` is the ternary operator. 

If we had `onMyUserUpdate` then we would only need to dispatch that one action. 

https://redux.js.org/style-guide/#avoid-dispatching-many-actions-sequentially

Innders (changes requested):

src/pages/SettingsPage/UsersSettings/userDetail.jsx
Innders: Sorry `dispatch(updateUserData(data))` does get called because it's checking if the value is true or false.

If for any reason `dispatch(updateUserData(data))` was false then `dispatch(updateUserAttribs(data))` would not run. This should be an `if` statement.

Innders: Also try updating the fullName and pressing save, it doesn't update the fullName in the user menu. Probably because the wrong data is being sent.


https://github.com/ynput/ayon-frontend/assets/49156310/f1f2adcb-c85f-4724-a539-01cd318f0813


image

src/pages/AccountPage/ProfilePage.jsx
LudoHolbik: Removed.

src/containers/header/AppHeader.jsx
LudoHolbik: Added toggleDevMode in user slice, used here.

PR #400, opened 2024-05-10
fix(comment): \\ replaced with \\\ so that it renders as \\

Innders:
## PR Checklist

-   [ ] Commenting a file path `D:\\file\to\path` renders as `D:\\file\to\path` with preserved double backslash.
-   [ ] Other uses of `\` to escape characters are preserved.

## Description of changes

Comments with `\\` are replaced with `\\\` so that it will render as `\\` when parsed as markdown.

### Attachments

https://github.com/ynput/ayon-frontend/assets/49156310/9b6ed6ad-447f-4d4c-91b0-54617431d24d

### Additional context

Fixes this specific issue with paths but pasting paths could be way more useable. Maybe a path could be "attached" like an image and have actionable features, like copying or even opening file.

### DIFF
--- a/src/components/CommentInput/helpers.jsx
+++ b/src/components/CommentInput/helpers.jsx
@@ -35,6 +35,16 @@ turndownService.addRule('taskListItems', {
   },
 })
 
+// if there is a double backslash followed by text, \\text, replace with \\\ so it resolves to \\text
+turndownService.addRule('doubleBackslash', {
+  filter: function (node) {
+    return node.innerText?.includes('\\\\')
+  },
+  replacement: function (content) {
+    return content.replaceAll('\\\\', '\\\\\\')
+  },
+})
+
 // replace <p> with <br> for line breaks
 const replaceLineBreaks = (html) => {
   return html.replaceAll('<p>', '').replaceAll('</p>', '\n').replaceAll('```', '')
--- a/src/components/CommentInput/helpers.jsx
+++ b/src/components/CommentInput/helpers.jsx
@@ -41,7 +41,7 @@ turndownService.addRule('doubleBackslash', {
     return node.innerText?.includes('\\\\')
   },
   replacement: function (content) {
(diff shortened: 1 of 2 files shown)


src/components/CommentInput/helpers.jsx
```diff
@@ -35,6 +35,16 @@ turndownService.addRule('taskListItems', {
   },
 })
 
+// if there is a double backslash followed by text, \\text, replace with \\\ so it resolves to \\text
```
BigRoy, line 38: Was this comment lacking a backslash or is there a reason it's three?
```suggestion
// if there is a double backslash followed by text, \\text, replace with \\\\ so it resolves to \\text
```

Innders: I don't see why it would be 4 slashes? The string `\\test` if being replaced with `\\\test` (3 \\\).

Innders: Okay I see what you mean now. More a question of if `\\test` should be replaced with `\\\test` or `\\\\test`.

BigRoy: Yup - I wonder what, if escaping only one what would happen with text like `\\n` becoming `\\\n` which technically would mean a backslash and a new line. Whereas `\\\\n` would mean two backslashes and the `n` character?

LudoHolbik: I am a bit struggling here why does it matter since from what Luke wrote in either case (triple or quadruple backslash) it can be resolved back to `\\n`. Only case if you want make comment AND make a new line (?)

That being said, making it double and then diving it into half seems more reasonable approach... (so `\\` to `\\\\` and back  to `\\`).

Innders: Okayyyyy fine 4 it is. But only because it makes this code even more silly `content.replaceAll('\\\\', '\\\\\\\\')`

LudoHolbik (approved):

PR #404, opened 2024-05-10
fix(bundles): ensure only addons that are still installed copy across

Innders:
## PR Checklist

-   [ ] Duplicate a bundle that once had a deleted addon (no idea how you know this)
-   [ ] You shouldn't get compatibility error `{addon_name}: {addon_name} {addon_version} is not active`

## Description of changes

When duplicating a bundle, ensure that are addons that are now **not** installed are removed from the new bundle.


martastain (approved):
Works great

PR #435, opened 2024-05-17
Onboarding: Show addon version when selecting addons

### LINKED ISSUE #430: Bootstrap: Show addon version on pick addons page
## Story

A user wants to know if a particular addon version is included in a release.

### Proposal

- On `AddonSelectStep` show each addon with it's version.
- Grid columns to 2

### Attachments

image

LudoHolbik:
## Description of changes

- added information about version for addons selection
- added spinner (it can take up to few seconds to display addons)
- used title instead of name for addons
- changed 3 columns to two

### Additional context

https://github.com/ynput/ayon-frontend/assets/16327908/efa88d59-ffaa-4878-a6e8-924118ffef46

closes #430

### DIFF
--- a/src/pages/OnBoarding/util/OnBoardingContext.jsx
+++ b/src/pages/OnBoarding/util/OnBoardingContext.jsx
@@ -118,9 +118,9 @@ export const OnBoardingProvider = ({ children, initStep, onFinish }) => {
   const [selectedAddons, setSelectedAddons] = useState([])
 
   // get selected release data
-  const { data: release = {} } = useGetReleaseQuery(
+  const { data: release = {}, isLoading: isLoadingAddons } = useGetReleaseQuery(
     { name: selectedPreset },
-    { skip: !selectedPreset || stepIndex !== 7 },
+    { skip: !selectedPreset || stepIndex < 6 },
   )
 
   // // once releases are loaded, set selectedPreset to the first one and pre-cache each release
@@ -265,6 +265,7 @@ export const OnBoardingProvider = ({ children, initStep, onFinish }) => {
     setIsConnecting,
     isConnecting,
     isLoadingRelease,
+    isLoadingAddons,
   }
 
   return <OnBoardingContext.Provider value={contextValue}>{children}</OnBoardingContext.Provider>
--- a/src/pages/OnBoarding/util/OnBoardingContext.jsx
+++ b/src/pages/OnBoarding/util/OnBoardingContext.jsx
@@ -120,7 +120,7 @@ export const OnBoardingProvider = ({ children, initStep, onFinish }) => {
   // get selected release data
   const { data: release = {}, isLoading: isLoadingAddons } = useGetReleaseQuery(
     { name: selectedPreset },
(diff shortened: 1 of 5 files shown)


LudoHolbik:
This is for review. I just updated the branch to be up-to-date with develop.

src/pages/OnBoarding/util/OnBoardingContext.jsx
```diff
@@ -118,9 +118,9 @@ export const OnBoardingProvider = ({ children, initStep, onFinish }) => {
   const [selectedAddons, setSelectedAddons] = useState([])
 
   // get selected release data
-  const { data: release = {} } = useGetReleaseQuery(
+  const { data: release = {}, isLoading: isLoadingAddons } = useGetReleaseQuery(
     { name: selectedPreset },
-    { skip: !selectedPreset || stepIndex !== 7 },
+    { skip: !selectedPreset || stepIndex < 6 },
```
Innders, line 123: After testing I think this should be `< 5`. When going to step 6 we have to wait for the release to load the data. It's smoother if we pre-fetch the release on step 5 so the user doesn't have to wait on step 6.

src/pages/OnBoarding/util/OnBoardingStep.styled.js
```diff
@@ -188,3 +188,13 @@ export const NextButton = styled(SaveButton)`
       }
     `}
 `
+
+export const Spinner = styled.div`
```
Innders, line 192: Nice spinner but unfortunately we don't use spinners.

We try and use skeleton shimmers instead. In this case it would be the addon placeholder cards with the shimmer effect over them. We actually already have the number of addon placeholders to show from `releases` data, which is nice.

`${getShimmerStyles()}`

image

Innders (changes requested):
Working well, nice one. Just a couple small things to make it perfect 👌

src/pages/OnBoarding/util/OnBoardingStep.styled.js
LudoHolbik: Added shimmer effect. This is now how it looks now when loading state is true.


https://github.com/ynput/ayon-frontend/assets/16327908/02d6f0e6-509b-4aec-9aea-a57ad2e8d221

src/pages/OnBoarding/util/OnBoardingContext.jsx
LudoHolbik: Moved to step 5.

Innders (approved):
Placeholder loading look nice and slick now.

There were a couple things I have cleaned up.

1: isLoading to isFetching because if a different release is selected then the loading state needs to show again. See diff between the two here: https://redux-toolkit.js.org/rtk-query/usage/queries#frequently-used-query-hook-return-values

2. I noticed a quick flash when no addons were visible at all. This was because of some stupid useEffect I was using, I have removed it now.

3. The number of placeholders can match the number of actual addons loading, we already have that information even if `release` is loading. So use that length instead of fixed 20.

PR #444, opened 2024-05-20
Testing: setup playwright and first basic end to end tests

Innders:
## Description of changes

End to end testing is now implemented locally but NOT in the CI.

CI was taking too much effort for what it's worth. Local testing is fine.

Basic first tests:

- Create and delete project
- Create and delete user
- Create and delete bundle

### Technical details

First install all deps with `yarn install`

Test your PR locally using `yarn test`.

For a more visual test use `yarn test-ui`.

Create new tests with the help of `yarn test-codegen`.

Add new env variables for user auth (otherwise tests will fail)
```env
NAME=*********
TEST_SERVER_URL=********
PASSWORD=*********
```

library link: https://playwright.dev/


Innders (commented):
All the tests pass but I can't get authentication to pass. Maybe I have set it up wrong though.

image

Do you think that the tests could be separated even more now we have fixtures?

For example we have create and delete folders as one test. Does it make sense to now have two separate tests: **create folder** and **delete folder**. Or is that getting too separated?

PR #447, opened 2024-05-20
Editor: Editor dragging a value does not trigger thumbnail upload

Innders:
## Description of changes

Dragging a value inside of the editor panel should not show the file upload dropzone.

### Testing

Select some text in the editor panel and drag it, nothing should happen.


martastain:
Checked and it's fixed - just one small thing. Maybe you could include it in this PR if it is a small one, or i can fill a new issue:
I noticed that when you upload a thumbnail by dragging a file to that small box, the underlying data is not updated, so when you click another entity and back, the thumbnail disappears until you reload the page.

martastain (approved):

PR #451, opened 2024-05-21
Browser: Columns select dropdown

Innders:
### Description of changes

- Replace Prime react multiselect with arc dropdown.
- Value now defaults to empty array `[]` and will show all columns.
- Fixed issue where clearing would not work.
- Removed "(Single)" and "(Multiple)" as it was just confusing.
- New placeholder "Filter columns..."

### Additional context

https://github.com/ynput/ayon-frontend/assets/49156310/8bec03c2-81da-4237-9493-ad5e73d76fb4


martastain (approved):

BigRoy:
Ah, something I had missed - this is the Browser column filter. Does this also fix the Editor column filter?
And, additionally, the browser column filter shows "nice labels" for the columns whereas the Editor column filter seems to show the columns in the dropdown using the key like `frameStart` instead of `Start Frame`.

PR #493, opened 2024-06-03
Comment: reference link styling, deleting and tooltips

Innders:
## Description of changes

Reference links inside the comment editor now act the same as they would inside a comment.

- Added reference styling.
- Hover over the reference to see it's tooltip.
- Click on a reference to open the slide out.
- Hitting backspace now delete the whole reference (instead of the just the first character).

### Technical details

- This required creating a new Quill format called `mention`.
- Moved the reference tooltip state to redux so it could be mutated from anywhere (inside quill)

### Additional context

https://github.com/ynput/ayon-frontend/assets/49156310/1139919b-2e52-4ce8-bf17-ad924f09aa96

### DIFF
--- a/src/components/CommentInput/CommentInput.jsx
+++ b/src/components/CommentInput/CommentInput.jsx
@@ -24,6 +24,20 @@ import { useGetTeamsQuery } from '/src/services/team/getTeams'
 import { useSelector } from 'react-redux'
 import { getModules, quillFormats } from './modules'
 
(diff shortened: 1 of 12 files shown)


BigRoy (approved):
Just going off of the video preview I really like this ❤️

martastain (approved):
Works really well. I noticed that user hover does not show the fullname even though it is used in the link already, but i assume this will change slightly with the new suggestion resolver anyways.

PR #494, opened 2024-06-03
Attributes: Min and max values

Innders:
## Description of changes

Instead of `gt, ge, lt, le` values use `min` and `max`.

When `type=float` give the option to toggle `>, ≥, <, ≤`.

### Technical details

The returned attribute can not have both `ge` and `gt` at the same time (same for `lt` and `le`).

### Additional context

image


BigRoy:
Could you share a screenshot of the tooltip? Is the "Currently" value the the example value or the default value?

Innders:
> share a screenshot of the tooltip

image

> And I assume this also changes/fixes the UX for the integer attributes?

Integer is the same except those buttons are hidden because it will always be ≥ or ≤

BigRoy:
> ≥ or ≤

Are you making smileys!?
Anyway, visually looks good to me.

PR #509, opened 2024-06-07
Tasks: thumbnail does not always use latest version

Innders:
## Description of changes

If no task thumbnail is available then it should use the next latest version thumbnail.

### Technical details

I thought versions came sorted from the query but I've added in another sort to be extra sort it's sorting from highest to lowest.

If there is a HERO version with a thumbnail, then that will always be used. 🦸


Innders:
@BigRoy Could you please let me know if this doesn't work and we can debug it further.

PR #522, opened 2024-06-12
Services: data fetching rewrite

Innders:
## Description of changes

Better error messages.

### Technical details

- Migrated all queries and mutations to RTK Query.
- Migrated new service dropdowns to ARC dropdowns.

### DIFF
--- a/src/pages/ServicesPage/NewServiceDialog.jsx
+++ b/src/pages/ServicesPage/NewServiceDialog.jsx
@@ -1,8 +1,7 @@
 import { useState, useMemo } from 'react'
 import { toast } from 'react-toastify'
 import { rcompare } from 'semver'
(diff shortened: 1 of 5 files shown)


martastain (changes requested):
Originally setInterval was used to periodically load service data - especially status. This is no longer available, so service status is not updated and it is necessary to reload the page. 

When you starting or stopping a service, you should get a feedback whether service actually started/stopped.

setInterval is probably not the greatest way of doing that, but it worked. Maybe service heartbeat could rather dispatch a fire-and-forget service.heartbeat message with RT updates?

Innders:
`  const { data: services = [] } = useGetServicesQuery(undefined, {
    pollingInterval: 2000,
  })`

image

martastain (approved):
works like a charm

PR #535, opened 2024-06-18
Tasks: Show all tasks and select all users

Innders:
## Changes description

- Select/deselect all users
- This will show all tasks that have any user assigned
- Showing all tasks regardless of assignee is out the scope of this PR but lays the groundwork for it

image


martastain (approved):
Works fine, but timeouts with extreme numbers of tasks. that should be improved using new Kanban resolver

PR #539, opened 2024-06-19
Browser: you can now properly filter versions with multiple folders selected.

Innders:
## Description of changes

Before: selecting multiple folders in the browser and then filtering using "search" or "task types" was broken.
After: now you can properly filter even when the versions have been grouped.

Also fixes some type errors that would cause the page to crash.

### Technical details

Before it was incorrectly filtering **after** the grouping, now it does the filtering before the grouping.

### Testing
1. Go to browser
2. Select multiple folders
3. Search for something like "approved"
4. Filter by task type

image


martastain (approved):
works. no problems

PR #549, opened 2024-06-21
Settings: Handle null/undefined in useGetAddonSettingsListQuery

martastain:


src/services/addonSettings.js
```diff
@@ -22,11 +22,22 @@ const apiSuffix = (projectName, siteId, variant, asVersion) => {
...
     getAddonSettingsList: build.query({
-      query: ({ variant, projectName, siteId }) => ({
-        url: `/api/settings`,
-        method: 'GET',
-        params: { variant, project_name: projectName, site_id: siteId, summary: true },
-      }),
+      query: ({ variant, projectName, siteId }) => {
+
+        // this should prevent passing null/undfiend values to the query
+        // params once for all (until we have typescript)
```
BigRoy, line 28:
```suggestion
        // this should prevent passing null/undefined values to the query
        // params once and for all (until we have typescript)
```

src/services/addonSettings.js
```diff
@@ -22,11 +22,22 @@ const apiSuffix = (projectName, siteId, variant, asVersion) => {
...
+        const params = {}
+        if (variant) params.variant = variant
+        if (projectName) params.project_name = projectName
+        if (siteId) params.site_id = siteId
```
BigRoy, line 32: Funny how `params` is `const` but you can add stuff to it? Oh, funny javascript. :)

Innders: But this would cause an error, it's the wild west of languages. 🤠 
```js
const params = {}

params = {}
```

PR #579, opened 2024-06-28
Feed: Show new comments in real time

Innders:


src/context/websocketContext.jsx
```diff
@@ -96,7 +96,7 @@ export const SocketProvider = (props) => {
 
       if (topic === 'server.restart_requested') setServerRestartingVisible(true)
 
-      if (sender === window.senderId) {
+      if (sender === window.senderId || !sender) {
```
martastain, line 99: You cannot ignore messages without a sender. When server sends message by itself, they don't have a sender ID. This breaks progress bars in bootstrap for example.

martastain (changes requested):

src/containers/WatchActivities.tsx
```diff
@@ -0,0 +1,85 @@
...
+  const dispatch = useDispatch()
+
+  const store = useStore()
+
+  //   subscribe to inbox.message topic
+  usePubSub(
+    'activity',
+    (topic: string, message: ActivityMessage) => {
+      const references = message.summary?.references
+      const entityIds = references?.map((reference) => reference.entity_id) || []
```
Anirudhsoni101295, line 35: I think we should these two lines in else condition because we are not using this in if condition, so why to perform map function for if condition, it could be a expensive operation, so can we please move these two lines in else condition?

src/context/websocketContext.jsx
Innders: Removed. Sometimes `activity.created` will have sender=null? 
image

src/containers/WatchActivities.tsx
Innders: Agree from a readability and formating perspective so I have moved.

It would have negligible affect on performance as references would only ever be 1-20ish items. Maybe if it was 1000+ plus items.

Anirudhsoni101295: Oh understood, I thought references could be 1000+

src/context/websocketContext.jsx
martastain: Yep... In this case sender id was lost (activity is created by a event hook, not a user directly. would it make sense to pass  the sender back? since null is evaluated as "different sender, so I'll invalidate my cache", the worst case scenario is an unnecessary data loading?

martastain (approved):

src/context/websocketContext.jsx
Innders: Cool, I was just checking it wasn't an issue with BE. This actually fixes an issue where the status change activity wouldn't show, but now it does. I might try to get the single activity and patch it into the cache, instead of invalidating and fetching 30 again.

PR #580, opened 2024-06-30
Projects: show and manage pinned projects in main projects list

Innders:
## Description of changes

- Pinned projects will be sorted to the top and have a pin icon.
- Right click on a project/s to pin/unpin.

### Additional context

https://github.com/ynput/ayon-frontend/assets/49156310/9dd1e7a2-cbf2-4d33-9e7f-12beb0427f98

### DIFF
--- a/src/components/CollapseButton.jsx
+++ b/src/components/CollapseButton.jsx
@@ -38,7 +38,7 @@ const StyledChevron = styled(Button)`
   }
 `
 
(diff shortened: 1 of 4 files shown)


martastain:
Could we use something like `pinnedProjects` here? We might want to allow pinning different things in the future :)

```
"frontendPreferences": {"pinned": ["asdf", "AY_CG_demo"],
```

PR #582, opened 2024-07-01
RTK Query: use the same slice for all APIs

Innders:
This does some crazy stuff to combine all the generated rest, graphql and legacy queries into one api slice.

It means you can call `api` from anywhere and have access to all endpoints.


martastain:
Sign-out does not redirect you to the login page

martastain:
Not sure if it is related, but downloading icon won't clear even if the installation is finished. Tried 2x, once it show the restart banner (without clearing download indicator), the second time, it didn't show the restart banner at all.

image

martastain:
Bootstrap does not show release packages. Current `dev` image works fine. No errors in the console

image

PR #585, opened 2024-07-02
Comment: many bug fixes

Innders:
## Description of changes and testing

- [x] Checklists in editor, multiple lines for one item would cause alignment issues.
- [x] Checklists in comment, multiple lines for one item would cause alignment issues.
- [ ] Blockquote: add quote line to the left for quotes.
- [x] Blockquote: maintain quote when editing.
- [x] Hovering over user mentions now always shows tooltip.


martastain:
Othewise everything works really good.

PR #588, opened 2024-07-03
Feed: Show checklists count

Innders:
## Description of changes

If an entity has some checklist comments, show the figure in the filter button

image

## TODO

Maybe there could be a dedicated endpoint for this? The graphql query is a little hacky.


BigRoy:
Of course everyone is going to start asking: Does this live update when chatting and checking boxes?

PR #621, opened 2024-07-17
fix(user settings): Only admins can be developers, managers cannot se…

### LINKED ISSUE #596: Dev Mode: Only admins can be developers
`/settings/users`

- [x] No one should have the option to make managers developers. Only admins can be developers.
- [x] Managers should not have the option to make admins developers. Only admins can set developers.

image

prodigel:
Disabling switch if the following conditions are not fulfilled, updating tooltip accordingly:
- Managers cannot set admins as developers
- Only admins can be set as developers

closes #596


Innders (approved):
Awesome, works really nicely!

PR #636, opened 2024-07-19
Users: Custom form for a adding service user

flynput:
Existing non service users cannot be converted to Service Users and vice versa. The following actions were taken to avoid that:
- Custom form for service user creation, only username, active/inactive & generated key form fields are required
- Non service users cannot be converted to Service users, option removed from form

### DIFF
--- a/src/containers/Feed/hooks/useUserMutations.js
+++ b/src/containers/Feed/hooks/useUserMutations.js
@@ -1,0 +1,42 @@
+import { useEffect, useState } from 'react'
+
+const useUserMutations = ({}) => {
+  const initFormData = {
+    userLevel: 'user',
+    userActive: true,
+    UserImage: '',
+    isGuest: false,
+    accessGroups: {},
+    defaultAccessGroups: [],
+  }
+
+  const initialFormDataCallback = () => initFormData
+
+  const [addedUsers, setAddedUsers] = useState([])
+  const [password, setPassword] = useState('')
+  const [passwordConfirm, setPasswordConfirm] = useState('')
+  const [formData, setFormData] = useState(initialFormDataCallback)
+
+  useEffect(() => {
+    // set initial form data
+    setFormData(initialFormDataCallback())
+  }, [])
+
+  const resetFormData = ({ password, passwordConfirm, formData, addedUsers }) => {
+    // keep reusable data in the form
+    setPassword(password)
+    setPasswordConfirm(passwordConfirm)
+    setFormData(formData != undefined ? formData : initialFormDataCallback)
+    setAddedUsers(addedUsers)
+  }
+
+  return {
+    password, setPassword,
+    passwordConfirm, setPasswordConfirm,
+    formData, setFormData,
+    addedUsers,
+    resetFormData,
+  }
+}
+
+export default useUserMutations
--- a/src/containers/Feed/hooks/useUserMutations.js
+++ b/src/containers/Feed/hooks/useUserMutations.js
@@ -1,6 +1,6 @@
 import { useEffect, useState } from 'react'
 
-const useUserMutations = ({}) => {
+const useUserMutations = () => {
   const initFormData = {
     userLevel: 'user',
     userActive: true,
@@ -14,6 +14,7 @@ const useUserMutations = ({}) => {
 
   const [addedUsers, setAddedUsers] = useState([])
   const [password, setPassword] = useState('')
+  const [apiKey, setApiKey] = useState('')
   const [passwordConfirm, setPasswordConfirm] = useState('')
   const [formData, setFormData] = useState(initialFormDataCallback)
 
@@ -33,6 +34,7 @@ const useUserMutations = ({}) => {
   return {
     password, setPassword,
     passwordConfirm, setPasswordConfirm,
+    apiKey, setApiKey,
     formData, setFormData,
     addedUsers,
     resetFormData,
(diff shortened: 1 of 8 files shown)


Innders (changes requested):
Getting there but we just need a few UI changes.

### For a service user remove all the unnecessary fields.

image

It should look something like this:
image

- label width `160px`
- API key padding `var(--padding-l)`
- API key background color `var(--md-sys-color-surface-container-high)`

### For a normal user remove the service key fields

image

### Suggestion

Write two completely separate forms for **normal user** and **service user** whilst sharing some logic like submitting.

Thanks!

Innders:
Also small note, I just pushed a commit to switch the buttons around :)

BigRoy:
> Getting there but we just need a few UI changes.
> ### For a service user remove all the unnecessary fields.
> 
> image
> 
> It should look something like this: image
> 
>     * label width `160px`
> 
>     * API key padding `var(--padding-l)`
> 
>     * API key background color `var(--md-sys-color-surface-container-high)`
> 
> 
> ### For a normal user remove the service key fields
> 
> image
> ### Suggestion
> 
> Write two completely separate forms for **normal user** and **service user** whilst sharing some logic like submitting.
> 
> Thanks!

Question. Under the separator/divider should that **Access Control** label be like the header of that area - similar in size to the "Create new user" header? It feels a bit uesless at the same size with no attribute/control next to it. Or we just remove the label?

Innders:
> > image
> 
> Question. Under the separator/divider should that **Access Control** label be like the header of that area - similar in size to the "Create new user" header? It feels a bit uesless at the same size with no attribute/control next to it. Or we just remove the label?

Agreed, I vote just remove

martastain (changes requested):
When a service user is created, its API key does not work. When regenerated from the user editor, it works correctly. I can see you are handling "password" somehow, but API keys are stored differently: 

- create user without any password
- call `/api/users/{username}/password` with "apiKey" field set

that should ensure "apiKeyPreview" will be available in the user entity.

src/pages/SettingsPage/UsersSettings/newServiceUser.jsx
```diff
@@ -0,0 +1,180 @@
...
+  const preparePayload = (formData, attributes, password) => {
+    const payload = {
+      data: {
+        isService: true,
+        defaultAccessGroups: formData.defaultAccessGroups || [],
+        accessGroups: formData.accessGroups || {},
+      },
+      attrib: {},
+      name: formData.Username,
+      password: password ? password : undefined,
```
Innders, line 61: I think most of these fields can be removed now as they don't apply to service users?

Password needs to be apiKey instead.

image

Innders (changes requested):
Ensure PascalCase is used file names and component names.

src/pages/SettingsPage/UsersSettings/newServiceUser.jsx
```diff
@@ -0,0 +1,171 @@
...
+
+  const handleSubmit = async (close) => {
+    const validationResult = validateFormData(formData);
+    if (validationResult !== null) {
+      toast.error(validationResult)
+      return
+    }
+
+    try {
+      await addUser({ name: formData.Username, user: preparePayload(formData, attributes, password) }).unwrap()
```
Innders, line 64: This needs to use `apiKey` instead of `password` as per docs: https://ayon.dev/docs#tag/Users/operation/create_user

src/containers/Feed/hooks/useUserMutations.js
```diff
@@ -0,0 +1,44 @@
...
+    defaultAccessGroups: [],
+  }
+
+  const initialFormDataCallback = () => initFormData
+
+  const [addedUsers, setAddedUsers] = useState([])
+  const [password, setPassword] = useState('')
+  const [apiKey, setApiKey] = useState('')
+  const [passwordConfirm, setPasswordConfirm] = useState('')
+  const [formData, setFormData] = useState(initialFormDataCallback)
```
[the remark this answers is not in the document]
prodigel: That's how it was in the newUser component, no idea why, removing it

src/pages/SettingsPage/UsersSettings/newServiceUser.jsx
```diff
@@ -0,0 +1,176 @@
...
+
+  if (!open) return null
+
+  return (
+    <Dialog
+      onKeyDown={(e) => callbackOnKeyDown(e, {
+        validationPassed: formData.Username && password,
+        callback: handleSubmit,
+      })}
+      isOpen
```
Innders, line 120: I was trying to work out why cmd+enter wasn't working and realised that it's checking for a password. There is no password.
```suggestion
          validationPassed: formData.Username,
```

src/pages/SettingsPage/UsersSettings/newServiceUser.jsx
```diff
@@ -0,0 +1,176 @@
...
+        <UserAttribForm
+          formData={formData}
+          setFormData={setFormData}
+          attributes={[{
+            name: 'Username',
+            data: { title: 'Username' },
+            input: { placeholder: 'No spaces allowed', autoFocus: true, ref: usernameRef },
+          }]}
+          customFormRow={FormRowStyled}
+          {...{ password, setPassword }}
```
Innders, line 158: I guess this and any other reference to password can be removed?
```suggestion
```

Innders (changes requested):
Almost everything works 🥳 

Just one thing left now thanks

PR #654, opened 2024-07-23
Users: Fixed renaming dialog username

### LINKED ISSUE #594: User settings: Set user name [Object Object]
### To reproduce

1. Right click on a user **not** in the selection
2. Click on "Set user" or "Set password"

## Environment

- **Server version**: 
- **Server host OS**: 
- **Browser**: 
- **Client version**: 

### Additional context

https://github.com/ynput/ayon-frontend/assets/49156310/54fbe3b2-d6eb-4530-ba4a-f2c436328bed

flynput:
Fixed bug where user name was wrongly displayed due to object/string management mismatch.

closes #594


src/pages/SettingsPage/UsersSettings/UserList.jsx
```diff
@@ -29,10 +29,21 @@ const UserList = ({
...
     [selectedUsers, selectedProjects, userList],
   )
 
+  const onContextMenu = (e) => {
+    if (selectedUsers.includes(e.data.name)) {
+      onSelectUsers([e.data.name, ...selectedUsers])
+    } else {
+      onSelectUsers([e.data.name])
+    }
+    ctxMenuShow(e.originalEvent, ctxMenuItems(selectedUsers))
```
Innders, line 44: The new selection can either be the current `selectedUsers` **or** `[e.data.name]`. This only ever passes `selectedUsers` which could be wrong if `onSelectUsers([e.data.name])` is called.

Use a new constant `newSelectedUsers` that gets defined above based on the if statement.

src/pages/SettingsPage/UsersSettings/UserList.jsx
```diff
@@ -29,10 +29,21 @@ const UserList = ({
...
-    () => userList.filter((user) => selectedUsers.includes(user.name)),
+    () => {
+      return userList.filter((user) => selectedUsers.includes(user.name))
+    },
     [selectedUsers, selectedProjects, userList],
   )
 
+  const onContextMenu = (e) => {
+    if (selectedUsers.includes(e.data.name)) {
+      onSelectUsers([e.data.name, ...selectedUsers])
```
Innders, line 40: This if statement checks if the right click is within the selection (you are right clicking something already selected). If this is true then we want to do nothing because the selection should not change. This code duplicates the selection state with the same users over and over again on evert right click, look at the console that is logging `selectedUsers`.


https://github.com/user-attachments/assets/10c56248-1cd0-420e-89a6-a70726f05e12

src/pages/SettingsPage/UsersSettings/UserList.jsx
```diff
@@ -41,8 +52,8 @@ const UserList = ({
   }
 
   // IDEA: Can these go into the details panel as well?
-  const ctxMenuTableItems = useMemo(
-    () => [
+  const ctxMenuItems = () => {
```
Innders, line 55: The selection is passed to ctxMenuItems on like 44 but it's never used here.

`onDelete` has a bug where selecting one user and then right clicking on another and then deleting will delete the old selection and not the new one.

See how user35 is the one being deleted even though we right clicked on user33.

To fix this pass the newSelection to `onDelete` and use that selection argument on UserSettings.jsx onDelete function.


https://github.com/user-attachments/assets/a72140fa-dabb-426e-9d30-d8baea78545c

Innders (changes requested):
Getting there :)

This context menu selection out of sync has always been a tricky one.

Innders (approved):
Works great thanks!

Small note: seems your prettier config is still different, let me know if you need our config.

PR #656, opened 2024-07-23
Editor: Naming consistency

Innders:
## Description of changes 
Ensure that capital letters are preserved when parsing label to name.

## Technical details

- Removes leading and trials all special characters
- Replaces all special characters with `_`


MustafaJafar:
Currently, My AYON setup doesn't allow me to test it.
Can we push it to maybe experimental dev ?

MustafaJafar (approved):
Thanks, it works. 
Here are some test runs: 
image
image
image
image
image
image

PR #667, opened 2024-07-24
Settings: Scroll to the correct field using URI paste

martastain:


Innders (changes requested):
If you come from a different page the scrolling doesn't work.

For example use this URI from a page other than settings.

`ayon+settings://applications/applications/equalizer/variants/0/arguments/windows`

Innders (approved):
Very powerful! Now we just need this for everywhere in the app.

PR #670, opened 2024-07-24
Browser: keyboard changes selection

Innders:
## Description of changes 
You can now use the up and down arrows to change the selection.

## Technical details

Before it would only focus the table row and not select it.

## Additional context

https://github.com/user-attachments/assets/006feb2e-0749-4264-8484-a8413e1f919f

## Testing

1. http://localhost:3000/projects/demo_Commercial/browser
2. Test each table, folders, tasks, products.
3. Using arrows keys should change selection and update URI.
4. Holding down shift should select multiple.

### DIFF
--- a/src/containers/Feed/hooks/useTableKeyboardNavigation.ts
+++ b/src/containers/Feed/hooks/useTableKeyboardNavigation.ts
@@ -7,12 +7,36 @@ export const extractIdFromClassList = (classList: DOMTokenList): string | null =
   return idClass ? idClass.split('-')[1] : null
 }
 
+export const getMultiSelection = (
+  selection: { [key: string]: boolean },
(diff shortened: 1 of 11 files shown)


src/containers/Feed/hooks/useTableKeyboardNavigation.ts
```diff
@@ -0,0 +1,103 @@
...
+    }
+  }
+
+  return newSelection
+}
+
+// Type definitions for props
+interface UseTableNavigationProps {
+  tableRef: RefObject<TreeTable>
+  treeData: any[]
```
flynput, line 36: Is this a real any or a 'improve in the future' any?

flynput (approved):

PR #671, opened 2024-07-24
fix(addons): Sorting addon versions using semver library algorithm

### LINKED ISSUE #599: Addons sorting in addons tab
## Story
List of versions in Addon tab seems a bit random in default state. It is hard to compare them to currently used versions, and easily detect really old versions.

### Proposal
Sort addon versions using semver in Addons tab from newest to latest. Not sure what is current default order of versions but it seems to be random.

#### Current default state
image

#### Expected default state
Note that the versions sorting does not have to be enabled...
image

flynput:
Using semver comparison algorithm when sorting addons by version in settings -> addons -> version panel

closes #599

### DIFF
--- a/src/pages/SettingsPage/AddonsManager/AddonsManager.jsx
+++ b/src/pages/SettingsPage/AddonsManager/AddonsManager.jsx
@@ -1,4 +1,3 @@
-import compare from 'semver/functions/compare'
 import { Button, Section } from '@ynput/ayon-react-components'
 import { Splitter, SplitterPanel } from 'primereact/splitter'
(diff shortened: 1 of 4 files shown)


Innders (changes requested):
The sorting works great. Now we need to ensure the default sorting is also the same.

Currently the default sorting is based off status but it should be sorted by version.

You might need to create a reusable sorting function to be used for default sort (arrays of items going in) and table sort (what you have now).

https://github.com/user-attachments/assets/9d471055-d09f-41e8-a6b6-17e5698632e2

PR #674, opened 2024-07-26
Watchers: Get important notifications about any entity

Innders:
## Description of changes 
- Add yourself or others as a watcher on any entity.
- When watching, receive important notifications when there are new comments.
- Add/remove watchers to multiple entities at one time.

https://github.com/user-attachments/assets/4b5b55f7-5eb8-4abd-8a2e-9df0244e2b1a


flynput (approved):
The code looks good, unfortunately I can't test, some weird issues related to case-sensitivity on the 'Watchers'.

PR #675, opened 2024-07-26
Addons: Added senderId to iframe url to always refresh html content

flynput:
While testing addon changes in dev iframe doesn't reload content due to caching setup. PR forces new urls on every visit, making sure that iframe content is also refreshed.


martastain:
Yes. In production, it's called "version" and it is already a part of the path :) 
So refreshing is needed only when you are developing the addon and don't bump the version every time you save it.

Innders (approved):

PR #677, opened 2024-07-29
Comment: Cycling through attachments via keyboard or mouse

### LINKED ISSUE #604: Feed: cycling through attachments
## Story
As a user, when I'm looking at attachements from an activity comment, I want to be able to see the next attachement or previous attachement using my keyboard or my mouse.

flynput:
Description:
When previewing comment attachments one can navigate through the list via keyboard or mouse.

https://github.com/user-attachments/assets/36672f51-94db-4f43-8da3-5f43bbcb8c6f

closes #604

### DIFF
--- a/src/components/Feed/ActivityComment/ActivityComment.jsx
+++ b/src/components/Feed/ActivityComment/ActivityComment.jsx
@@ -162,6 +162,7 @@ const ActivityComment = ({
               <FilesGrid
                 files={files}
                 isCompact={files.length > 6}
+                activityId={activityId}
                 projectName={projectName}
                 isDownloadable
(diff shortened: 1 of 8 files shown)


Innders (changes requested):
I love it! This is so cool 🥳 I think we could push it one step further and do better than everyone else. Could it navigate between **all** attachments, it's not just limited to the one comment. That would make it really easy to explore all the attachments.

image

What do you think?

- [ ] Navigate across all attachments
- [ ] Escape button shortcut to close dialog

Innders (approved):
Just tested it again and it works like a charm. Very nice feature.

PR #678, opened 2024-07-29
589 bundles re check of compatibility on path change

### LINKED ISSUE #589: Bundles: Re-check of compatibility on path change
## Story
In dev bundles it is possible to define custom path for client code. Each change of the path field is considered as addon change and combination of addons is re-validated for compatibility.
### To reproduce
1. Create dev bundle.
2. Enable custom path on a addon.
3. Start to fill path manually.
4. The label showing validation should start flickering as you type.

### Expected behavior
The path field does not affect in any shape or form the compatibility so the check can be skipped (the same for the checkbox that enables the path).

flynput:
When adding a new bundle, check requests are skipped on add-on path changes. The rest of form changes trigger API checks as expected.

https://github.com/user-attachments/assets/2d3e82ec-f7b1-44a1-be96-305764bbe6fc

closes #589

### DIFF
--- a/src/hooks/usePrevious.js
+++ b/src/hooks/usePrevious.js
@@ -1,0 +1,12 @@
+import { useRef, useEffect } from 'react'
+
+function usePrevious(value) {
+  const ref = useRef()
+  useEffect(() => {
+    ref.current = value
+  })
+
+  return ref.current
+}
+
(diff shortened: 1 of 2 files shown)


Innders (approved):
Looks good.

Issue #653, opened 2024-07-22
Activity Feed: Added assignees should be listed by nice name


BigRoy:
## Story

I want to see the artist's nice name (full name) for the added/removals in the assignees in activity feed (right hand side here:)

![image](https://github.com/user-attachments/assets/e2dda668-2cd5-4ae6-9a2e-802aa46a96c9)

### Problems

Consistency in user names.

### Proposal

Make them match the user's full name. (maybe even allow hovering to see their cards to see who they are, etc.)

## TODO

- [ ] Use nice names for add/removal assignees right hand side entries

PR #689, opened 2024-07-31
Activities: collapsing add/remove activities for same user

### LINKED ISSUE #652: Activity Feed: Merge addition/removal of assignees
## Story

Adding/removing multiple assignees at once should display as one activity feed entry to not pollute the activity feed.

### Problems

Prettier. No tearing out my eyeballs. No pulling out my hairs.
There may be tears though. Just a few, of joy.

### Proposal

Current:
image

Maybe match Github:
image

## TODO

- [ ] Merge assignees activities

flynput:
Merging multiple add/remove changes in activity timeline into 1 entry when the user/action is the same

https://github.com/user-attachments/assets/76d91bbe-4d6d-440d-8742-55d53530e1df

Closes #652 

## Description of changes

### DIFF
--- a/src/components/Feed/ActivityAssigneeChange/ActivityAssigneeChange.styled.js
+++ b/src/components/Feed/ActivityAssigneeChange/ActivityAssigneeChange.styled.js
@@ -17,8 +17,8 @@ export const StatusChange = styled.li`
 
 export const Body = styled.div`
   display: flex;
-  flex-wrap: wrap;
-  align-items: center;
+  flex-wrap: nowrap;
+  align-items: start;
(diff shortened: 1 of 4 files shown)


Innders (changes requested):
Really nice, just a couple things i've noticed. 👍 

- [ ] What is the order of the users? I would expect it to by activity createdAt, with earliest coming first.
- [ ] Does it make sense to solve this as well? - #653 

- [ ] We have lost the grouping of activities.
Current
image

Before
image

Innders (approved):
Looks good to me! 😎

PR #692, opened 2024-07-31
Details panel: Click on path to open slide out

Innders:
## Description of changes

- Click part of the path to open it in the slide out
- Right click to open the context menu (copy ID etc.)
- If there are more than 4 path items, the starting items will be collapsed into a `...` menu.
- Versions will show sibling versions dropdown.

https://github.com/user-attachments/assets/39760390-b1a6-42b8-990e-19c242c4fb57


Innders:
Fixes #692

BigRoy:
Nice! I did get a _bit_ confused by the slide-out sometimes being two lines instead of just the one and then missing that the 'last entry' is actually on the line below it. I wonder if we should force it to be on one line but they may be good reasons not do to that to ensure the right type of "access" to the breadcrumbs?

Innders:
> I wonder if we should force it to be on one line

I wondered this too... one solution would be to truncate the path more. It's a little tricky as I would need to calculate the width of the rendered path and then change it if it was too wide for it's parent.

BigRoy:
Not saying it's a good idea - but it may provide more space and consistency if the last entry is **always** on the second line and also maybe lighter in color so you know that is the actual entity you're looking at.

I feel stupid saying that because I think it'll look odd as soon as there's a very short path - but just wanted to throw it out there.

martastain (approved):
Works great!

Innders:
> Not saying it's a good idea

I cried multiple times getting this to work and nearly gave up at least 10 times.

The path will be truncated down to always be one line. 

I'm not exactly sure how it works so don't ask me. I mainly bashed my face against the keyboard for a while until some miracle happened.


https://github.com/user-attachments/assets/d7b5b1c3-e7b8-4310-b722-c7383e7ea166

BigRoy:
Looks like wizardry - quite nice!

martastain (approved):

PR #700, opened 2024-08-02
Activities: feed cannot identify status changes for multiple selected folders/tasks/versions

### LINKED ISSUE #665: Feed: Cannot identify status changes for multiple selected folders/tasks/versions
There's folders selected, which status changes belongs to which folder?
**Current**
image

**Target**
image

- [ ] When activity is **not** origin as `(label || name)` in brackets at the end.
- [ ] Versions suffixes add the product name and version `(productName - version)` `(layoutMain - v002)`
- [ ] Activities from all from different entities are getting "merged". Only merge activities of same entity (id).

**Pending Approval**
- [ ] Only ever show "origin" status activities. (Folder doesn't show product/version/tasks statuses).

flynput:
Showing detailed version info in status change activities depending on selection context
* displaying folder/task name on activity status change items when focused 
* product name included on version selection
* folder name included if multiple selected
* reverted change  that introduced bug which removed activity status icons

https://github.com/user-attachments/assets/f55c1513-42c1-4224-9cfe-3467d2942e60

closed #665

### DIFF
--- a/src/components/Feed/ActivityStatusChange/hooks/getContextParents.js
+++ b/src/components/Feed/ActivityStatusChange/hooks/getContextParents.js
@@ -13,7 +13,7 @@ const useGetContextParents = (activity, entityType) => {
 
   if (focusedFolders?.length > 1 && !tagTypes.folder) {
     tagTypes.folder = true
-    tag.push(activity.activityData.parents.find((p) => p.type == 'folder')?.name)
+    tag.push(activity.activityData.parents?.find((p) => p.type == 'folder')?.name)
   }
 
   if (activity.activityData.origin.type == 'task' && !tagTypes.task) {
@@ -29,7 +29,7 @@ const useGetContextParents = (activity, entityType) => {
   if (entityType == 'version' || activity.activityData.origin.type == 'version') {
     if (!tagTypes.product) {
       tagTypes.product = true
-      tag.push(activity.activityData.parents.find((p) => p.type == 'product')?.name)
+      tag.push(activity.activityData.parents?.find((p) => p.type == 'product')?.name)
     }
     if (!tagTypes.version) {
       tagTypes.version = true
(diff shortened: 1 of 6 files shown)


BigRoy:
Even though it's better (You can see what it's about) I wonder whether it's clean enough.

1. It feels somewhat disconnected from the status change text above it. It feels like a new line instead of like a subtitle or connected header/label.
2. I wonder if we need some consistency to still use ` / ` as separator like in other areas or whether ` | ` is better in this case and whether we can somehow get it clear what it's linked to (point 1) and then avoid the `( )` brackets surrounding it to avoid clutter.

Other than that, it's somewhat of an improvement so we could save the cosmetics for later if that's preferred.

src/components/Feed/ActivityStatusChange/hooks/getContextParents.js
```diff
@@ -0,0 +1,44 @@
...
+
+
+  if (activity.activityData.origin.type == 'folder' && !tagTypes.folder) {
+    tagTypes.folder = true
+    tag.push(activity.activityData.origin.name)
+  }
+
+  if (focusedFolders?.length > 1 && !tagTypes.folder) {
+    tagTypes.folder = true
+    tag.push(activity.activityData.parents.find((p) => p.type == 'folder')?.name)
```
Innders, line 16: I managed to create a type error by selecting multiple folders and then selecting a task.

image
image

Innders (changes requested):
Looks good. We may tweak it in the future based on user feedback :)

Just one error to cleanup

Innders:
> 1. feels somewhat disconnected

I get your feeling. I've played around with it, what do you think of this @flynput @BigRoy ?

1. Removed `status changed` -  do we really need this text, is not obvious enough?
2. Moved entity name to after the user and replaced `|` with `/` as it's more a path.

I think this reads better. **User that did it** - **on something** - **and that change**

New format
image

Old format
image

Innders (approved):
💯

PR #701, opened 2024-08-02
Chore: PrimeReact Upgrade

Innders:
## Description of changes 
- Upgrade to primereact `9.6.2`.
- Fixes to solve breaking changes.
- Watchers button inside button fixes.

## Technical details

PrimeReact context menu, completely breaks on `9.3.0` for no reason.
Best solution is to keep it on an older version `"primereact-context": "npm:primereact@9.2.3"`

## Testing
Test all tables everywhere and check nothing awful has broken.


martastain:
Not sure if it's related (but it doesn't happen in ayon:dev). Marking anatomy preset as primary creates a duplicate record that disappears when the page is reloaded.


image

Innders:
The new table component really didn't like duplicate name keys. I have reworked the design as I think it would quite confusing anyway.

image

flynput (approved):

PR #705, opened 2024-08-05
enhancement(users): Greyed out rows - icons included - for inactive u…

### LINKED ISSUE #647: Users: Grey out inactive users
When a user is inactive, the table row should be greyed out slightly. 

Try using `color: var(--md-sys-color-outline);` (I think it's that)

flynput:
Using grey shade for inactive users in the list, reducing user icon opacity for a similar effect 

https://github.com/user-attachments/assets/fbeb8c89-f434-4e3b-a80e-4f4e27a0f813

Closes #647


Innders (approved):
Works for me. Thanks.

PR #711, opened 2024-08-06
Anatomy Presets: Rework primary and default

Innders:
## Description of changes 
- Remove button at the top.
- Add "built in" table row at the top of the table.
- Other small fixes

## Additional context

https://github.com/user-attachments/assets/9db0cdc5-eab3-4c40-b3a6-40ae5b3548ab


martastain (changes requested):
For some reason, the logic is still broken. When i set a preset as primary, and open "create new project" dialog, the pre-selected value in anatomy presets is the first item of the list, not the primary preset itself. 

That worked in 1.3.1

martastain (approved):

PR #714, opened 2024-08-06
fix(addons) Using a link instead of button for external links section

### LINKED ISSUE #709: Market: Use `a` tags for links.
Market addon links should use `a` tag instead of onClick event.

https://github.com/ynput/ayon-frontend/blob/7466e8b4fac13d94dfac5ac937754c726c1e6cac/src/pages/MarketPage/AddonDetails/AddonDetails.jsx#L277

image

flynput:
Using links instead of buttons on Addon Market external links card

https://github.com/user-attachments/assets/f69493f5-ea44-4ec3-a268-31827a00cd2f

Closes #709

### DIFF
--- a/src/pages/MarketPage/AddonDetails/AddonDetails.jsx
+++ b/src/pages/MarketPage/AddonDetails/AddonDetails.jsx
@@ -271,14 +271,15 @@ const AddonDetails = ({ addon = {}, isLoading, onDownload, isUpdatingAll }) => {
                     label={capitalize(group.type)}
                   >
                     {group.links.map((link) => (
-                      <Styled.UseButton
+                      <Styled.ExternalLInk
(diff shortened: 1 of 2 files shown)


src/pages/MarketPage/AddonDetails/AddonDetails.jsx
```diff
@@ -271,14 +271,16 @@ const AddonDetails = ({ addon = {}, isLoading, onDownload, isUpdatingAll }) => {
                     label={capitalize(group.type)}
                   >
                     {group.links.map((link) => (
-                      <Styled.UseButton
+                      <Styled.ExternalLInk
+                        className="externalLink"
```
flynput, line 275: class is redundant, removing ...

PR #721, opened 2024-08-08
Refactor temporal milestone declaration

martastain:
### Context

As part of the ongoing efforts to maintain code accuracy and reflect real-world parameters, this PR updates the self-referential age-related comment. The relative temporal offset from the inception date has been adjusted to mirror the current chronological state.

### Changes

- Modified the age comment to accurately represent the present age.
- Enhanced the integrity of developer commentary for future reference.

### Impact

This change ensures that our codebase remains a true reflection of its maintainers' life stages, fostering an environment of transparency and honesty.

### Validation

- Review the updated comment for correct age representation.
- Ensure all relevant documentation reflects this milestone appropriately.

### DIFF
--- a/src/containers/CopySettings/CopySettingsNode.jsx
+++ b/src/containers/CopySettings/CopySettingsNode.jsx
@@ -379,7 +379,7 @@ const CopySettingsNode = ({
   )
 
   // is it a table? it is. So i'm using a table. don't judge me!
-  // i am almost 40 years old and i can use a table if i want to.
+  // i am 40 years old and i can use a table if i want to.
 
   const body = (
     <NodePanelBody>


antirotor (changes requested):
I strongly disagree with this code change as it critical freezes temporal related information directly to the code. What about next year? Will we hit 40Y problem?

BigRoy:
Love it.

> I strongly disagree with this code change as it critical freezes temporal related information directly to the code. What about next year? Will we hit 40Y problem?

You got to work in some of that job security. Shall I start and already reference this in a new issue in advance?

Innders (approved):

PR #727, opened 2024-08-14
fix(users) Showing 'none' access users at the bottom of the table whe…

### LINKED ISSUE #718: User Settings: Sorting by project access
### To reproduce

1. Go to user settings
2. Select a project in the left list
3. Uncheck "filter users by project access"
       a. You are now left with ALL users not just on selected project.
5. Sort by "Project access"
6. Users are sorted alphabetically so "none" is in the middle and not at the bottom.

### Expected behavior

Sorting should be by access level and not alphabetically.

Here is the order it should be:

1. Admin
2. Manager
3. User - has access through an access group
    a. Number of access groups. Highest first.
4. No access at all

### Additional context

image

flynput:
Users with no Project access are filtered out and appended to the sorted list when sorting by Project access, no matter the order.

https://github.com/user-attachments/assets/a236ce7a-05f2-4deb-90b0-6ebb5cf22e52

Closes #718

### DIFF
--- a/src/helpers/user.ts
+++ b/src/helpers/user.ts
@@ -1,0 +1,59 @@
+import { UserModel } from "@api/rest"
+import { ColumnSortEvent } from "primereact/column";
+
+type EnrichedUserModel = UserModel &  {
+  isAdmin: boolean;
+  isService: boolean;
+  isManager: boolean;
+  accessGroupList: string[];
+}
+
+const accessGroupsSortFunction = (sortEvent: ColumnSortEvent) => {
+  // Early return while data is still loading
+  if (sortEvent.data[0].isLoading) {
+    return sortEvent.data
+  }
+  const adminFilter = (user: EnrichedUserModel) => user.isAdmin && !user.isService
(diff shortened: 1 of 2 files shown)


Innders (changes requested):
Users with no access have gone to the bottom, nice!

Two things to do now:

- [ ] Inactive users have no access to even logging in so they should be last. (sorry I missed this)
- [ ] Better sorting of users with access described below.

The order should go (from top to bottom):

1. admin
2. manager
3. user with access (users with most access groups first)
4. service
5. users with access

image

flynput:
Updated sorting strategy as requested.

https://github.com/user-attachments/assets/e585fa58-ff85-4703-8ebb-753984c8be2a

Innders (approved):
Awesome!

Users with access don't quite sort by number of access groups but it's none essential for this.

image

And last thing to consider... this has become a pretty big function, is it worth breaking it out into its own file. Maybe in a helpers file or something? That would also allow for it to be written in TS maybe.

Innders (approved):

PR #730, opened 2024-08-15
Progress Task View

Innders:
progress-full

## Description of changes

- New page inside a project `/projects/{project_name}/tasks`
image

## Main Features

- Select a parent folder to see all children folders and tasks.
- Collapse parent folders
- See the number of folders and tasks in the parent folder
- Status colour chart of the total progress of the folder.
progress-full-full

- Multi cell selection
image

- Sortable columns
image

- Editable assignees and status
progress-editing

- Expanded cards
progress-expanded


BigRoy (approved):
Testing this I was unable to break it, so I'll approve it.

---

I did have these cosmetic notes:



Can we have it list the task name on hover even when the card is expanded @Innders ?
In this case I'm not sure which animation is which unless I open the sidebar (double click)

image

---

Cosmetics / nice-to-have (like follow up PR). But when hovering this section you get a tooltip about the column:

image

Would be nice if:

- The names match the color of that status
- Maybe also have the icon to the left. 
- The percentages would align to the right?

---

Cosmetics: When hovering a section that does not have entries the tooltip shows `NaN` (not a number).

image

Should we make that just 0% instead?

---

Also that tooltip is slightly annoying, it comes up quite fast and if you just go up with the cursor it stays around because your mouse is over it.... making it hard to e.g. toggle the column sorting or the "Expand all rows" button because there's a high likelihood the tooltip will get in your face.

image

EDIT: Actually, it's super annoying. 

---

When 'details' (sidebar) is open, the context menu still shows "Show details" which seemingly does nothing?

image

---

For whatever reason the "Done" column has less padding on the left?

image

martastain (approved):

flynput (approved):

PR #739, opened 2024-08-20
fix(taskList): adaptive width after render to avoid width jumping col…

### LINKED ISSUE #728: Tasks list: Jumping widths
### To reproduce

1. Go to `/dashboard/tasks?view=list` for the first time.
2. See how the task names can be out of line.
3. Select a task and they jump to be in line.

### Expected behavior

They should be all in line on first page load. The position is correct AFTER they jump.

### Additional context

https://github.com/user-attachments/assets/0588111b-18b5-4c42-bf2f-0a4864adcd1c

flynput:
Using MutationObserver to wait for the task list node tree to get populated before calculating min widths for task/folder columns. Without it, useEffect executes early setting min width values to 0.

https://github.com/user-attachments/assets/cb7c8bb0-360a-4a11-b168-57b302c6bccb

closes #728


Innders (changes requested):
Awesome, I've never used MutationObserver before but it works nicely here!

I'm going to be super picky here. There is a split second when you can see the items out of line.

task_name_out_of_line


I know the items must be rendered in order to calculate their width. Could this be done behind the loading shimmer so that jump is hidden. It would mean the loading shimmer doesn't disappear until the widths have been calculated.

Innders:
I've noticed another issue. Because the items are lazy loading to increase performance not all the widths are calculated. Sometimes you can scroll down and there will be a larger folder name and the widths will shift as you scroll.

flynput:
Updated PR, going for 200px width for task and folder columns with overflowing content hidden.


https://github.com/user-attachments/assets/251f6117-0f53-4d61-974d-e26ba005397e

Innders (approved):
Looks good thanks 😎

PR #742, opened 2024-08-20
fix(folders): Label takes precendence to name when sorting folders in…

### LINKED ISSUE #732: Editor: Sorting by Name ignores Label
## Story

I want to be able to see my folders A->Z or Z->A

### To reproduce

1. Got to /projects/{project_name}/editor
2. Create some folders/tasks
3. Change the labels so they do not match the original `name`
4. Now try sorting by name.
5. The labels will be ignored.

### Expected behavior

- Sort by label otherwise name.

### Additional context

image

flynput:
Default sorting uses node.data props, which defaulted to 'name', a new prop was added to prioritize label before name where possible

https://github.com/user-attachments/assets/f8b1a71b-f0df-4422-a8b2-6b538000755c

Closes #732


src/helpers/sortByHelpers.js
```diff
@@ -0,0 +1,19 @@
...
+    let x, y
+    if (a.data.label !== undefined || !b.data.label !== undefined) {
+      x = a.data.label
+      y = b.data.label
+    } else {
+      x = a.name
+      y = a.name.toLowerCase()
+    }
+
+    return x.toLowerCase().localeCompare(y.toLowerCase())
```
Innders, line 12: If a folder/task has no label I get this error.
image

src/pages/EditorPage/EditorPage.jsx
```diff
@@ -477,21 +477,24 @@ const EditorPage = () => {
         const node = {
           key: childId,
           name: rootData[childId].data?.name,
-          data: rootData[childId].data,
+          data: {
+            ...rootData[childId].data,
+            labelThanName: rootData[childId].data.label || rootData[childId].data.name,
```
Innders, line 482: Just small typo` labelThenName` 🙂

src/helpers/sortByHelpers.js
Innders: Maybe this file could be written in typescript, would that have prevented this?

Innders (changes requested):
Just a couple things to clean up but other than that good. Now we know that data is important for sorting 🙄

Innders (approved):
Looks good, thanks 😄

PR #744, opened 2024-08-21
fix(users): Automatically disabling developer rights when updating us…

### LINKED ISSUE #740: Developer mode: Normal users shouldn't have access to developer mode
Normal users can have developer mode enabled.
Most probably a bug, user was once an admin + developer and then was changed to a user. The developer switch should be turned off when setting to user automatically. 

Image

If the user is set to dev by accident it shouldn't show for the said user.

flynput:
Disabling developer attribute when updating users that are not admins.

https://github.com/user-attachments/assets/dd13ae4e-181c-4a0f-9d17-eb97fd119c58

Closes #740


Innders (approved):
Great

PR #746, opened 2024-08-21
fix(tasks): Allowing task filtering by user assignee name/full name also

### LINKED ISSUE #720: Tasks List: Search filter supports assignees
Searching should support assignees so I could search for "Luke" and it would match all tasks that have "Luke..." as an assignee.

### Additional context
image

flynput:
Allowing task filtering by assignee also. Using User name/Full name when filtering

https://github.com/user-attachments/assets/31962969-063f-4e51-bda1-c3d3ca38160d

Closes #720

### DIFF
--- a/src/pages/UserDashboardPage/util/getFilteredTasks.js
+++ b/src/pages/UserDashboardPage/util/getFilteredTasks.js
@@ -15,7 +15,8 @@ export const getFilteredTasks = (tasks = [], filter = '', projects = []) => {
     const taskStatus = task.status
     const taskType = task.taskType
     const taskPath = task.folderPath
-    const matchingStrings = [taskName, taskFolder, taskStatus, taskType, taskPath]
+    const taskAssignees = task.assigneesData.map(assignee => [assignee.name, assignee.attrib.fullName || ''].join(':')).join('::')
+    const matchingStrings = [taskName, taskFolder, taskStatus, taskType, taskPath, taskAssignees]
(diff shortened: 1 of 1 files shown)


src/pages/UserDashboardPage/util/getFilteredTasks.js
```diff
@@ -15,7 +15,8 @@ export const getFilteredTasks = (tasks = [], filter = '', projects = []) => {
     const taskStatus = task.status
     const taskType = task.taskType
     const taskPath = task.folderPath
-    const matchingStrings = [taskName, taskFolder, taskStatus, taskType, taskPath]
+    const taskAssignees = task.assigneesData.map(assignee => [assignee.name, assignee.attrib.fullName || ''].join(':')).join('::')
```
Innders, line 18: What's the reason for all the `.join(':')).join('::')`?

Could you not just spread the name and fullName into assignees? 

`const matchingStrings = [taskName, taskFolder, taskStatus, taskType, taskPath, ...taskAssignees]`

Innders, line 18: What's the reason for all the `.join(':')).join('::')`?

Could you not just spread the name and fullName into assignees?

`const matchingStrings = [taskName, taskFolder, taskStatus, taskType, taskPath, ...taskAssignees]`

Innders (approved):
Cool, it works for me :D

PR #750, opened 2024-08-23
Settings: Promote bundle including project settings

martastain:


src/containers/AddonSettings/AddonSettings.jsx
```diff
@@ -502,8 +502,8 @@ const AddonSettings = ({ projectName, showSites = false }) => {
           Are you sure you want to push <strong>{bundleName}</strong> to production?
         </p>
         <p>
-          This will mark the current staging bundle as production and copy all staging studio
-          settings to production as well.
+          This will mark the current staging bundle as production and copy all staging
+          settings (studio and projects) to production as well.
```
BigRoy, line 506: I just wonder whether being slightly more wordy about this may avoid some confusion.
```suggestion
          This will mark the current staging bundle as production and copy all staging
          studio settings and staging projects overrides to production as well.
```

So leaving this here for others to confirm/deny.

PR #752, opened 2024-08-26
Settings: Migrate settings by bundle

martastain:
## Description of changes 

Exposes a "Copy Settings" toggle upon creating a new bundle using a new dialog.

Also exposes "Copy Settings From..." feature on the bundles page that allows you to copy all studio settings and project settings overrides between two bundles.

### Testing notes

1. Validate that Studio Settings and Project Settings overrides copy as you'd expect
2. Validate the extra dialog on creating a new bundle is understandable and NOT an inconvenience.
3. Validate the "Copy settings from..." functionality on the Bundles page context menu is understandable and works as intended (as point 1).

## Additional context

Some test recordings:

https://github.com/user-attachments/assets/47e5ab12-254d-4178-8cf6-5557102cc638

_This shows the "Copy Settings" being off by default when creating the new bundle, however - this is currently enabled by default and was just an older recording_

https://github.com/user-attachments/assets/fde378fd-5b25-47d4-8967-ad96b0e8275c


BigRoy:
### Wrong default selection for copy settings on "Duplicate and edit" existing bundle

On a bundle that is currently only staging, but dev mode is enabled for my user, when clicking "Duplicate and edit" it initializes as "dev bundle" (which has always been the case); when disabling that dev bundle then creating the bundle defaults the dialog to:

image

Which in my case is both the WRONG source bundle and WRONG source variant - I would have expected it to default to from which bundle a I duplicated.

I duplicated from the selected entry here:

image

For whatever reason it took my Production bundle instead of the Staging bundle to copy from... and it took the Dev variant instead of the production variant?

### Copying bundle settings to new addon that has new settings

When copying settings from one addon version to another, when the target addon already has overrides to attributes that did not exist in the source addon - those overrides will be removed.

So say:

1. Addon 0.1: has two settings
2. Addon 0.2: has three settings

Copying from 0.1 to 0.2 will copy the two settings, and CLEAR any other overrides on Addon 0.2.
The question is: will this be the most sensible to users? Or should we preserve any overrides that do not exist in Addon 0.1 (e.g. not clearing any overrides on that third setting).

Non critical for now - but just wanted to mention that it _will_ drop/clear overrides in that case.
This currently means that copying doing a round-trip `0.2 -> 0.1 -> 0.2` the 0.2 settings would've lost some.

These 'skipped' keys I believe also are logged in the Event Viewer / server logs:

image

### Copy dialog -> create button should be copy button

image

This dialog's accept button should not be labeled "create" but "copy" or "apply".

---

👍 ✅ 

Other than that, this functionally seems to work. It copies over the Studio Settings **and** Project Settings using the copy settings in the Create Bundle dialog or the separate "Copy settings" dialog on the Bundles page (B+B).

I must admit - that each time when I wanted to do that copy settings or click it that I felt a bit worried since each time I didn't feel entirely sure what it would be doing and what would be the full extent of the changes that would occur. I can imagine anyone else also being 'scared' when clicking around and being unsure of what this might affect for their studio.

BigRoy:
_Copying some notes that I shared elsewhere:_

We may need some very clear wording in the documentation that we have a few levels of granularity to copying settings:

1. Copy settings from... on **bundles page** copies all studio settings and project overrides for that bundle
2. Copy settings from... on **studio settings** page copies only the studio settings between the bundles
3. Copy settings from... on **studio settings page on a particular addon** copies between studio settings for only a specific addon version.
4. Copy settings from... on **project settings** page copies only the project settings for the selected project between the bundles.
5. Copy settings from... on **project settings page on a particular addon** copies project overrides for only selected project for only specific addon version.

The "Copy Settings" toggle on creating a new bundle is basically that top one (number 1) described above - it will copy all studio settings and project setting overrides between the bundles.

That's a lot of different dialogs we have. And someone may be as stupid as me and definitely at one point will by accident be using the wrong page, using the wrong dialog unintentionally.

---

I feel that particular "Copy" dialog may need a dedicated label explaining that it copies BOTH studio settings and project setting overrides for ALL projects. To make it entirely clear that's what it'll end up doing, similar to what the "Promote bundle" dialog shows.

---

Something tells me "Copy settings" seems so 'default behavior' to everyone (it's exactly the reason why most clients reported the issue) that the tooltip or label on that should explain WHY that's an option, preferably linking to Bundles documentation that very clearly explains that:

- Addons have settings PER variant PER addon version.

Because that is the hardest concept to grasp that you're copying to variant for each addon version... and not to the bundle.

Innders:
> Copying from 0.1 to 0.2 will copy the two settings, and CLEAR any other overrides on Addon 0.2.

Surely the must assume that the "removed overrides" from 0.1 are intentional. It will be confusing when the settings don't match exactly 1 to 1 between the two once copied over. 

> This dialog's accept button should not be labeled "create" but "copy" or "apply".

I can understand why it is created because the new bundle hasn't been created yet. We will need to work out exactly the order and naming of these events. I would say the bundle is created and then the dialog pops up as an after action. So in that case you would be correct.

> We may need some very clear wording in the documentation that we have a few levels of granularity to copying settings

We should probably make it absolutely clear at the top of each dialog the exact copying that is taking place. It should also be reflected in the buttons you are clicking "Copy addons settings" "Copy all settings" "Copy project settings".

BigRoy:
> Surely the must assume that the "removed overrides" from 0.1 are intentional. It will be confusing when the settings don't match exactly 1 to 1 between the two once copied over.

Yup - it's mostly the round trip 0.1 -> 0.2 -> 0.1 _losing data_ which may be dangerous/confusing to some.

> I can understand why it is created because the new bundle hasn't been created yet. We will need to work out exactly the order and naming of these events. I would say the bundle is created and then the dialog pops up as an after action. So in that case you would be correct.

This is in the other dialog - that appears on Copy settings, not after the Create Bundle.

Innders:
Could we have a second round of testing please with the new design and slight changes in mechanics.

- Does not invoke when creating a new bundle.
- Invokes when setting an environment status for a bundle (staging, prod, dev).
- Preselects bundle based on what status was set. Setting a bundle to staging with default copy the settings from the current staging bundle.
- Removes the choice to select bundles that have **no** status.
- Removes choice of bundle variant and will default to the status of the bundle. A bundle with status staging will copy over the staging variant.
- Clean up general design.

BigRoy (changes requested):
So, I gave this a go. Very nice! :) I did have a question. I got for example this:

image

Which would be an incorrect left hand side, because it's unclear whether it's copying from Staging OR Production variant. It should display each of these as separate entries (or basically, all variants as separate entries)... and default to the one that matches the variant you're setting.

(Note also that left side the green button lacks margins to the left of it?)

Similarly when setting to Production a bundle that's currently in Staging... that Bundle's staging variant is not listed left hand side, but it should. Because that staging variant may have different settings than the Production variant you're settings.

image
(in this screenshots see my active bundles top left).

Once that's tweaked I guess it makes more sense to continue testing.

Also, I do wonder whether it'd be worth it to have a "cancel" button as well that allows you to _even_ cancel e.g. the setting of the variant "Production" itself... Because now clicking "Do not copy now" may still appear confusing that it did visibly change your active bundles. I mean, it makes sense - I just wonder if anyone would ever want to back out as soon as they see the dialog?

Innders:
- The status now gets set straight away.
- If the status is no longer on the source bundle show it crossed out.
- Update title and description copy.

image

tests/fixtures/bundlePage.ts
```diff
@@ -1,50 +1,89 @@
...
-    await this.goto()
-    await this.page.goto('/settings/bundles')
     await this.page.getByRole('button', { name: 'add Add Bundle' }).click()
     await this.page.getByRole('main').getByRole('textbox').fill(name)
     await this.page.getByRole('button', { name: 'Select an option...' }).click()
     const launcher = '1.0.2'
     await this.page.getByText(launcher).click()
+    // check that bundle is compatible
+    const compatibleMessage = 'Checks complete: Bundle is compatible'
+    await expect(this.page.getByText(compatibleMessage)).toBeVisible()
```
Innders, line 26: This line fails on the second bundle creation and I have no idea why. It seems that bundle title and launcher are not set correctly so compatibility checks fail. Any ideas @flynput ?

flynput: I'm still checking it. The 'anomaly' seems to be caused by a form reset, but I can't pinpoint yet what causes that.

BigRoy:
_Tested using `experimental` just now.

It now updates live in the background visually, to the new variant that you set and prompts me:

image

### Incorrectly marking other variant also as crossed out

When setting a bundle to Production **from** a bundle that has both Production and Staging the left list ALSO has `Staging` (correct!) **but** it also has it crossed out.

image

### Set as dev also lists the _now newly set_ dev bundle left hand side

When setting a bundle to dev the left hand side incorrectly also lists the newly set active variant on left hand side:

image

**Also side note:** When setting the new bundle as dev the old bundle isn't necessarily dev anymore... hence, the `dev` left side being "crossed through" is maybe odd or incorrect?

### Tooltip for stupid me?

image
Do we need a tooltip on the disabled menu entry saying something like:
"You can only copy settings to an active variant like production, staging or dev."

### Context menu -> Copy settings from... on production bundle

When using Copy Settings From on the production bundle in context menu I get this:

image

tests/fixtures/bundlePage.ts
flynput: @Innders Fixed the issue, please check the commit. The problem was caused by addons fetching which reset the form. I changed the logic a bit to keep the name/installer version when that happens.

Innders:
> When setting a bundle to Production from a bundle that has both Production and Staging the left list ALSO has Staging (correct!) but it also has it crossed out.
✅ Fixed!

>When setting a bundle to dev the left hand side incorrectly also lists the newly set active variant on left hand side:
✅ Fixed!

>the dev left side being "crossed through" is maybe odd or incorrect?
✅ Fixed!

>When using Copy Settings From on the production bundle in context menu I get this:
✅ Fixed!

PR #754, opened 2024-08-27
URI: Sync URI in the address URL (share context with URL)

### LINKED ISSUE #669: URI: Pass URI in URL to be resolved
## Story

I want to share a full link to a specific entity with someone.

### Problems
- The URI currently doesn't work on all pages, like profile, users etc. It only works with entity browser and settings.
- The current technical implementation is quite messy and could do with a rethink.

### Proposal

- [ ] URI bar has a second button to copy full link with URI attached.
- [ ] `?uri=...` query param that is resolved.

image

flynput:
Keeping ayon-entity URI in sync with the browser URL, so that users copy-pasting the URL can easily share their full context with others. Also added an edit icon to the left hand side of the URI component for a more intuitive experience.

https://github.com/user-attachments/assets/6f519293-e828-4397-b0b7-836b1766ec7a

Closes #669


Innders (changes requested):
- [ ] Get settings to work so it's a global uri= rather than entity specific.
- [x] Going between different versions on browser not working.

src/containers/Breadcrumbs/Breadcrumbs.jsx
```diff
@@ -133,6 +144,14 @@ const Breadcrumbs = () => {
   return (
     <Styled.Wrapper>
       <Styled.Crumbtainer>
+        {uriDisplay && localUri && (
+          <Button
```
Innders, line 148: Can this button but part of the breadcrumbs so when hovering both are highlighted at the same time.

image

src/containers/Breadcrumbs/Breadcrumbs.jsx
```diff
@@ -133,6 +144,14 @@ const Breadcrumbs = () => {
...
+            icon="edit"
+            style={{ opacity: editMode ? 0 : 1, width: editMode ? 0 : 'auto', borderRadius: '4px 0 0 4px'  }}
+            onClick={() => setEditMode(true)}
+            variant="tonal"
+          />
+        )}
         <Styled.CrumbsForm onSubmit={goThere}>
           <label data-value={inputValue}>
             <InputText
```
Innders, line 157: Could we please have a button at the end here so that it's more obvious that clicking/hitting enter will navigate to the URI. Maybe `arrow_forward` would work?

Also make sure there is border radius 4px on the left hand side of the input and right hand side of the new button.

image

Innders (changes requested):
Technically everything seems to work flawlessly 💯 There are just a couple CSS things to clean up.

There are situations where there is no copy button when there is no valid URI. In this case the right hand corners need rounding.

image
 
 When hovering the edit button the corners break.
 
https://github.com/user-attachments/assets/d4f4af02-78b8-4ade-bd76-9b82c31c1dce

Innders (approved):
Awesome, this is going to be such a useful feature!

PR #756, opened 2024-08-27
Bundles: Initial bundle selection is not influenced by bundleList changes

### LINKED ISSUE #755: Changing status bundle resets bundle list selection to production-or-first
Whenever you click "Set as staging" for a currently selected bundle, it sets it as staging but also directly selects the production bundle instead, which is counterintuitive

flynput:
Production bundle selected by default - or first if non is production is not affected by bundleList changes. This fixes the issue raised, where setting any other bundle to staging would reset the original selection

https://github.com/user-attachments/assets/b1bfb970-4cc4-46c4-ac57-f24ce4907602

Closes #755


Innders (approved):
👍

PR #757, opened 2024-08-28
Editor: Listing & form improvements

### LINKED ISSUE #708: Editor: Enum dropdown fixes and enhancements
Handling string attributes with enumerator has flaws as it is not used much. Here's a list of issues and possible enhancements, that should also cover implementation of https://github.com/ynput/ayon-backend/issues/313

- [ ] In the table, the value is shown instead of the label:

image

- [ ] Label for inherited value is not shown at all:

image

- [ ] "Clear to no value" does not make sense as the value always need to be one of the enum list or null. Currently it tries to set empty list (the same logic as for `list_of_string` attributes), which cause errors. IMO we could remove this button for single value dropdown completely.

image

Additional `color` and `icon` can be used with https://github.com/ynput/ayon-backend/pull/314 In that case, dropdowns could be rendered similarly to status dropdowns.

- [ ] Create an enhanced dropdown that supports icons and colored backgrounds.

Custom dropdown examples:
InstallerSelector: https://github.com/ynput/ayon-frontend/blob/2e6e030e67474ec3fe8a5b6477bf45aed5f33172/src/pages/SettingsPage/Bundles/InstallerSelector.jsx
VariantSelect: https://github.com/ynput/ayon-frontend/blob/b678ff3fd9dc4715f415e6a90be08bdc12c1f776/src/containers/AddonSettings/VariantSelector.jsx
StatusSelect: https://github.com/ynput/ayon-frontend/blob/f24b1a52ce5cab800942d2be2f8ced078021fe15/src/components/status/statusSelect.jsx

flynput:
"None" placeholder in editor listing when no tools are selected.
Refactoring formatAttributes and converted to TS.
Removed clear all from single-select enum fields in Editor details panel.

Closes #708

### DIFF
--- a/src/pages/EditorPage/EditorPanel.jsx
+++ b/src/pages/EditorPage/EditorPanel.jsx
@@ -1,5 +1,9 @@
-import React from 'react'
+import React, { useState, useEffect, createRef  } from 'react'
+import { useSelector } from 'react-redux'
 import PropTypes from 'prop-types'
+import { Link } from 'react-router-dom'
+import styled from 'styled-components'
+import { isEmpty, isEqual, union } from 'lodash'
 import {
   Panel,
   Button,
@@ -14,16 +18,11 @@ import {
   TagsSelect,
 } from '@ynput/ayon-react-components'
 
-import { useSelector } from 'react-redux'
-import { useState } from 'react'
-import { useEffect } from 'react'
 import getFieldInObject from '@helpers/getFieldInObject'
-import { isEmpty, isEqual, union } from 'lodash'
-import StatusSelect from '@components/status/statusSelect'
-import TypeEditor from './TypeEditor'
 import EntityDetailsHeader from '@components/Details/EntityDetailsHeader'
-import { Link } from 'react-router-dom'
-import styled from 'styled-components'
+import StatusSelect from '@components/status/statusSelect'
+
+import TypeEditor from './TypeEditor'
 
 const SubRow = styled.div`
   flex: 1;
@@ -378,7 +377,7 @@ const EditorPanel = ({
       finalValue = type
     }
 
-    return { value: finalValue, isChanged, isOwn, multipleValues }
+    return { value: isOwn ? finalValue : null, isChanged, isOwn, multipleValues }
   }
 
   // update the local form on changes
@@ -396,7 +395,7 @@ const EditorPanel = ({
     const type = oldValue?.attrib?.type
     let newValue = value
 
-    if (Array.isArray(newValue) && newValue.length == 1 && newValue[0] === null) {
+    if (Array.isArray(newValue) && newValue.includes(null)) {
       newValue = null
     }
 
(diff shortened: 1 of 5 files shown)


Innders (changes requested):
Tools only has an option to clear to empty list now when it needs to be able to clear to null so it can be inherited.

_PR_
image
_current (except that tooltip is wrong because that's clear to inherit not empty list)_
image


It works when I choose "a (inherited)" value but then I don't see the inherited value.

https://github.com/user-attachments/assets/0818939d-e58f-468f-990a-dbfd168f5d65

src/pages/EditorPage/EditorPanel.jsx
```diff
@@ -644,36 +677,60 @@ const EditorPanel = ({
...
+                    }
+                    if (!isMultiSelect) {
+                      return (
+                        'inherited - ' + attrib.enum.filter((el) => el.value === parentValue)[0]?.label || null
+                      )
+                    }
+                    const labels = attrib.enum
+                      ?.filter((el) => parentValue.includes(el.value))
+                      .map((el) => el.label || el.value)
+                    return `inhreited - (${labels.join(', ')})`
```
Innders, line 692: type

Innders, line 692: typo `inherited`

src/pages/EditorPage/EditorPanel.jsx
Innders, line 1: When selecting "inherited" value the dropdown should close straight away. It does not make sense to continue selecting more options.

image

And trying to save this selection causes an error.

image

Innders, line 1: If I set inherited and then set a different value then the inherited value will still show.

https://github.com/user-attachments/assets/876c159d-4457-47ce-8b10-acea6c054844

src/pages/EditorPage/EditorPanel.jsx
```diff
@@ -644,36 +677,60 @@ const EditorPanel = ({
...
+                  if (inheritedOptionLabel) {
+                    placeholder = inheritedOptionLabel
+                    options = [{ value: null, label: inheritedOptionLabel }, ...attrib.enum]
+                  }
+
                   // never show value when inherited, just show placeholder
                   if (!isOwn) enumValue = null
-
                   input = (
                     <Dropdown
```
Innders, line 712: Overflow styling issue. Maybe it's an issue ARC?

image

src/pages/EditorPage/EditorPanel.jsx
```diff
@@ -644,36 +677,60 @@ const EditorPanel = ({
...
-                      style={{ flexGrow: 1 }}
+                      style={{
+                        flexGrow: 1,
+                        fontStyle: isOwn ? 'normal' : 'italic',
+                        color: isChanged
+                          ? 'var(--color-on-changed)'
+                          : !isOwn
+                          ? 'var(--md-ref-palette-neutral-variant60'
```
Innders, line 719: use `--md-sys-color-outline` instead

Innders (changes requested):
Getting there but still have some major bugs to clean up.

flynput:
Code rewritten, PR open on new branch, closing this one.

PR #758, opened 2024-08-28
Dashboard: Create project button removed for normal users 

### LINKED ISSUE #753: Remove 'Create first project' button from dashboard page for non-admin users when no projects are assigned to it
Users with no project creating credentials and no assigned projects are shown a 'Create first project' button when visiting the dashboard. This should be hidden, attempting to create a project will fail anyway.

image

flynput:
Non admin/manager users cannot create projects, the "create new project" button was removed.
Also applied minor styling changes.

Screenshot 2024-08-28 at 10 47 15

Closes #753


Innders (changes requested):
Make the button a bit taller please

PR #767, opened 2024-08-30
Bundles: Migrate getBundles queries to auto-gen typescript

Innders:
## Description of changes 
Now bundle (GET) queries will use the auto-generated RTK queries. Now bundle query data have types! 🥳 

## Testing
These queries are used everywhere so extensive testing is needed.

- [ ]


flynput (approved):

martastain:
I wasn't able to break it

martastain (approved):

PR #786, opened 2024-09-09
708 editor enum dropdown fixes and enhancements remade

flynput:


flynput:
Recreated changes in a clean branch

PR #788, opened 2024-09-09
708 editor enum dropdown fixes and enhancements remade

### LINKED ISSUE #708: Editor: Enum dropdown fixes and enhancements

flynput:
Several changes were implemented to improve the editor user experience:
- Listing tools column uses custom formatter to display inherited values / No values
- Removed clear buttons on all Dropdown form items for consistency
- Added new option for non-custom Dropdown fields which resets value to 'inherited' when selected

https://github.com/user-attachments/assets/de261059-391b-4d87-898d-4167b3849300

Closes #708


flynput:
> I'm a bit confused with the dropdown listing `a` for test, and when set... the column in editor displaying `aa`? :)

True, it should use the label instead 👍

flynput:
@BigRoy Fixed the value/label issue, attaching a new screen recording:

https://github.com/user-attachments/assets/f037d4b3-337d-42d5-9849-24b8246162b7

BigRoy:
Nice @flynput - it's a bit unclear but is now in both displaying the **label** or the **value**? Not sure which `a` was of two.

flynput:
> Nice @flynput - it's a bit unclear but is now in both displaying the **label** or the **value**? Not sure which `a` was of two.
> 
> If it's the **label**, then this looks good to me.

Certain attributes have custom formatting - i.e. the Tools one - the rest were displaying by default the value.
That changed to displaying the label if it's an enum & has labels, both in the listing and the form

BigRoy (approved):
Didn't test but videos look good to me. :)

Innders (approved):

PR #789, opened 2024-09-09
Update product type `vdb` to `vdbcache` name for icon

BigRoy:
## Description of changes 

I'm pretty sure that VDBs published from Houdini and Maya do not show the relevant `vdb` icon because the product type is `vdbcache` and not `vdb`. So this refactors the hardcoded icon to match `vdbcache` name.

I did not test this. :)

## Technical details

Also see: https://github.com/ynput/ayon-traypublisher/pull/13

## Additional context

No idea whether changing this may affect anything else.

### DIFF
--- a/src/features/project.js
+++ b/src/features/project.js
@@ -15,7 +15,7 @@ export const productTypes = {
   setdress: { name: 'setdress', icon: 'forest' },
   groom: { name: 'groom', icon: 'content_cut' },
   matchmove: { name: 'matchmove', icon: 'switch_video' },
-  vdb: { name: 'vdb', icon: 'local_fire_department' },
+  vdbcache: { name: 'vdbcache', icon: 'local_fire_department' },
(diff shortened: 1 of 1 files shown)


BigRoy:
vdb_icon

Here's an example of the issue actually.

The maya and houdini publishers (and loaders) use `vdbcache` which currently do not display in the frontend with an icon. With https://github.com/ynput/ayon-traypublisher/pull/13 the traypublisher would _also_ publsh it as `vdbcache` so that the product type is the same everywhere - making this new icon name the right default.

Meaning that in this example above, where `vdbMain` and `vdbTest` (considered to be the wrong product type `vdb`) would then display the default box icon but `vdbcacheMain` would display the fire.

🔥

Innders (approved):
I can't say I have ever seen the "local fire department" icon anywhere so far so probably safe to change.

If someone complains we can add it back.

Thanks @BigRoy 😊

PR #797, opened 2024-09-12
Releases: Install latest release tool

Innders:
A useful tool for installing the latest release bundle without having to go through bootstrap. It will install selected addons along with selected launchers and dependency packages.

image
image

## Description of changes 
- New menu item `Install latest release`
- Addon selection - preselected production bundle addons otherwise all addons.
- Platform selection - preselects production bundle installer platforms otherwise guesses based on browser.
- Installs release addons, installers and dependency packages from the cloud.
- Realtime download progress bar.
- Cloud connection is required.

## Technical changes
- Redux slices now have `useAppSelector` and `useAppDispatch` with TS support. Now use these instead of usual (hooks)[https://redux-toolkit.js.org/tutorials/typescript#use-typed-hooks-in-components].
- Updated redux to 9.X.X
- Converted addons queries to TS
- Converted installers and dep packages queries to TS 
- Converted many components to TS

### Playwright
`yarn test installRelease.spec.ts`

### Install release
1. Open dialog from app menu (dots menu)
2. Change some addons
3. Change platforms
4. Click install
5. Watch progress
6. See success screen and restart server

### Bootstrap
Start bootstrap and test everything still installs correctly.

### Other testing
- Addons manager still works
- Creating bundles still works (getting addons, installers and dependency packages)
- Anywhere addons are... still works.
- Uploading addons, installers and dep packages.


Innders:
Added creation of a new bundle after finishing install. Ready for final testing @flynput :)

flynput (approved):
Tested, works as expected, bundle created every time with the correct configuration

martastain:
How do you compute the overal progress?  
since the launchers are installed first, it takes quite a long time before the progress bar moves. wouldn't it be better to start with addons? Just cosmetics, probably does not matter, but for some time i thought the process is stalled

Innders:
> it takes quite a long time before the progress bar moves

Latest commit uses the sizes of the files to better show the real progression. The size of the addons are extremely inflated as they are so small.

Also hidden the Launch Bootstrap Setup menu item to developer mode only.

image

PR #800, opened 2024-09-12
URI: Removed extra url encoding when synching URI with url param

### LINKED ISSUE #799: Encoded 'uri' param which stays in sync with the app 'URI' is double encoded
We need to remove our explicit URI encoding, since the URLSearchParams library is already doing it for us.

Current value:   ayon%252Bentity%253A%252F%252FTest%252FFolder1%252FFolder2
Expected value: ayon%2Bentity%3A%2F%2FTest%2FFolder1%2FFolder2

flynput:
Removed unnecessary encoding.

https://github.com/user-attachments/assets/0cc199bc-2a5d-4623-9274-b21820f8e36a

Closes #799


Innders (approved):

BigRoy (approved):
URI was indeed double encoded - so this makes sense. Didn't test it. :)

PR #801, opened 2024-09-13
Global: 401 error not handled correctly

Innders:


martastain:
Now it works for some of the queries, but not all of them. Opening a project from the dashboard for example redirects me to the login page, but if i have a project opened, other window logs me out and in the original window i switch to task progress, folder selection triggers 401s without redirection:
image

Innders:
This should now handle 401 for both rest and graphql

PR #805, opened 2024-09-13
603 feed emoji reaction

### LINKED ISSUE #603: Feed: Emoji reaction
## Story

As an user, I want to quickly express a reaction to an entry in an activity feed my setting an emoji on it.

(Something similar to Github, Discord, Slack, Whatsapp, etc...)

flynput:
Added emoji reactions to comments. A limited set of reactions are available: 👍 👎 ❤ 🎉 📝

https://github.com/user-attachments/assets/a42b4a81-7a7e-4757-a35b-79c9e1195f75

Closes #603


src/components/ReactionContainer/ReactionStyles.styled.ts
Innders, line 1: What do you think about blocking all other clicks when the menu is open? Take a look at how github works, it's quite simple.

 This would solve two problems with some very simple CSS and logic:

- Closing the menu you click outside of the menu.
- Prevent two emoji menus open at one time.

src/services/reaction/updateReaction.ts
Innders, line 1: 1. I have an issue where sometimes the reaction doesn't update in the UI, I can't track down any pattern to it.
2. The reaction takes a split second to appear because we are waiting for the mutation and invalidated query to finish. Could we use [optimistic updates](https://redux-toolkit.js.org/rtk-query/usage/manual-cache-updates) here to make instant?

Innders (changes requested):
A really cool feature! Now just a little bit of polish to get it to 100% 👍

src/components/ReactionContainer/Reaction.tsx
```diff
@@ -0,0 +1,31 @@
...
+    <Styled.Reaction
+      className={clsx(variant, { active: reaction.isActive })}
+      onClick={onClick}
+      data-tooltip={users}
+    >
+      <div className='inner-content'>
+        {reactionMappingObj[reaction.type]}
+        {reaction.users && <span>{reaction.users.length}</span>}
+      </div>
+    </Styled.Reaction>
```
Innders, line 27: Only the emoji needs moving down 1px
```suggestion
      <div className="inner-content">{reactionMappingObj[reaction.type]}</div>
      {reaction.users && <span>{reaction.users.length}</span>}
```

src/components/ReactionContainer/Reactions.styled.ts
```diff
@@ -0,0 +1,106 @@
...
+    height: 32px;
+    padding: 8px;
+    border-color: transparent;
+    border-radius: var(--border-radius-m);
+    &.active {
+      background-color: var(--md-sys-color-secondary-container);
+      border-color: transparent;
+      &:hover {
+        border-color: var(--md-sys-color-surface-container-highest-hover);
+      }
```
Innders, line 40: I think the hover effect on the bg works better then on the border 😊

```suggestion
      background-color: var(--md-sys-color-secondary-container);
      &:hover {
        border-color: transparent;
        background-color: var(--md-sys-color-secondary-container-hover);
      }
```

src/components/ReactionContainer/Reactions.styled.ts
```diff
@@ -0,0 +1,106 @@
...
+  left: 0;
+  z-index: 60;
+
+  background-color: var(--md-sys-color-surface-container-high);
+  padding: 4px;
+  border-radius: 4px;
+  box-shadow: 0px 0px 6px 2px rgba(0, 0, 0, 0.15);
+
+  animation: 0.03s ease-in forwards;
+  transform-origin: bottom left;
```
Innders, line 79: This line does nothing because it has no animation

Innders (approved):
❤️ 😍 🚀 🎉 💯 🥇

Issue #712, opened 2024-08-06
Bootstrap: info about used linux variant


antirotor:
## Story
When running bootstrap process, platform selection allows to select `windows` | `macos` | `linux`.
With linux selection, I'll always automatically get rocky9 specific binaries of launcher and dependency packages. Currently, we don't have any mechanims on pipeline side to differentiate, so at least information about what will be used should be presented.

### Problems
User expirience after bootstrapping on linux environment other than Rocky 9 can be confusing as ayon launcher and/or dependency package won't probably work.

### Proposal
We should show information about currently supported linux platform at least until we can provide the selection of different linux variants. Something along the line: "This will install Rocky 9 specific environment that should be compatible with all EL9 linux variants".

## TODO
If we ever decide on extended support for mac - Intel AND ARM-based, we will have to provide something similar there. Currently it could use some information too, like that current support is limited to Intel based platform (or using Rosetta emulation).

Innders:
The only information in the release that relates to rocky-9 is the filename and MacOS has nothing to help with Intel/ARM.
![image](https://github.com/user-attachments/assets/3a9cb97a-09e8-406d-981c-958fb4f05eea)

Because this can't be pulled from the release itself can the "notes" be hardcoded. If so what exactly should they say?

![image](https://github.com/user-attachments/assets/d4c8287a-f029-458c-b41e-3d82cab5f0c7)

Innders:
> what exactly should they say?
@antirotor could you help out please so that @flynput can implement it.

PR #807, opened 2024-09-17
fix(onboarding): Added Linux EL9 disclaimer on platform selection as …

flynput:
Added EL9 Linux platform disclaimer - hard coded for now 

https://github.com/user-attachments/assets/a6b1cd61-634c-4ac7-b859-ff0521224f39

### DIFF
--- a/src/components/AddonCard/AddonCard.styled.ts
+++ b/src/components/AddonCard/AddonCard.styled.ts
@@ -33,6 +33,14 @@ export const AddonCard = styled.button`
     line-height: 100%;
   }
 
+  .icon.info {
+    position: relative;
+    top: -8px;
+    left: -4px;
+    font-size: 12px;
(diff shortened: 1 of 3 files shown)


flynput:
@Innders Please confirm the implementation is as requested.

flynput:
@antirotor Let us know if the implementation is good enough to solve #712 .

Innders (changes requested):
Please remove the icon. Just need a tooltip that appears when hovering the platform card.

flynput:
> Please remove the icon. Just need a tooltip that appears when hovering the platform card.

Updated, screen recording also.

antirotor:
I would say this will do fine. We should provide better compatibility info somewhere in the docs and later on provide link from this page. But so far, this will do.

Innders (approved):

PR #808, opened 2024-09-18
Bundles: Cancelling file upload on window close

### LINKED ISSUE #733: Bundles: Upload launcher dialog closes during upload
When user uploads a launcher from the `bundles` page (and i assume the same applies to dependency packages and addons), and they click outside the dialog during the upload, dialog closes.

It seems that the upload somehow continues in the background and is eventually uploaded, but it is very confusing and feels like the upload was aborted without a warning.

flynput:
When uploading installers/addons, whenever the upload dialog is closed, an abort signal is sent to axios, stopping the upload.

https://github.com/user-attachments/assets/5a5a7996-8b18-47d0-949a-234753d0a9d6

closes #733

### DIFF
--- a/src/components/AddonDialog/AddonDialog.jsx
+++ b/src/components/AddonDialog/AddonDialog.jsx
@@ -5,8 +5,10 @@ import AddonUpload from '@pages/SettingsPage/AddonInstall/AddonUpload'
 const AddonDialog = ({ uploadOpen, setUploadOpen, uploadHeader }) => {
   // keep track is an addon was installed
   const [restartRequired, setRestartRequired] = useState(false)
+  const abortController = new AbortController()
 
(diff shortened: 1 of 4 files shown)


martastain:
I think it should primarily prevent accidental closing of the dialog during the upload rather than aborting the process. It is really easy to click outside and lose the progress.

Innders (changes requested):
Cool, could we also get a toast to let the user know the upload was cancelled? 

And also what Martin said:

- Prevent closing on escape and clicking outside when an upload it happening (a.k.a only the Close button works).
- Rename `Close` to `Cancel`

mkolar:
Looking good to me.

Innders (approved):

PR #809, opened 2024-09-18
reactions: Reactions not working in Firefox in certain scenarios

### LINKED ISSUE #603: Feed: Emoji reaction

flynput:
Falling back to empty array when updating reactions list to avoid uncaught errors on first reaction.

Closes #603


Innders (changes requested):
If `reactions` is undefined I think it would be good to fix the source of the issue which is probably there is an activity being added to the cache with `reaction: undefined`? 

Probably an issue in `updateActivities` maybe?

But this also works as a backup of course.

Innders (approved):

PR #810, opened 2024-09-18
enhancement(Users): Warning message shown on deletion attempt, user n…

### LINKED ISSUE #796: Users: Deleting users should be really hard
## Story

Deleting a user is really bad and can muck up a lot of things. Inactivating a user is prefered.

### Proposal

1. The dialog should have a warning message like the rename users dialog. It could say:
```Deleting users can have unintended consequences. Consider deactivating the user instead?```

[example]
image

2. Deactivate user button

3. [BONUS] The user must type the username into an input to confirm they are deleting the correct user.

[github example]
image

## TODO

- [ ] Warning message
- [ ] Deactivate user secondary button
- [ ] Type to delete if they do try and delete.

flynput:
Deleting users dialog features a warning message recommending disabling users instead of deleting.
The users is also required to fill out the to-be-deleted user(s) to confirm his action.

https://github.com/user-attachments/assets/f459cb49-06f0-4c7e-b6a0-6cf7a46f5406

Closes #796

### DIFF
--- a/src/pages/SettingsPage/UsersSettings/DeleteUserDialog.tsx
+++ b/src/pages/SettingsPage/UsersSettings/DeleteUserDialog.tsx
@@ -1,0 +1,59 @@
+import { useState } from 'react'
+import { InputText, FormLayout, FormRow, Dialog, Button } from '@ynput/ayon-react-components'
+import InfoMessage from '@components/InfoMessage'
+
+type DeleteUserDialogProps = {
+  onHide: () => void
+  selectedUsers: string[]
+  onSubmit: () => void
+}
+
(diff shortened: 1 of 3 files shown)


Innders (changes requested):
Could we please have a button bottom left that allows you to deactivate the user instead of deleting them.

`active=false`

flynput:
@Innders Added disable button, updated screen recording also.

Innders (changes requested):
Dialog works nicely 👍 

Just one small annoying thing... when disabling the user the selection on the user is lost. It felt a bit jarring and I had to select the user again in order to check they were actually disabled.

Innders (approved):

PR #822, opened 2024-09-24
Attributes: New enum editor with support for colours and icons

flynput:
Attribute enum component now supports color and icon values, both optional. Defined items are also draggable, can be duplicated and easily removed.

Uploading Screen Recording 2024-09-25 at 17.28.20.mov…


Innders (changes requested):
Working really nicely, awesome stuff, just need some small tweaks 🤏 

- [x] - Click on header to open and close
- [x] - Open header hover bg styles and pointer cursor
- [x] - Opening/creating/duplicating should auto focus on label input
- [x] - Inputs full width
- [x] - color and icon default off to remove question mark
- [x] - Value needs to fill as label is filled out
- [x] - icon dropdown full width
- [x] - Use real buttons with no background
- [x] - Animation: remove opacity, cut off items and double speed
- [x] - Drag pointer
- [ ] - Drag item goes behind items and is cut off on the sides
- [ ] - Add new item button is full width and aligned left
- [x] - Remove colored text and instead have a rounded square with color
- [ ] - Managed to create a enum item with no label or value

martastain:
Could you also please disable/hide "Default" input when "Scope" does not contain Project?

https://github.com/ynput/ayon-frontend/issues/824

Innders (changes requested):
- [ ] When the item is open then the header should still have a hover style.
- [x] Animation speed 220ms
- [x] Styling of color is wrong when check is on.
- [ ] For duplicate and remove button use surface-highest-hover for hover.
- [ ] Duplicate doesn't copy over value, maybe this is on purpose to prevent duplicates? When duplicating copy over value and then suffix with -[n]. The label text should also be preselected so you can quickly change it.
- [ ] The syncing of state seems to be broken now? When you save the attribute the enum changes are lost.
- [ ] Probable want to use <DragOverlay /> to fix dragging issues.
- [ ] Add new item button is full width and aligned left
- [ ] Managed to create a enum item with no label or value
- [ ] Remove Icon gap when there is no icon
- [ ] Move color preview to left side and make 14px
- [ ] Weird animation glitch caused by setting focus whilst animating `focusLabelIfExpanded()`.

Things to think about

- Sync handler called when opening/closing the item?
- `values: $Any` is the type of an enum item not well defined?
- Rename `AttributeDropdownItem` to `EnumItem` and `AttributeDropdown` to `EnumEditor`. It's not an attribute or a dropdown.
- Use relative names for styled components if they are exclusive to that component. `AttributeDropdownItemBody` -> `Body`. What do you think, I'm always unsure on naming for styled components?
- `syncHandler`:  would just `onChange` make more sense?

image

Innders (changes requested):
- [x] Color no value shows nothing - need a placeholder
- [x] Color input style needs to match other inputs, border-radius, border and placeholder.
- [ ] Swap icon and color positions in the header (color comes first).
- [ ] The name `DraggableAttributeEnum` makes it sound like the whole thing is draggable, WDYT? Maybe just `AttributeEnumEditor` or something? `DraggableAttributeEnumItem` makes sense because you can drag the item.
- [ ] syncHandler is called when opening/closing the item.
- [ ] Disabled inputs opacity: 0.5 (icon and color).
- [ ] Value should not change based on label unless new attribute.

### Duplicating
- [ ] When duplicating focus and preselect the label text so it can be typed over straight away.
- [ ] When duplicating include the suffix in the label as well.
image
- [ ] Further duplication breaks the suffix counter.
image



Previous requests not resolved:
- [ ]  When the item is open then the header should still have a hover style.

Innders (approved):
Works flawlessly and really fun to use, awesome! 🥳 

>Could you also please disable/hide "Default" input when "Scope" does not contain Project?

@martastain could you please check the small `default scoped to projects only` that fixes #824.

Below is some nice marketing material for any docs or release notes. In the future @flynput could you please have a go at creating these yourself when the feature/enhancement is interesting and worth shouting about :)

- Custom Attribute Colors and Icons: Enumerators now support colors and icons, enabling highly customizable attributes for visually engaging user interfaces.

attribute-enum-editor

martastain:
Displaying colors in the dropdown and icons&colors in table views is out of the scope of this PR?
image

martastain:
> @martastain could you please check the small default scoped to projects only that fixes https://github.com/ynput/ayon-frontend/issues/824.

This works as expected.

martastain (approved):

PR #837, opened 2024-10-01
Editor: Fixed crash condition on assignee update

flynput:
Crash caused by complicated logic when identifying if values have changed in the editor page

https://github.com/user-attachments/assets/2d1ca9ba-f980-4046-8eab-f31cc5e27776

### DIFF
--- a/src/pages/EditorPage/EditorPanel/EditorPanel.jsx
+++ b/src/pages/EditorPage/EditorPanel/EditorPanel.jsx
@@ -223,6 +223,7 @@ const EditorPanel = ({
                             nodes,
                             setLocalChange,
                             setForm,
+                            isAttribute: false,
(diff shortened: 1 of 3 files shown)


Innders (approved):
It all seems to work now. But I noticed one styling issues which I think is unrelated to this but should be fixed.


https://github.com/user-attachments/assets/1b22ffb1-1efe-4a84-a4a7-21808847875d

flynput:
image


> It all seems to work now. But I noticed one styling issues which I think is unrelated to this but should be fixed.
> 
>  Screen.Recording.2024-10-02.at.07.43.14.mov

I opened a new ticket specifically for that issue: https://github.com/ynput/ayon-frontend/issues/839
Merging this PR.

PR #840, opened 2024-10-02
Statuses: Statuses filtered by project scope configuration

### LINKED ISSUE #823: Entities: scoped statuses
Entity statuses can be scoped to an entity type(s). By default, `scope` parameter of the status contains all entity types, but the availability can be limited to only certain entity types. 

image

Status selector across the interface should reflect this setting and hide the option if the entity type is not listed in the `scope` parameter of the status. Since the status selector already takes data (icon, color, short name) from the same dataset, it should be straightforward.

image

Relevant backend PR: https://github.com/ynput/ayon-backend/pull/287

flynput:
Entity status selector now reflects the project anatomy configuration. States that are not available for the entity are hidden

https://github.com/user-attachments/assets/905345d7-bf85-4b17-a461-3c278a3ee70a

Closes #823

### DIFF
--- a/src/app.jsx
+++ b/src/app.jsx
@@ -151,7 +151,7 @@ const App = () => {
   // DEFINE ALL HIGH LEVEL COMPONENT PAGES HERE
   const mainComponent = useMemo(
     () => (
-      <ErrorBoundary FallbackComponent={ErrorFallback}>
+      <>
         <Favicon />
         <WatchActivities />
         <Suspense fallback={<LoadingPage />}>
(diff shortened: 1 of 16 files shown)


Innders (changes requested):
- [ ]  Invalidation didn't work in the dashboard.

https://github.com/user-attachments/assets/41d250a3-9712-41ae-bac2-d031ef0d17fb

- [ ] Task progress page entity cards don't respect the scopes

https://github.com/user-attachments/assets/63172d70-eaf3-4843-9952-a6de1327daeb

- [ ] If the status does not exist on the card then the column should not show. (sorry told you the wrong thing)

image

Innders (approved):
All my tests work!

PR #865, opened 2024-10-10
854 support drag and drop reordering for list fields

flynput:
RJSF library was update to version 5, which allows for list and list item level template customization.
The code was refactored to make it easier to understand, the logic was kept the same with only minor alterations.

https://github.com/user-attachments/assets/55b3b981-2eed-496b-a827-0a32b66ffc80


Innders:
When the screen gets too small. Might be an issue with settings in general and not this PR though.

image

Innders:
I have added a small commit that adds a hover style for dragging and deleting.

image
image

Innders (changes requested):
🤯 🤯 🤯 

This is going to make changing these settings so much nicer!

One small thing, the page crashes when selecting Example addon for whatever reason.

martastain:
Actually many models crash the page (Nuke, Maya, Hiero, Shotgrid...).
In most cases the crash is tied to: 

```
chunk-4IVSI7TS.js?v=165ee28d:9129 Uncaught TypeError: Cannot read properties of undefined (reading 'type')
    at ObjectFieldTemplate.tsx:92:66
```

I assume it is tied to custom model attributes such as model_switcher

flynput:
> Actually many models crash the page (Nuke, Maya, Hiero, Shotgrid...). In most cases the crash is tied to:
> 
> ```
> chunk-4IVSI7TS.js?v=165ee28d:9129 Uncaught TypeError: Cannot read properties of undefined (reading 'type')
>     at ObjectFieldTemplate.tsx:92:66
> ```
> 
> I assume it is tied to custom model attributes such as model_switcher

Added a fix for the crashes, also removing some unneeded auto-generated labels for string lists.

flynput:
> > Actually many models crash the page (Nuke, Maya, Hiero, Shotgrid...). In most cases the crash is tied to:
> > ```
> > chunk-4IVSI7TS.js?v=165ee28d:9129 Uncaught TypeError: Cannot read properties of undefined (reading 'type')
> >     at ObjectFieldTemplate.tsx:92:66
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I assume it is tied to custom model attributes such as model_switcher
> 
> Added a fix for the crashes, also removing some unneeded auto-generated labels for string lists.

Nevermind, I still see it crashing... I'll keep you posted.

Innders (approved):

PR #871, opened 2024-10-11
Specs: Fixing faulty specs

### LINKED ISSUE #818: Testing: Delete users test broken due to confirmation dialog
image

image

flynput:
A few fixes to the specs, attempting to make them pass on each repeat.

Closes #818


Innders (commented):
All worked except deleting users would always get stuck here:

image

Innders (approved):

PR #877, opened 2024-10-16
Permissions: More granular access and Project Access page

martastain:
## Conditional access to project settings tabs based on user permissions

https://github.com/ynput/ayon-backend/pull/389 implements user/permission endpoints that return granular project permissions. Frontend should reflect these permissions and use this object instead of relying on isManager flags.

https://experimental.ayon.dev/docs#tag/Users/operation/get_current_user_permissions
https://experimental.ayon.dev/docs#tag/Users/operation/get_current_user_project_permissions

When the user is a manager, this object is returned accordingly (with all restrictions disabled)

This should (for now) be implemented only in the project settings page

## Add Users page to Project settings

Add a new sub-page to project settings that would allow non-managers (but users with `project.users` write access) to add and remove users to the project (by assigning access groups to them).

1. Original users page stays the same until we decide what to do with it.
2. Users pages on project settings is called **Users Projects**
3. Use single project selection on the left hand side to filter users by the project.
4. Project selection also affects project access column to show for that project.
5. Checkbox button in the toolbar that toggles to filtering by project. This is needed in order to see users with no project access, so they can be added to a project. The switch is called **Filter by project access** and is default **off**.
6. Columns are `username, full name, project access`.
7. We should only see users that are `active=true`.
8. Right hand side should show selected user header with access summary for project and access groups table (full height).
9. Feel free to rewrite all of this inspired by user settings.
10. Use new SearchFilter bar for filtering of users.
11. SearchFilter bar could filter by project which could remove need for 3. and 5.

### Read only users
- Access Groups table will need a read only mode.

This requires using a new set of endpoints, more information will be provided.


Innders (changes requested):
- [ ] Random gap here
image

- [ ] Admins/managers can't see Anatomy, Project Settings or Project Access
image

Innders (changes requested):
- [ ] Site settings blocked
image

- [ ] I can land on anatomy tab even if I do not have access to it. I should be redirected to a page I have access to.
image

- [ ] Changing project management permissions seems to make no difference.
image

flynput:
@Innders pushed the latest changes, I realized there is one issue remaining: getting the list of users involved in the selected projects only considers the first project in the selection, that's because we cannot fetch user access groups for multiple projects at once. I would gladly use another endpoint that provides said information @martastain

Innders:
> we cannot fetch user access groups for multiple projects at once

The API may not allow it but we can definitely do it by looping over a single endpoint for different projects. Take a look a custom query -> https://redux-toolkit.js.org/rtk-query/usage/customizing-queries#customizing-queries-with-queryfn

There are many examples in the codebase of this.

https://github.com/ynput/ayon-frontend/blob/cface9e94915602fb0c461a16a3383e7d4a2e5c5/src/services/userDashboard/getUserDashboard.ts#L240-L269

Innders (changes requested):
User management access

- [x] For projects where there is no access, show greyed out in the table and sorted to the bottom.
- [x] Tabs are jumping around, they should be calculated based on permissions and stay consistent. Project selection should make no difference.

Project Access

- [x] Page padding
- [x] Page title Project access
- [x] Missing table titles
- [x] Only show active projects
- [ ] User rows too heigh
- [x] Add shortcut to button (Shortcut Component)
- [ ] Profile image 20px
- [ ] Top right buttons, correct labels and remove shortcut
- [ ] Context menu (I will create a mockup in figma)
- [ ] Access groups width the same as middle table and weird padding on right
- [ ] Toast not very descriptive “Operation successful”
- [ ] Can’t scroll access groups
- [ ] Resize access groups
- [ ] Block double tapping “A” (or think of shortcuts that don’t conflict)
- [ ] Shortcut needs to work when just hovering at the moment it only works once selected.
- [ ] Empty place holder shouldn’t be table item.

Filtering

- [ ] Project filtering doesn’t work
- [ ] Text search doesn’t work
- [ ] User options should not show admins, managers or services.
- [ ] User filtering doesn’t filter users in access group tables.
- [ ] If there is only once access group (from filtering or otherwise) adding should skip dialog and add them straight to that group.
- [ ] Filtering access groups should also filter options in adding modal
- [ ] Select/deselect all access groups

flynput:
The following items were implemented, the ones missing are either in progress or need more feedback/discussion:

Tabs are jumping around, they should be calculated based on permissions and stay consistent. Project selection should make no difference.
Page padding
Page title Project access
Missing table titles
Only show active projects
User rows too high
Profile image 20px
Top right buttons, correct labels and remove shortcut
 Context menu (no dynamic secondary context menu FTM)
Access groups width the same as middle table and weird padding on right
Toast not very descriptive “Operation successful”
Can’t scroll access groups
 Empty place holder shouldn’t be table item.

 User options should not show admins, managers or services.
 User filtering doesn’t filter users in access group tables.
 If there is only once access group (from filtering or otherwise) adding should skip dialog and add them straight to that group.
 Filtering access groups should also filter options in adding modal
 Select/deselect all access groups

flynput:
Added new features to the list: 
 Add shortcut to button (Shortcut Component)
 Block double tapping “A” (or think of shortcuts that don’t conflict)
 Shortcut needs to work when just hovering at the moment it only works once selected.

Pending issues:
- Resize access groups vertically - not sure how to do it
- dynamic context menus

Innders (changes requested):
Project Settings
- [ ] Add new project button is appearing when it shouldn’t. Home page and users settings page.
- [ ] Add new project button showing when user does not have permissions.
- [ ] If user does not have access at all to a project they should not see it in the list. (server issue)
- [ ] Greyed out project row not working for project settings.
- [ ] Greyed out project not visible for project access. (only works on anatomy)
- [ ] Weird loading glitch

Project access
- [ ] Border radius on panels
- [ ] Gap at the top too big.
- [ ] Gap too big between filter and titles
- [ ] Users don’t see users (are you using graphql?)
- [ ] Profile image is too big 26px not 20px.
- [ ] User rows too high, they should be 32px as per figma.
- [ ] Hover missing on edges for add button.
- [ ] Rename no access column header Username -> User

Adding/Removing
- [ ] When multiple selected, pressing A should add all.
- [ ] Add more should use shortcut A
- [ ] Add more dialog should show current groups selected
- [ ] Add more not working?
- [ ] Remove access button enabled when no user selected or non access user selected.
- [ ] Context menu, ignore dynamic menu for now. Remove all access menu icon=remove_moderator.
- [ ] Context menu, match figma text and icons.

Dialog
- [ ] Select all no background, deselect uses prim container BG. No icon.
- [ ] Item height should be 32px not 28px.
- [ ] Keyboard controls: press enter to select/unselect.
- [ ] Keyboard controls: Cmd/Ctrl + Enter to submit.
- [ ] Keyboard controls: Cmd/Ctrl + A select all (bonus if easy)
- [ ] Dialog max height 600px

Filtering
- [ ] Project options show inactive projects.
- [ ] Project text search not working.
- [ ] User options should not show admins, managers or services.
- [ ] Access Group, has value and has no value (bonus).
- [ ] Prevent adding search options more than once.
- [ ] Filters aren’t saving to user frontend preferences.

Innders (changes requested):
Project Settings
- [x] During loading there is a flash of another message.
- [ ] Project selection state should follow URL query like the other pages.
- [x] Add new project button showing when user does not have permissions.
- [ ] Greyed out project not visible for project access, it shows as (read only).

Read Only
- [x] Remove context menu.
- [x] Height of users rows different.

Users Page
- [x] New projects access to Default project access.
- [x] Default project access tooltip: ‘When a new project is created, the user will automatically be added to the selected access groups. To grant a user the ability to manage project access for others, ensure an access group with those permissions is included here.’

Dialog
- [x] Select all button. Use Shortcut widget inside button to show shortcut. Does it change to `Ctrl+A` for windows?
- [ ] Using `Cmd+Enter` closes dialog but doesn’t save access groups.

Filtering
- [ ] Project text search does not work with multiple values. Example: “my” or “ay”. 
- [ ] User text search does not work with multiple values.
- [ ] Access groups text search does not work with multiple values.

Innders (changes requested):
Project Settings
- [x] [x] Project selection state updates URL query but doesn’t persist when refreshing page or changing pages.
- [ ] [x] Add new project button showing on `/dashboard/overview`.

Filtering
- [ ] [x] Text searches don’t work when inverted (excludes). For example the text search of `bi` when inverted should exclude all matches to `bi`. This is the same for all three filters.

image

Innders (approved):
Everything seems to working flawlessly, awesome stuff 👍

I have marked as approved but could you please write some simple tests.

1. Add user to access group.
2. Remove user from access group.

Innders (changes requested):
**Middle table**
- [ ] [x] Middle table of users should show all users regardless of project access.
- [ ] [x] Rename middle table to “All users”.
- [ ] [ ] New column on middle column called “Project access groups”. It should list the access groups and for multi project selection show mixed values are greyed out italics (like in project settings).

**Filtering**
- [ ] [ ] Right click on user “Filter by user” sets that user in the filter bar. Multi selection of users adds all users to filter.
- [ ] [ ] You should be able to do a global search for any search query that matches any user.

**Access groups**
- [ ] [ ] Redesign of access group cards.
- [ ] [ ] Click on whole access group header to open and close.

image

Innders (changes requested):
Middle table
- [ ] [x] Width of columns jumps around depending on selected projects.

Add access dialog
- [ ] [x] When multiple projects are selected and the user has an access group not on all projects, the item is ticked when it should have the dash.
- [ ] [x] When multiple users are select, just show user count in title and not every user name.
- [ ] [x] When clicking on a dashed access group I would expect it to be selected first, not deselected.
- [ ] [x] When selecting a user and then clicking add on an unselected user the dialog will use the selection not the hovered row.

Filtering
- [ ] [x] Filter by user works great but needs to be on access groups table as well.

Access groups
- [ ] [x] minHeight should be 96px (3 rows)
- [ ] [x] placeholder for empty access groups is too big. Just use some simple text, “No users assigned”.
- [ ] [x] Border radius on group header does not go all the way around when expanded.
- [ ] [x] Main panel scrolling broken.
- [ ] [x] Remove scrolling in the tables.
- [ ] [x] Inside of table remove TablePanel with position absolute.
- [ ] [x] Height of header is glitchy and can go too small. Set explicit height instead of padding.

Innders (changes requested):
- [x] Remove top of table border radius
image

- [ ] Gaps between access groups need to be 8px.
image

- [ ] Users table doesn't fill screen height.
image

- [ ] Love the tooltip here! Can we replace the dotted underline with just a normal underline?
image

PR #881, opened 2024-10-16
Maintenance: Code & console warnings cleanup

flynput:
These changes remove most - if not all - console warnings in the settings page. 
They seem safe, please have a look though, I might have removed some that are needed

### DIFF
--- a/src/containers/SettingsEditor/SettingsEditor.styled.tsx
+++ b/src/containers/SettingsEditor/SettingsEditor.styled.tsx
@@ -1,11 +1,7 @@
 import styled from 'styled-components'
 
-type Props = {
-  currentSelection: string
-}
-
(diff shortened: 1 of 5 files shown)


src/containers/SettingsEditor/SettingsEditor.styled.tsx
```diff
@@ -0,0 +1,46 @@
+import styled from 'styled-components'
+
+type Props = {
+  currentSelection: string
```
Innders, line 4: This styled prop should use $ prefix `$currentSelection`.

https://styled-components.com/docs/api#transient-props

Innders (changes requested):

Innders (approved):

PR #897, opened 2024-10-22
fix(Editor): Fixing thumbnails upload in editor page

flynput:
Issue was caused by wrong entity type passed around on while gathering detail data.

https://github.com/user-attachments/assets/c1dfaf03-bfa8-4f08-99d1-f5e1a5eaaaa6


Innders (approved):

BigRoy:
Yes, this works!

PR #899, opened 2024-10-24
fix(Settings): Removed excessive scrollbars, other styling tweaks

flynput:
Added missing right hand padding for top form container, restricted fields max width to avoid horizontal scrolling, removed excessive scrollbars

https://github.com/user-attachments/assets/43c51816-d1fd-443b-a967-3b1479ff2c66

### DIFF
--- a/src/containers/AddonSettings/AddonSettings.jsx
+++ b/src/containers/AddonSettings/AddonSettings.jsx
@@ -37,6 +37,13 @@ import { getValueByPath, setValueByPath, sameKeysStructure, compareObjects } fro
 import arrayEquals from '@helpers/arrayEquals'
 import { cloneDeep } from 'lodash'
 import { usePaste } from '@context/pasteContext'
+import styled from 'styled-components'
+
+const StyledScrollPanel = styled(ScrollPanel)`
+> div {
+  padding-right: 8px;
+}
+`
 
 /*
  * key is {addonName}|{addonVersion}|{variant}|{siteId}|{projectKey}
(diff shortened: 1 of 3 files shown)


src/containers/AddonSettings/AddonSettings.jsx
```diff
@@ -698,7 +705,11 @@ const AddonSettings = ({ projectName, showSites = false }) => {
         <Section className={showHelp && 'settings-help-visible'}>
           {settingsListHeader}
           <Section>
-            <ScrollPanel className="transparent nopad" style={{ flexGrow: 1 }} id="settings-scroll-panel">
+            <StyledScrollPanel
+              className="transparent nopad"
+              style={{ flexGrow: 1 }}
```
Innders, line 710: Might as well move flexGrow to StyledScrollPanel.

Innders (approved):

BigRoy:
YES - very nice!

It fixes the scrollbars, the oversized width and the padding is very nice - lovely.

I did notice this:

image

So yes, the width doesn't go beyond the container now.... but at the same time we're not using all the screen-space we do have available there? It's now always capped to 600px wide?

It appears everywhere by the way:

image

image

Is there any way we can have it grow at max to the width of the container it's in instead of hardcoding the max-width?

flynput:
> YES - very nice!
> 
> It fixes the scrollbars, the oversized width and the padding is very nice - lovely.
> 
> I did notice this:
> 
> image
> 
> So yes, the width doesn't go beyond the container now.... but at the same time we're not using all the screen-space we do have available there? It's now always capped to 600px wide?
> 
> It appears everywhere by the way:
> 
> image
> 
> image
> 
> Is there any way we can have it grow at max to the width of the container it's in instead of hardcoding the max-width?

We can enable text wrapping when rendering textarea elements, would that work? 
image

BigRoy:
> We can enable text wrapping when rendering textarea elements, would that work?

It may be nice for _some of the fields_ but not all - I wouldn't want that on my JSON data or application envs. So I'd say for now, let's not do that.
It's mostly a cosmetic thing - there's screen space available; at best we use it instead of limiting the usage always to 600px.
It also feels a bit like a regression because before it did use it completely, except for the multiselection thing going beyond the borders, right?

PR #901, opened 2024-10-25
fix(Settings): Restoring full width for non-select values, converting styles to scss format

flynput:
600px limitation applies only to select components - they're the ones breaking the layout when content is too large - the rest of form components will stretch to 100% of the available space.
I also converted the sass file to scss for better readability

https://github.com/user-attachments/assets/50934c72-2fad-40e8-9677-f35fa0ec3903


martastain:
Some alignments and still feel weird to me.

 capping dropdown doesn't prevent this:
image

While it feels kinda random here:

image


I don't think there's an universal "it will look great" solution tho.

BigRoy:
> capping dropdown width doesn't prevent this:

I guess it may make sense to also set a minimal width to 'fit' the available width up to the maximum capped width?
Now it seems to resize to the content, but maybe we can just size it always up to the max width?

Then at least those dropdowns there would be equal width?

---

> While it feels kinda random here:
> 
> image

I agree - preferably it would just behave like the others "fit to the width of the container regardless of the content of the drop-down"?

flynput:
Closing, as settings form field width issues were solved in another PR.

PR #904, opened 2024-10-25
fix(Settings): Using default value for new/unset fields

flynput:
@martastain Please confirm that I'm not bloating the form with default values.

https://github.com/user-attachments/assets/6b86a7c2-bdd6-4339-8072-2ce41ffd0dca


Innders (approved):
I get these defaults, but someone with better settings knowledge will need to test it.

image

iLLiCiTiT:
Looks like it works for `str` but not for others, like bool and color.

image

iLLiCiTiT (changes requested):
Works for `str`, does not work for bool, color and maybe few more.

flynput:
> Works for `str`, does not work for bool, color and maybe few more.

Just pushed changes that should fix the default value issue for bools/colors/selects fields also.

iLLiCiTiT (approved):
Works for me

Innders (approved):

PR #907, opened 2024-10-29
Onboarding: Setup pipeline prompt

Innders:
## Description of changes
When no installers or dependency packages are installed then this prompt will show.

Clicking on the prompt will open the **Install pipeline release** dialog.

The aim is to help direct new admins to set up the pipeline for the first time.

This new prompt will be reference by a new tutorial called **Pipeline setup**.

image

## Other changes

- Improved the mandatory buttons inside **Install pipeline release** dialog to say "Pick {items}" rather than just "Change".
- "Pick Addons" button shows when only mandatory addons have been selected.

## Testing

1. Ensure the server has no installers or packages.
image
2. Ensure user logged in is an admin.
3. Click on setup pipeline prompt and install release modal should open.
image
4. After install release is finished and the server has restarted the prompt should not show.
5. Dismissing the prompt will set a local storage flag to hide it.

## Further notes
This could be extended in the future to show whenever there is a new release that needs to be downloaded. Acting a bit like a notification for new released.

### DIFF
--- a/src/components/InstallerDownload/InstallerDownloadPrompt.jsx
+++ b/src/components/InstallerDownload/InstallerDownloadPrompt.jsx
@@ -33,12 +33,11 @@ const InstallerDownloadPrompt = () => {
     <Styled.Container>
       {directDownload && (
         <Styled.DownloadButton icon={'install_desktop'} onClick={handleDirectDownload}>
-          Download Launcher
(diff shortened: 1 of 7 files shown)


mkolar:
This is good! happy to merge

martastain (approved):

PR #909, opened 2024-10-29
Settings: Basic search filtering

flynput:
Studio settings addon filtering works by mapping schema & settings objects for selected addons, and using the resulting paths to only display the matching tree items - and their parents/descendants

What's working:
- searching for labels & nested labels - mapped out addon schema titles & descriptions
- searching for values - mapped out string and numeric values from addon settings
- multi addon search - works also if the selection gets updated after initial search

What's not working OR missing
 - some deep nested references might be missing for both labels and values
 - list items not mapped at the moment

I did limited testing, it definitely needs more of that from users that actually need to filter/change settings.

https://github.com/user-attachments/assets/b777c69a-9f16-4f49-946a-18797f5b1123


flynput:
Settings filtering can now be done on any number of selected addons, it uses - most - of the settings label and a string values. Screen recording was also updated. 

Let me know what you think.

src/containers/AddonSettings/SettingsListHeader.tsx
```diff
@@ -0,0 +1,116 @@
...
+      })
+    }
+
+    return []
+  }
+
+  return (
+    <Toolbar>
+      <SearchDropdown
+        placeholder="Search & filter settings"
```
Innders, line 83: Search and filter settings

Innders (changes requested):
- [ ] Number values aren't matched.
- [ ] Clearing the search field doesn't always clear the filtering
https://github.com/user-attachments/assets/b58152b9-a1f8-489a-bfae-058506543c8f


- [ ] Labels don't match
https://github.com/user-attachments/assets/c267b91e-c221-4834-a44c-b8fad39ed3c7

- [ ] Sometimes the groups don't close.
- [ ] Should the groups expand when opening?

https://github.com/user-attachments/assets/6396f336-ff59-4b71-8dd7-ea1630ea4ba4

- [ ] Expanding and collapsing affects all these groups?
https://github.com/user-attachments/assets/ff8ec12e-f7ea-434b-be21-661bd967e166

- [ ] Value not match here
https://github.com/user-attachments/assets/4bb30cc4-02c1-4e5e-806e-5fd3977b5417

flynput:
Fixed the following issues:
- Labels and values for numbers (int & numeric schema types) should match now, same goes for list items labels/values
- Clearing filter text should reset form to full nodes tree

Innders (approved):
Approved, lets get it deploy to experimental for some more testing!

BigRoy:
Tested Experimental `1.5.4+202410311657` + Windows + Firefox

---


image

1. Super-wide multiselection enum styling is off: `ayon+settings://applications/project_applications/profiles/0/applications`

Issue seems even worse in project settings? `ayon+settings://applications/project_applications/profiles?project=ayontest`

image

2. Also, not sure if it should but searching for "Blender" does not find that value in there. Should it search for values?

Searching for "python" also doesn't find e.g. the Blender app definition arguments which has `--python-use-system-env` as value.

---

3. Is it correct that these list sections aren't full width anymore?

image
image

---

Nice, search even works selecting multiple addons:

image

However in this case trying to expand one of the addon sections for me seems to expand all four of them? Bug?

Innders:
> However in this case trying to expand one of the addon sections for me seems to expand all four of them? Bug?

Separate bug not relevant to this PR.

m-u-r-p-h-y:
Searching for tag webreview used in settings is not returning correct result

webreview tag is used:
image

but not found
image

m-u-r-p-h-y:
`maya` is not found in application addon in profiles

image

BigRoy:
Nice! Much better now.

This is what stands out:

### Filter inside lists?

I wonder if with this search query it should filter out the other tools?

image

---

### Cosmetics: Should we highlight the search bar if active?

Also, I wonder if the search bar need some highlighting or whatever if there's an active search filter because leaving e.g. a space in there will be VERY confusing. But also just having a decent text filter set but then going to other addons it can be easily missed that the search filter is still active.

---

### Cosmetics: Search bar rounded corners?

On firefox it looks like the search bar styling lacks rounded corners like other text fields.

**Firefox:**

image

**Chrome:**

image

m-u-r-p-h-y:
I'm not sure if this is relevant to this PR, but it would be great to have a `clear` button so we do not have to click the search bar and manually remove the search string to restore it to its default state. (quite annoying)

martastain (approved):

PR #918, opened 2024-11-05
fix(Settings): Aligning form fields in settings pages

flynput:
Form fields stretch as much as possible up to 800px, included date fields and selectors

https://github.com/user-attachments/assets/d1c23976-b7cd-4dbe-8f38-8aa408deee73

### DIFF
--- a/src/containers/SettingsEditor/FormTemplates/ObjectFieldTemplate.tsx
+++ b/src/containers/SettingsEditor/FormTemplates/ObjectFieldTemplate.tsx
@@ -104,7 +104,7 @@ function ObjectFieldTemplate(props: { id: string } & ObjectFieldTemplateProps) {
         else otherFields.push(element.content)
       }
       return (
-        <div className="foo">
+        <div style={{width: '100%'}}>
           {longDescription}
           <div className={className}>
             <div className="name-field">{nameField}</div>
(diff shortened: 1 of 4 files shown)


BigRoy:
I'm mostly curious to see how this behave if the bars do not fit in the full width, and what it looks like e.g. in a deeply nested settings widget.

flynput:
> I'm mostly curious to see how this behave if the bars do not fit in the full width, and what it looks like e.g. in a deeply nested settings widget.

Removing the max-width part you get something like this: occasional scroll bar for large multi-select values, otherwise it seems fine. I can't figure out a way to force the select component to respect the 100% parent width.

image

BigRoy:
> I can't figure out a way to force the select component to respect the 100% parent width.

Haha - sounds like exactly `css` hell that I always face when I ever tried coding anything with it. 
Does [`flex-grow`](https://developer.mozilla.org/en-US/docs/Web/CSS/flex-grow) mean anything in this context?

Anyway, them becoming too big is definitely a worse problem. But I suspect `max-width 800px;` may just have the same issue but less apparent on larger screens. But if e.g. nested deep it may still behave problematic, or?

flynput:
I managed to find a hackish way to keep the form fields at 100% while not overflowing, @Innders let me know what you think of that 'width: 0'
Tested in Chrome & Firefox, it looks good!

PR #933, opened 2024-11-21
Slicer: Folder hierarchy, users or type

Innders:
## Description of changes 
The folder hierarchy selection pane has got a new design and new features.

Select a folder like usual or "slice" by a different field, like users or status.

Slicing is a temporary filter that gets applied to any data you are looking through. It's useful for when you want to quickly focus on one specific user or status without having to change your previously setup filters.

## Next iteration
- Expand to more pages like the new editor and browser.
- Use dynamic fields based on attributes and current values.

## Technical details

The panel is using a new table library that has drastically improved the performance for large amounts of data.

## Additional context

https://github.com/user-attachments/assets/c12d5dab-0836-4d83-a7b1-866db67e4af2


martastain:
It works really nicely. but i had to comment out that "useRemote" features as it expects the example component and crashes without it. It should be a part of another PR i think. The table is a little noisy in the browser console as it logs it performance stats.

flynput (approved):

PR #944, opened 2024-12-02
Project Settings: limited access users UI/UX changes

flynput:
Users with no permissions on certain projects will now see corresponding placeholder when accessing Project -> Roots page.
'Add new project' button has been added back in the Project settings section for those with create permissions.

https://github.com/user-attachments/assets/8b3f5a9b-0f32-4ecf-9727-d4651f76f352


Innders (changes requested):
- [ ] [x] Roots: if there is no access to a project, it should appear greyed out in the same way on the other tabs.

PR #972, opened 2025-01-10
Roots: add scrolling for overflowing roots

Innders:
This is quite hard for me to test as I have no sites and no roots. Maybe @martastain you could quickly test?


martastain (approved):
Confirmed, fixes the issue.

PR #983, opened 2025-01-21
Customerly: rewrite customerly to work with playground

Innders:
## Description of changes 
Now when the playground addon is present the customerly bubble will appear.

## Technical details

Puts customerly on a global level so it's easier to enable for any other reason.


Innders:
closes https://github.com/ynput/ayon-server-playground/issues/3

PR #1016, opened 2025-02-12
Settings: Delete launcher and dependency files

flynput:
- Delete launcher and dependency files.
- Table sorting
- Table multi-selection

https://github.com/user-attachments/assets/7b366bae-86f7-4e82-b7b5-9de08540daae


martastain:
Deleting works just fine, but:

- it is super easy to accidentally delete a launcher / dp, that is a part of current production bundle, maybe the table could at least show a mark that the file is in use? (prod / staging / dev?) 

- shouldn't the buttons be labeled "manage launchers" / "manage dependency packages" when the dialog is not just about uploading anymore?

Innders:
Added bundle statuses.

Eventually this whole thing could be moved to a dedicated page 

https://github.com/ynput/ayon-frontend/issues/1018

martastain (approved):

Issue #976, opened 2025-01-15
Users: Deleting multiple users annoying


Innders:
## Story

I want to delete multiple users at once.

### To reproduce

1. Select multiple users
2. delete them
3. You have to put in every name to confirm which is annoying

### Expected behavior

You can type just "delete all".

<img width="895" alt="Image" src="https://github.com/user-attachments/assets/36fe1236-3a71-4856-a99c-388804eaaa26" />

PR #1019, opened 2025-02-17
Use `delete selected` as string when deleting multiple users instead of all usernames joined together

BigRoy:
## Description of changes 

Use `delete selected` as string when deleting multiple users instead of all usernames joined together

## Technical details

Now uses separate "delete selected" string as confirmation string in the dialog instead of the users joined with `", "`.

Fix https://github.com/ynput/ayon-frontend/issues/976

## Additional context

pass

### DIFF
--- a/src/pages/SettingsPage/UsersSettings/DeleteUserDialog.tsx
+++ b/src/pages/SettingsPage/UsersSettings/DeleteUserDialog.tsx
@@ -19,7 +19,7 @@ const DeleteUserDialog = ({ onHide, selectedUsers, onDelete, onDisable }: Delete
   }
 
   const selectedUsersString = selectedUsers.join(', ')
-  const confirmDeleteUsersString = selectedUsers?.length > 1 ? "delete selected" : selectedUsersString
+  const confirmDeleteUsersString = selectedUsers.length > 1 ? "delete selected" : selectedUsersString
   return (
     <Dialog
       size="md"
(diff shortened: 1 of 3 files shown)


src/pages/SettingsPage/UsersSettings/DeleteUserDialog.tsx
```diff
@@ -12,14 +12,14 @@ type DeleteUserDialogProps = {
...
 
-
   if (!selectedUsers?.length) {
     // this shouldn't happen
     onHide()
     return <></>
   }
 
   const selectedUsersString = selectedUsers.join(', ')
+  const confirmDeleteUsersString = selectedUsers?.length > 1 ? "delete selected" : selectedUsersString
```
BigRoy, line 22: Since `selectedUsers?.length` may be `null` (I assume?) this `> 1` check I suppose wouldn't work. Not sure what the best javascript equivalent is.

However, since above we're returning early on `!selectedUsers?.length` already I guess we're safe?

BigRoy: This may then be clearer that we're not expecting `null` selectedUsers to begin with.
```suggestion
  const confirmDeleteUsersString = selectedUsers.length > 1 ? "delete selected" : selectedUsersString
```

Innders (changes requested):
Code works but it's not obvious what you need to write in the text box to enable the delete button. Could we please add the default greyed input placeholder text of what needs to be written.

image

BigRoy:
> Code works but it's not obvious what you need to write in the text box to enable the delete button. Could we please add the default greyed input placeholder text of what needs to be written.

Should work with https://github.com/ynput/ayon-frontend/pull/1019/commits/0cf25935aa0107169d5071899b0af1a8ea1c925d

PR #1025, opened 2025-02-18
Bundle per project

martastain:


martastain:
@Innders if you could help me with this one, you'd have it much faster than me:

The idea is to have a button here:

(AddonSettings.jsx)
image
image


That would:
- open a dialog with a list of bundles, that have `isProject` flag set
- user can set or unset the bundle for the current project (using `[POST] /projects/{projectName}/bundles` ) and the current variant
- when bundle is changed, it would invalidate `addonSettingsList`

Innders (approved):

iLLiCiTiT:
I can see dev bundles in production/staging enum, and standard bundles in dev bundles enum.

image

PR #1027, opened 2025-02-18
fix: rjsf issue - work in progress

martastain:


Innders (changes requested):
There are issues uploading files with file names with spaces. Could we please sanitize the filename before uploading the file?

Issue #1030, opened 2025-02-19
User with existing "assignees" filter in Task view can't see tasks because it can't revert it to "me" filter.


BigRoy:
## Story

User with existing "assignees" filter in Task view can't see tasks because it can't revert it to "me" filter.

### To reproduce

1. Set up user with manager permissions
2. Then as that user in tasks view, set the filter to assignees.
3. Change user to "regular user"
4. As the user access task view. They now cannot see tasks nor change filter back to "me" because the box is hidden.

### Expected behavior

Filter states should be forced to "me" if the user can't access other users due to permissions.

## Environment

- **Server version**: 
- **Server host OS**: 
- **Browser**: 
- **Client version**:

PR #1031, opened 2025-02-19
Dashboard: Force assignees filter to "me" when user has no permissions to list users

BigRoy:
## Description of changes 

Force assignees filter to "me" when user has no permissions to list other users (and hence the `MeOrUserSwitch` component is hidden to switch it manually)

## Technical details

This seems to work, but this might need a `useEffect`?

```jsx
  useEffect(() => {
    if (!isManager && assigneesFilter !== "me") {
      console.log("Force assignees filter to me")
      setAssignees({
        assignees: [],
        filter: "me"
      })
    }
  }, [assigneesFilter, isManager])
```

Fix https://github.com/ynput/ayon-frontend/issues/1030

- [x] Note that the assignees and filter come from the `localStorage` and maybe (if they aren't yet) should be cleared on log out and log in to avoid the issue to begin with. Should be implemented with https://github.com/ynput/ayon-frontend/pull/1031/commits/1ab1c60a6e0ac5cdc73d1c13252c1872b21d7825


Innders (approved):
Seems to work for me! 🏅

PR #1032, opened 2025-02-19
Annotations: pass dimensions correctly for the upcoming new Power Pack version

### LINKED ISSUE #1021: annotation: scale/position of drawn annotations not aligned with image native size
## Story
Pencil drawn annotation shifts position after the comment is opened

### To reproduce

1. Resize the app window so that the rendered frame is not 100% (aka its native size)
2. Draw an annotation over a recognizable landmark location
3. Post the comment (that will include the annotation)

### Expected behavior
The annotation should be in the same relative position and scale as drawn 

## Environment

I used the current playground: 1.7.0+202501301817

### Additional context
The problem is that the annotation is composited over the background image at its native size (in this case HD) but the annotation, in certain cases, is drawn over a resized background image.

I confirmed this:
* I made the app window smaller so the 1920x1080 image was reduced in size
* I drew a red "+" in the center of the "O" in "COFFEE SHOP" (see attached image)
* The annotation shifted to the right and lower (as though it had been scaled up)

I also tested the converse, where 
* I made the app window larger, the source image was larger than HD, and, as expected, the annotations' locations/scale were smaller.

Image

jakubfiala:
## Description of changes 

- pass container dimensions (width/height) to `AnnotationsCanvas` instead of passing the `containerRef`, as per the new Power Pack version requirements
- use the actual media dimensions (i.e. video frame size and original image size) for the `AnnotationsCanvas`
- set the `ViewerImage`'s `` to `display: block` - gets rid of extra space at the bottom
- remove 2px offset for the video viewer - we believe this is no longer needed and makes the canvas fit perfectly on top of the video

## Technical details

- we are not 100% sure why the 2px offset was needed previously. We haven't been able to reproduce any issues once the offset was removed.
- canvas compositing can now take a while for larger images. This causes noticeable lag while the annotated thumbnail is being generated. To be addressed separately.

## Additional context

Should fix #1021 

The new Power Pack version should be configured so it will alert the user if they're using a version of AYON server that doesn't have these changes yet.

### DIFF
--- a/src/containers/VideoPlayer/VideoPlayer.jsx
+++ b/src/containers/VideoPlayer/VideoPlayer.jsx
@@ -94,6 +94,8 @@ const VideoPlayer = ({ src, frameRate, aspectRatio, autoplay, onPlay, reviewable
     width: null,
     height: null,
   })
+  // used to set intrinsic size for the Annotations canvas
+  const [actualVideoDimensions, setActualVideoDimensions] = useState(null)
 
   useGoToFrame({ setCurrentTime, frameRate, duration, videoElement: videoRef.current })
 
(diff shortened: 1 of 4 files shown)


Innders (changes requested):
Everything seems aligned now! 🥳 

Only thing is I break atrament when switching reviewables. It goes into this weird black mode and doesn't work. 


https://github.com/user-attachments/assets/5eb89e9e-8f8c-4d65-8872-403a94616b09

jakubfiala:
@Innders can replicate, gonna look into why that's happening.

jakubfiala:
@Innders should now be fixed via https://github.com/ynput/ayon-power-pack/pull/9/commits/6cb8de7c6d49a62ced6a5153d1891d01ec29f130

Innders (approved):

PR #1044, opened 2025-02-21
Annotations refactor

jakubfiala:


Innders (changes requested):
- [ ] Image over the top of the buttons

image

- [x] Drawing and then going to another frame too quickly causes the drawing to be lost. Maybe we should decrease the debounce time a bit?

https://github.com/user-attachments/assets/a48b311d-cf21-48bf-be1c-058d561f2dd4

src/components/CommentInput/hooks/useAnnotationsUpload.ts
```diff
@@ -1,25 +1,30 @@
...
+  const { useAnnotations } = useViewer();
+  const { exportAnnotationComposite, removeAnnotation } = useAnnotations();
 
-  const uploadAnnotations = async (annotations: Annotation[]) => {
+  const uploadAnnotations = async (annotations: AnnotationMetadata[]) => {
     try {
       const uploadPromises = annotations.map(async (annotation) => {
-        const img = annotation.compositeData
-        const blob = base64ToBlob(img)
+        const blob = await exportAnnotationComposite(annotation.id);
```
Innders, line 19: Will awaiting here cause a bottle neck?

jakubfiala:
@Innders sounds good, I've reduced the debounce delay (on latest powerpack branch) - I'll have a look at the border thing. It looks like an issue with the image aspect ratio

src/components/CommentInput/hooks/useAnnotationsUpload.ts
jakubfiala: as far as I can tell, it shouldn't - the promises are separate. However, rendering multiple composite images at the same time will likely be quite slow as it's all on 1 thread.

jakubfiala:
@Innders added a "submitting" state to CommentInput, for now it's just non-interactive and greyed out. Lmk if that feels sufficient or you'd rather have a spinner over it, too

src/components/CommentInput/CommentInput.styled.js
```diff
@@ -253,12 +253,17 @@ export const Comment = styled.div`
...
-  &.isLoading {
+  &.isLoading,
+  &.isSubmitting {
     cursor: default;
     pointer-events: none;
     user-select: none;
   }
 
+  &.isSubmitting {
+    opacity: 0.3;
```
Innders, line 264: This works well when there is a delay but it's a bit odd when this isn't a delay. I think a transition-delay on the CSS might fix this so that after 300ms or so it will go grey. WDYT?

https://github.com/user-attachments/assets/2bfc7ca2-bc67-4584-9481-7c0ec8f5a87e

Innders (changes requested):

src/components/CommentInput/CommentInput.styled.js
jakubfiala: @Innders great shout! will add

jakubfiala:
@Innders added a 250ms transition with a 250ms delay, seems like a happy medium. Lmk what you think. Also, if you pull latest powerpack, the annotated images now have unique filenames.

jakubfiala:
that should also fix the weird prefix dot. It was being added by the code that's meant to split file name from the extension, but it can't handle names without extensions :upside_down_face:

Innders (approved):
Great stuff, I'm excited to get this out for wider testing 👏

PR #1046, opened 2025-02-24
Only clear comment field if the submission resulted in a result entry

### LINKED ISSUE #1043: Comments: Restore comment on POST error
## Story

I have written a really long and emotional message. I click send and there is some error. The message disappears forever, just like my hopes and dreams.

### Expected behavior

If there is an error whilst posting the comment, the comment should be added back to the comment input box.

BigRoy:
## Description of changes 

Comment field text and content is only cleared if the submission resulted in a successful result

## Technical details

Fix #1043

### DIFF
--- a/src/services/activities/updateActivities.js
+++ b/src/services/activities/updateActivities.js
@@ -33,7 +33,7 @@ const patchActivities = async (
   const entries = api.util.selectInvalidatedBy(state, invalidatingTags)
 
   // now patch all the caches with the update
-  const patches = entries.forEach(({ originalArgs }) =>
+  const patches = entries.map(({ originalArgs }) =>
     dispatch(
       api.util.updateQueryData(
         'getActivities',
@@ -41,7 +41,7 @@ const patchActivities = async (
         (draft) => updateCache(draft.activities, patch, method === 'delete'),
       ),
     ),
-  ) || []
+  )
 
   try {
     await queryFulfilled
--- a/src/services/activities/updateActivities.js
+++ b/src/services/activities/updateActivities.js
@@ -52,9 +52,6 @@ const patchActivities = async (
     for (const patchResult of patches) {
       patchResult?.undo()
     }
-
-    // Allow components to catch failure and handle it
-    throw message
   }
 }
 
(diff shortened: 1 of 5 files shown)


Innders (changes requested):
This looks like it would work for submitting a comment but it wouldn't work for when updating a comment.

Look at the function that is passed to onSubmit in `ActivityComment.tsx - handleSave`. It doesn't return an results array.

https://github.com/ynput/ayon-frontend/blob/b952cfdb0c3d521e81266543926bcad15708674e/src/components/Feed/ActivityComment/ActivityComment.tsx#L108-L111

You will need to make sure this also returns a results array as well.

Wrap the logic in a try catch block and return results if it does not error and an empty array if it does error. Just like here -> https://github.com/ynput/ayon-frontend/blob/b952cfdb0c3d521e81266543926bcad15708674e/src/containers/Feed/hooks/useCommentMutations.js#L86-L92

- [x] Update `ActivityComment.tsx` handleSave function to return results.
- [x] Demonstrate commenting and updating working
- [x] Demonstrate commenting and updating failing the request and therefore not clearing the field.

src/services/activities/updateActivities.js
```diff
@@ -41,7 +41,7 @@ const patchActivities = async (
         (draft) => updateCache(draft.activities, patch, method === 'delete'),
       ),
     ),
-  )
+  ) || []
```
BigRoy, line 44: @Innders this was needed because `patches` would not be a list if `updateComment` failed due to e.g. too long body text. Likely because "updating" didn't directly generate something to patch? No idea. :D

Innders: We need to use .map that returns an array instead of .forEach that returns `null` and remove this empty `[]`.

Innders (changes requested):

BigRoy:
@Innders Thanks for push https://github.com/ynput/ayon-frontend/pull/1046/commits/92acc635a4b6bb6eb9399170b245d691a4ab3acb - works for me!

Innders (approved):

PR #1047, opened 2025-02-24
Re-enable video autoplay when opening the Viewer using the spacebar

jakubfiala:
## Description of changes 

- remove the `useEffect` hook which effectively disables the "autoplay on spacebar" feature of the Viewer


jakubfiala:
@Innders this seems to be doing what you said it should, but I'm not 100% sure if it's the correct behaviour

Innders (approved):
The autoplaying seems to be working for me now?

I agree that the implementation could be cleaner and this is based on legacy code. But for now it's a lower priority and good to go imo.

PR #1052, opened 2025-03-03
Settings: fix array copy in settings

martastain:
Due to missing "props.formData" dependency in useMemo for context menu items, copy function didn't work properly on arrays.
This is now fixed. Additionally a warning is now displayed if there's nothing to copy.

### Testing notes

- Right after page loads, find a setting of type `list[SomeSubModel]` in normal (collapsible) layout
- Copy the entire array by right-clicking the panel header
- paste the data somewhere

Note: This worked before when you switched to another addon and back, but not after page load.


Innders (approved):
LGTM

PR #1054, opened 2025-03-04
Settings: "Global" tab move to left so bundles and studio settings remain close to each other

BigRoy:
## Description of changes 

Settings: "Global" tab move to left so bundles and studio settings remain close to each other

image

## Technical details

Keeps the settings management where you usually play with bundles, and then studio settings close together.


Innders (approved):
I think this makes the most sense, Thanks 💯

PR #1061, opened 2025-03-12
Bundle necessary fonts to the app to support air-gapped environments

martastain:


BigRoy (approved):
Before this PR there is this network activity:
image

image

After this PR:
image

Which does not show any access to `gstatic` or `googleapis` anymore.
So I suppose this does what it should.

I do still think that e.g. `ayon-sitesync` addon frontend would still introduce a dependency on google fonts though. For example, here: https://github.com/ynput/ayon-sitesync/blob/d2d2b30f7c12a722675fcfd33cd036fafb815a21/frontend/src/index.sass#L5

But I suppose that would be a separate problem from this one.

Innders:
@martastain should this not be doner in ARC, then any addon using it would have this automatically?

martastain:
@Innders that would be ideal. But in that case, souldn't be all styles from src/styles moved there as well? where would be the line?

martastain:
Updated to pick the fonts from arc:
Requires https://github.com/ynput/ayon-react-components/pull/204

Issue #778, opened 2024-09-03
AY-6628_Sync Publish comments


ynbot:
### Additional context:
link to discussion on [Discord](https://discord.com/channels/517362899170230292/1277602462068506705/1277603264078155816)
_(might be a private channel)_

_This issue was automatically created from Clickup ticket [AY-6628](https://app.clickup.com/t/86c03v3by)_

PR #1065, opened 2025-03-18
Show version comment on the version activity in the activity feed

BigRoy:
## Description of changes 

Show version comment on the version activity in the activity feed

## Technical details

Show `version.attrib.commen` on the version activity in the activity feed

I moved the version number to after the product name, because otherwise the layout looked very odd.
Plus it matches visually with how we're display product+version on the breadcrumbs in the entity sidecard, like here:

image

Fix https://github.com/ynput/ayon-frontend/issues/778

## Additional context

Before:
image

After:
image

---
Without comment:

Before:
image

After:
image

### DIFF
--- a/src/components/Feed/ActivityVersions/ActivityVersions.jsx
+++ b/src/components/Feed/ActivityVersions/ActivityVersions.jsx
@@ -45,12 +45,10 @@ const ActivityVersions = ({
             <Styled.Card onClick={() => handleClick(id, productId)} key={id}>
               <Styled.Content>
                 <div>
-                <Styled.Title>
-                  <div style={{ lineHeight: '' }}>
-                    <span>{productName}</span><br/>
-                  </div>
-                  <ActivityDate date={createdAt} isExact />
-                </Styled.Title>
+                  <Styled.Title>
+                    <span>{productName}</span>
+                    <ActivityDate date={createdAt} isExact />
+                  </Styled.Title>
                   <span className="version">{name}</span>
                 </div>
                 <Styled.Thumbnail
--- a/src/components/Feed/ActivityVersions/ActivityVersions.jsx
+++ b/src/components/Feed/ActivityVersions/ActivityVersions.jsx
@@ -50,7 +50,6 @@ const ActivityVersions = ({
                   </div>
                   <ActivityDate date={createdAt} isExact />
                 </Styled.Title>
-
                 <Styled.Thumbnail
                   {...{ projectName }}
                   entityId={id}
--- a/src/components/Feed/ActivityVersions/ActivityVersions.jsx
+++ b/src/components/Feed/ActivityVersions/ActivityVersions.jsx
@@ -43,13 +43,15 @@ const ActivityVersions = ({
           (index < limit || showAll) && (
             <Styled.Card onClick={() => handleClick(id, productId)} key={id}>
               <Styled.Content>
+                <div>
                 <Styled.Title>
-                  <div>
+                  <div style={{ lineHeight: '' }}>
                     <span>{productName}</span><br/>
-                    <span className="version">{name}</span>
                   </div>
                   <ActivityDate date={createdAt} isExact />
                 </Styled.Title>
+                  <span className="version">{name}</span>
+                </div>
                 <Styled.Thumbnail
                   {...{ projectName }}
                   entityId={id}
--- a/src/components/Feed/ActivityVersions/ActivityVersions.jsx
+++ b/src/components/Feed/ActivityVersions/ActivityVersions.jsx
@@ -16,6 +16,7 @@ const ActivityVersions = ({
   filter,
 }) => {
   let { authorName, authorFullName, createdAt, versions = [] } = activity
+
   const [showAll, setShowAll] = useState(filter === 'publishes')
   const limit = 2
 
--- a/src/components/Feed/ActivityVersions/ActivityVersions.jsx
+++ b/src/components/Feed/ActivityVersions/ActivityVersions.jsx
@@ -3,7 +3,6 @@ import ActivityHeader from '../ActivityHeader/ActivityHeader'
 import * as Styled from './ActivityVersions.styled'
 import { useState } from 'react'
 import { More } from '../ActivityGroup/ActivityGroup.styled'
-import ActivityDate from '../ActivityDate'
 import { useDispatch } from 'react-redux'
 import { openViewer } from '@state/viewer'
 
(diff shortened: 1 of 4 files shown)


src/services/activities/activityQueries.js
```diff
@@ -117,6 +122,7 @@ export const getTypeFields = (type) => {
             path
           }
         }
+        comment
```
BigRoy, line 125: Not sure if this is needed. It seems to work without this too. 🤷 

```suggestion
```

src/components/Feed/ActivityVersions/ActivityVersions.styled.js
```diff
@@ -70,3 +70,10 @@ export const Thumbnail = styled(ThumbnailSimple)`
     object-fit: cover;
   }
 `
+
+export const Comment = styled.div`
+  margin-top: var(--base-gap-small);
+  color: var(--md-sys-color-on-surface-variant);
+  line-height: 1.25rem;
+  word-break: break-word;
+`
```
BigRoy, line 79: AI really wanted to style it like this for me and I couldn't be bothered since I was just trying to get the thing to work in the first place - please do recommend if there's a better ready component to use for the text as opposed to this. @Innders

Innders: We use `var(--md-sys-color-outline)` for the color when we want to have grey text.

line-height seems unnecessary?

src/components/Feed/ActivityVersions/ActivityVersions.jsx
```diff
@@ -38,16 +36,16 @@ const ActivityVersions = ({
...
-        ({ name, id, productId, productName, updatedAt, createdAt }, index) =>
+      {versions.flatMap((version, index) => {
+        const { name, id, productId, productName, updatedAt, comment } = version
+        return (
           (index < limit || showAll) && (
             <Styled.Card onClick={() => handleClick(id, productId)} key={id}>
               <Styled.Content>
                 <Styled.Title>
-                  <span>{productName}</span>
-                  <ActivityDate date={createdAt} isExact />
```
Innders, line 48: We seem to have lost the date here. It could still come after like name - version - date

src/components/Feed/ActivityVersions/ActivityVersions.jsx
```diff
@@ -38,16 +36,16 @@ const ActivityVersions = ({
...
+                  <span>{productName} - {name}</span>
                 </Styled.Title>
-                <span className="version">{name}</span>
+                <Styled.Comment>{comment}</Styled.Comment>
```
Innders, line 48: When there's no comment could we center all the text so that it's not floating at the top. An `align-items: center` will probably do the trick.

src/services/activities/activityQueries.js
Innders: This is only used for the tooltip (when you hover over a mention in a comment). We don't see comments on these tooltips so it can be removed safely. The comment for the activity item is coming from line 38.

Innders (changes requested):
I'm super stoked that this finally got added because I was too lazy to do it myself and it will probably make quite a big difference. Great work!

Just a couple small things to clean up.

src/components/Feed/ActivityVersions/ActivityVersions.jsx
BigRoy: Ah yes, I had removed that while I was fiddling with potentially moving the version further right - but didn't look nice at all. Will add it back.

src/services/activities/activityQueries.js
BigRoy: Thanks - yes, the `version.attrib.comment` had me on a good hunt to find to see where it was on the ActivityNode. 🕵️

src/components/Feed/ActivityVersions/ActivityVersions.styled.js
BigRoy: Would make it like this:

image

And with a regular comment below it:

image

Not sure if the grey is better or worse when comparing with regular comments. Thoughts?

src/components/Feed/ActivityVersions/ActivityVersions.jsx
BigRoy: Pushed.

src/components/Feed/ActivityVersions/ActivityVersions.styled.js
BigRoy: Pushed.

BigRoy:
Now looks like:
image

And:
image

philippe-ynput:
@BigRoy What happens when the comment string is really really long ?

BigRoy:
> something I did in the past in Qt but it would be more involved...
> 
>     * When the string is too long, cut it and add an ellipsis. That will preserve the vertical layout and may be more pleasing.
> 
>     * Provide another way to view the full string, like a tooltip or a disclosure like:
> 
> 
> > So, tell me John. What does a long comment look like? ...
> >  
> > So, tell me John. What does a long comment look like? ! really want to approve this beauty but I'm having some issues with previewing how bigger things look. I know it's not about the size of things, but about how you use it. But in this case, it's crucial we do an accurate size comparison and review what it looks like. So, take your time and prepare... but make sure I get a good size reference.
> 
> ```
> <details>
>   <summary>So, tell me John. What does a long comment look like? ...</summary>
>   <p>
> So, tell me John. What does a long comment look like? ! really want to approve this beauty but I'm having some issues with previewing how bigger things look. I know it's not about the size of things, but about how you use it. But in this case, it's crucial we do an accurate size comparison and review what it looks like. So, take your time and prepare... but make sure I get a good size reference.
>   </p>
> </details>
> ```

In this case I'd argue that directly seeing the full comment is more important - I've discussed with @Innders and opted moving to a design like this:
image

I just need to fix some minor things that I dislike, like  how close the version number is to the product name.. I want basically the style from before instead, like this:

image

First I need to do something else - and then I need to briefly investigate how I can best set this up. I'm learning along the way hehe.

src/components/Feed/ActivityVersions/ActivityVersions.styled.js
```diff
@@ -58,7 +55,8 @@ export const Title = styled.div`
 
 export const Thumbnail = styled(ThumbnailSimple)`
   width: 74px;
-  height: 100%;
+  min-width: 74px;
+  height: unset;
```
BigRoy, line 59: Technically without the comment now being next to the thumbnail this could be reverted again. But it may still be helpful for cases where e.g. product name is super long and with the "date" inline after it that the thumbnail for whatever reason would become super narrow - with this it'd now remain consistent size.

PR #1089, opened 2025-04-06
Pipeline Updater: Bundle name uses studio name or "Bundle"

Innders:
## Description of changes 
The new bundle created by the pipeline updater should be named with the release version not release name or "Bundle" if there is no name for any reason.

This ensures that the release suffix "kitsu", "ftrack", "shotgrid" is not included in the bundle name.


Innders:
fixes #1089

PR #1090, opened 2025-04-07
Settings: Select all option for dropdowns with more than 10 options

Innders:
## Description of changes 
Sometimes you just want to select all of the options for enum field in the settings. This is super annoying if there are lots of items to select.

This PR adds the "Select all" button if there are more than 10 options and the dropdown is multi-select.

image


martastain (approved):
Works!

PR #1094, opened 2025-04-09
Services: Service details panel and service editing

Innders:
## Description of changes 
- A details panel to view all data about a service.
- Ability to edit the service.

## Technical details

- Only the host and volumes are currently editable.
- When editing, the service must first be stopped and then run again after updating.

## Additional context

https://github.com/user-attachments/assets/a047f2ab-7b59-424b-9848-0863bea3b71f


martastain (approved):
Works with https://github.com/ynput/ayon-backend/pull/573

PR #1095, opened 2025-04-09
Use connected flag of /api/connect response

martastain:
This pull request includes a couple of changes to the `YnputConnector` component in `src/components/YnputCloud/YnputConnector.jsx` to improve the handling of connection data and error states.

Instead of relying on error code from /api/connect, ynputconnector now checks the boolean flag "connected" in the response that is always 200 as of https://github.com/ynput/ayon-backend/pull/574

That effectively enables displaying license dialog even if the instance is not connected to ynput cloud and enables viewing locally installed licenses.

image


Innders (changes requested):
There are a couple more places where `useGetYnputConnectionsQuery` is used in components that rely on the result being `truthy` or `null` to detect if the instance is connected to ynput cloud. 

Now that the result is always `truthy` these also need to use `.connected` now instead.

Innders (approved):

PR #1134, opened 2025-04-28
Entity lists

Innders:
- Adds a new page called "Lists" inside of a project.
- Adds the ability to add entities to a list from the overview, task progress and browser pages.

## What are lists?

A list can be an arbitrary collection of a single entity type. For example it could be 10 top priority tasks or 5 versions that need review.

A list must contain the same entityType, `folder`, `task` or `version` and this must be specified when creating the list. Note: `product` is supported in the API but is not currently supported on the client.

Lists can have a label set by the user and will support tags and categories in future iterations.

## List attributes (Power feature⚡ )

Each list can have it's own set of custom attributes created and used only the for that list. This makes list extremely powerful as they can be used for storing extra data on entities that it completely arbitrary and doesn't require studio or project wide changes.

## List reviews (requires Review Addon)

The `entityListType` field defines the type of a list and what it's main purpose is for. **review-session** lists are collection of versions compatible with the upcoming review addon. This gives the ability to open a set number of versions in a reviewable timeline.

## Future iterations

This is the first iteration of lists and we hope to add more features in the future.

- List categories for better organisation.
- List overrides to staging new values without committing them to the main entity (⚡ power feature).
- Filter by list on Overview, Task Progress and Browser page.
- Create new lists from right click menu.
- Create new review session from a list (requires Review Addon)
- Reordering of lists and list items

## Notes

- List power features require Powerpack Addon version `1.0.5` or above.


martastain (approved):
LGTM :D

No... really.... i tested it and it works nicely :)

PR #1159, opened 2025-05-09
Rename anatomy presets

martastain:
This pull request introduces a new feature to rename anatomy presets and refactors the preset creation dialog to improve usability and maintainability. The changes include the addition of a reusable `PresetNameDialog` component, new API support for renaming presets, and updates to the UI to integrate these enhancements.

### DIFF
--- a/src/services/anatomy/updateAnatomy.js
+++ b/src/services/anatomy/updateAnatomy.js
@@ -23,6 +23,17 @@ const getAnatomy = api.injectEndpoints({
         { type: 'anatomyPresets', id: 'LIST' },
       ],
     }),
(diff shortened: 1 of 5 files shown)


src/services/anatomy/updateAnatomy.js
Innders, line 1: Here we are injecting the endpoint when it already exists from the generated files.

**Really** this whole file should be migrated to use `enhanceEndpoints` where queries are already generated.

But let's at least use enhancedEndpoints for the new `renamePreset`.

Innders (changes requested):

Innders (approved):

PR #1160, opened 2025-05-09
Truncate last path segment in DetailsPanel to avoid panel width being affected

jakubfiala:
## Description of changes 

Sometimes even after some of the path segments are collapsed, the resulting path is still too wide to fit inside the panel width. This PR adds an extra layout effect which truncates the last segment's text to compensate for this.

## Technical details

It's possible sometimes the reduction will over-correct, because I wanted to avoid calculating the exact remaining path width again - I'm just using the current width at the time when the layout effect runs.

## Additional context

image


Innders (approved):
Much needed fix, thanks!

PR #1164, opened 2025-05-09
Fix: overview cannot copy empty array and clear field option

martastain:
This pull request introduces several changes to improve the handling of clipboard data, date values, and unrecognized types in the `ProjectTreeTable` component. The key updates include allowing empty clipboard text, refining how empty or invalid values are processed, and adding safeguards for unrecognized data types.


Innders (approved):
LGTM!

PR #1174, opened 2025-05-14
Project-level actions

martastain:
This pull request introduces several updates across multiple files to enhance UI styling, improve functionality for project-level actions, and add support for reloading data when hierarchy changes.


shared/src/hooks/useActionTriggers.ts
```diff
@@ -107,6 +112,12 @@ export const useActionTriggers = ({
         })
       }
     }
+
+    if (payload.extra_reload === true) {
+      dispatch(foldersApi.util.invalidateTags(['hierarchy']))
```
Innders, line 117: Can the addon action provide which tags to invalidate? Or alternatively invalidate the whole api cache?

martastain: I'd say it should eventually refetch the data relevant to the page you are on.

Innders: Actually this is already something that lists needs. When creating a new review session, we need to invalidate the lists cache

PR #1175, opened 2025-05-15
Fix fetching tooltip data for mentions

martastain:
This PR fixes a bug in the mention tooltip data fetching, that used the feed context entity type instead of the type of the hovered mentioned entity. That caused malformed requests and log flooding in the case when for example mentioned user tooltip was displayed in the version entity feed and data for `version id=username` was requested.


Innders (approved):
LGTM

PR #1184, opened 2025-05-22
Annotations storage improvements

jakubfiala:
## Description of changes 

The DetailsPanel component now handles annotations in a smarter way. Annotations are saved as 2 separate files - composite and transparent, alongside a bit of metadata that lets us reconstruct the annotation as necessary.

When Activities are loaded, the files which belong to the same annotation are "merged" and presented as a single attachment. This attachment then has the ability to jump to the video frame where the annotation range starts. Also, the DetailsPanelContext now exposes a `feedAnnotations` array which lets the consumer app display all the annotations currently in the feed.

### DIFF
--- a/shared/src/containers/Feed/Feed.tsx
+++ b/shared/src/containers/Feed/Feed.tsx
@@ -1,4 +1,4 @@
-import { useMemo, useRef } from 'react'
+import { useEffect, useMemo, useRef } from 'react'
 import ActivityItem from './components/ActivityItem'
 import CommentInput from './components/CommentInput/CommentInput'
 import * as Styled from './Feed.styled'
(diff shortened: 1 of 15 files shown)


Innders (changes requested):
In ayon-frontend we aren't supporting this yet so can we please hide the play button.

image

Fix optimistic updates causing flashing when submitting comment.

For the hover effect, remove the buttons background, space out the icons evenly and highlight the hovered icon with opacity 100% and `font-variation-settings: "FILL" 1, "wght" 400, "GRAD" 200, "opsz" 20;`.

You shouldn't have to hover directly on the icon, just in it's half of the attachment.

image

Innders (approved):
Nice one 👍

PR #1192, opened 2025-05-26
Create new product and versions (simple)

Innders:
## Description of changes 
- Create a new product and version together or create a new version on an existing product.
- In the Browser page use the context menu on a folder or an existing version.
- Upload reviewables with the version.

image

image

## Next iteration
- Improve the UX by allowing you to drop files on folders and products to invoke the upload dialog quicker.
- Automatically set the thumbnail for the new version

### DIFF
--- a/src/services/product/getProduct.js
+++ b/src/services/product/getProduct.js
@@ -209,7 +209,7 @@ export const getProductApi = api.injectEndpoints({
         },
       }),
       transformResponse: (response) => parseVersionsData(response.data) || [],
-      providesTags: (result, _e, { ids }) =>
+      providesTags: (result, _e, { ids = [] }) =>
         result
           ? [
               ...result.map(({ versionId }) => ({ type: 'version', id: versionId })), // all version tags with id


Innders:
> Instead, you specify a "variant" and then based on anatomy templates and settings that defines the resulting Product Name.

Are these set up by default? The whole point of this PR to get studios and running with some production tracking software without any pipeline. You need to be able know 0 about the pipeline to get going.

BigRoy:
> Are these set up by default? The whole point of this PR to get studios and running with some production tracking software without any pipeline. You need to be able know 0 about the pipeline to get going.

The Product Name Profiles are based on `ayon+settings://core/tools/creator/product_name_profiles`

image

In the end the 'user' still just picks a product type, and a "name". It's just that this _name_ goes through the templating system to define a full "product name" based on the project settings.

MustafaJafar:
I like the idea of creating and adding product versions dierctly from the web ui.

> The whole point of this PR to get studios and running with some production tracking software without any pipeline. You need to be able know 0 about the pipeline to get going.

Also, I agree with you.
But let me add my 2 cents.

--- 

If AYON is a product tracker only, then we won't need to worry as much.
But, The issue here that we actually have a pipeline line therefore we need to follow it.

This [page](https://help.ayon.app/articles/7070980-about-ayon-pipeline) contains some info about AYON pipeline, However it explains the topic for the perspective of using our desktop tools for publishing.
Breifly, 
- Creating product is not the same as publishing which includes more steps.
- when creating products we have: 
    1. Naming conventions
    2. Move file to proper publish location 
    3. Some basic validations to run
    4. A version can include multiple representations. e.g. a model product version can have `.abc` model and `.mp4` reviewable.

At the moment there are two generic ways (without using a DCC) to publish files to AYON:
1. [From Client machine] Using [Tray Publisher ](https://help.ayon.app/articles/5385780-working-with-ayon-tray-publisher).
image
2. [From Web UI] Using Web Publisher. 
image


and now, this PR adds an additional way which may lead to inconsistency or make it harder when maintaining these several ways over time.

PR #1200, opened 2025-05-28
HOTFIX: handle the case where annotations metadata is mysteriously there but undefined

jakubfiala:
## Description of changes 

Add some defensive checks for when annotation metadata array attached to an Activity is loaded, but an item in the array is undefined. This seems to happen for regular attachments.

## Additional context

I will try to find out why this is happening in the first place, but this should fix the error at least.


martastain (approved):
Works

PR #1203, opened 2025-05-28
Improve how annotations are displayed in the Feed

jakubfiala:
## Description of changes 

- fixed a wrong offset when skipping to the annotation's frame
- if a comment attachment is an annotation, we now don't show the file name. Instead, we show a small annotation icon and the frame range

image

## Technical details

- for an unsaved annotation, instead of just passing a boolean `isUnsavedAnnotation` to `FileUploadCard`, we pass all the annotation data. This way, even unsaved annotations can show the range instead of file name.


BigRoy:
This may look nicer if the icon is bottom left and the frame range is bottom right. Somehow this icon really distracts me from being able to quickly read the frame range. (It's quite a busy icon)

PR #1235, opened 2025-06-12
Add project name column to inbox

### LINKED ISSUE #953: Can project name/code in the Inbox messages
When we have more than 2 projects running, can the Inbox messages add the project code/name in the messages?
Image

martastain:
This pull request introduces a visual enhancement to the `InboxMessage` component by adding a new `project-name` element alongside the existing `title` element and styling it for improved readability and consistency. The changes include updates to the JSX structure and corresponding CSS styles.

image


Innders (changes requested):
I love the idea and the project name should definitely be included.

When testing on my smaller screen this new column was causing some issues as it takes up a lot of room.

The project name is more secondary information and so I think we can afford to place it somewhere else that's less space hogging.

What do you think about simply putting it at the end of the first column with the outline colour?

image

1. It's more compact and takes up less horizontal space
2. The slight text colour difference indicates it's secondary information whilst also making it easy to pick out.

BigRoy:
> Idk... that way it would like american date formats. i'm not a UX expert, but from the semantic point of view, i'd respect the cardinality of the information. and big endian > little endian, because it's big :)

I actually agree here - project name, even though likely repetitive in the inbox seems like the initial thing to "filter by" when going over my notifications.

martastain:
I like that actually. It's busy, but i am able to filter it

PR #1243, opened 2025-06-16
overview: Fixed an issue where changes did not update on other pages

Innders:
We now properly invalidate the correct tags for task progress and user dashboard to ensure they are always up to date with the overview.


alainxi:
Thanks, now a tag change in Overview is reflected in Task Progress when I filter using the same tag.

But, if the tag change was done on the  Task Progress itself, then it is not taken into account when I filter using the same tag (also on Task Progress).

PR #1259, opened 2025-06-17
Disable password login option for users

martastain:
This is a frontend counterpart to https://github.com/ynput/ayon-backend/pull/633

This pull request introduces a new `disablePasswordLogin` field to manage user authentication settings, updates the GraphQL schema, and integrates the field into the user management UI and API queries.

### DIFF
--- a/gen/graphql.schema.json
+++ b/gen/graphql.schema.json
@@ -11986,6 +11986,22 @@
             "deprecationReason": null
           },
           {
+            "name": "disablePasswordLogin",
+            "description": null,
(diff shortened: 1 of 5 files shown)


Innders:
@MustafaJafar could we get some updates docs on this please. I'm not sure if we have a users page yet, but we should have one.

MustafaJafar:
> @MustafaJafar could we get some updates docs on this please. I'm not sure if we have a users page yet, but we should have one.

Sure thing!

PR #1301, opened 2025-07-01
Project lists redesign

Innders:
## Description of changes 
Brings the projects lists design in line with the slicer using the improved table component.

It also features an improved header with a new search functionality to quickly find projects.

The old collapsable menu has been replaced with a dynamic resizer with project names switching code names when the width gets small enough.

image
image
image
image


martastain (changes requested):
Switching from Kanban or overview to planner tab crashes the page (works in develop)

image


Switching tabs in project settings causes strobes as the project list changes its width.... it's cosmetics, but not very nice

https://github.com/user-attachments/assets/704ac77d-9ccb-4da5-a706-fba258166abf

martastain:
Multiselection of projects should be disabled in project settings page - it is not supported for any of the pages IIRC and it translates to malformed requests:

image

PR #1304, opened 2025-07-03
Review: Show splash screen when instance does not have the review addon

Innders:
## Description of changes 

When the user does not have the review addon installed show a splash screen. This means the Review tab is always present.

image

When clicking on "Try for free" the support chat will open with a pre-populated message with the format `I would like to know how I can try out the "{feature_name} {feature_type}?`

This is updated for the power features dialog in general and show work the same way when encountering a Power Feature the user does not have access to.


martastain:
LGTM

martastain (approved):

Innders:
Fixes #1302

PR #1328, opened 2025-07-14
Expose ports for new service

antirotor:
## Description of changes 
Expose service port settings to user.

## Technical details
This is just exposing `ports` already available in [API](https://docs.ayon.dev/api#tag/Services/operation/spawn_service) to user when creating new service.

## Additional context
Question, if we should expose others - (`volumes` are already exposed, but `memLimit`, `user`, `env` and `storagePath` are not. Exposing all of them is of course debatable.

Ports are handled similar to volumes now:

image

(don't mind the values in the screenshot)

With this changes, the new service is saved correctly but I suspect the ports value is not used when spawning (this needs to be investigated but it doesn't affect this PR).

Last question - from the usability point of view, we should maybe display exposed ports in service list view as it might impose some security risks and so on - but on the other hand running services with ports will be pretty limited and so to be considered as an edge case?


antirotor:
To test it, you need also https://github.com/ynput/ash/pull/15

antirotor:
> Just to be sure, you can add multiple ports but they must be separated by new lines?

yes, but I haven't tested more complex scenarios like `1234/tcp: [2000, 2001]` and so on. It should work with single ports and ports pairs on individual lines.

PR #1378, opened 2025-08-13
Add HelpButton component to ProjectPage

filipvnencak:
## Description of changes 
Implemented a contextual Help button feature for the ProjectPage. The button uses the useFeedback hook to provide support, either by opening specific help articles or by creating a pre-filled support request based on the current page.

This change involved:

- Creating a new HelpButton React component.
- Adding a helper module (helpArticles.ts) to configure help article IDs and fallback messages.
- Integrating the HelpButton into the ProjectPage navigation.
- Updating AppNavLinks to provide the current page context.

### DIFF
--- a/src/components/HelpButton/HelpButton.tsx
+++ b/src/components/HelpButton/HelpButton.tsx
@@ -1,0 +1,37 @@
+import React from 'react'
+import { Button } from '@ynput/ayon-react-components'
+import { useFeedback } from '@shared/components'
+import { getHelpForPage } from '@helpers/helpArticles'
+
+interface HelpButtonProps {
+    module: string
+    pageName?: string
+    className?: string
+}
+
+const HelpButton: React.FC<HelpButtonProps> = ({ module, pageName, className }) => {
+    const { openSupport } = useFeedback()
+    const handleHelpClick = () => {
+        const help = getHelpForPage(module, pageName)
+
+        if (help.articleId) {
+            // Open specific help article
+            openSupport('ShowArticle', help.articleId)
+        } else {
+            // Open support with a prefilled message
+            openSupport('NewMessage', help.fallbackMessage)
+        }
+    }
+
+    return (
+        <Button
+            icon="help"
+            variant="text"
+            onClick={handleHelpClick}
+            className={className}
+            data-tooltip="Get help for this page"
+        />
+    )
+}
+
+export default HelpButton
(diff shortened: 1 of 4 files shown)


Innders (changes requested):
Cool this is a good start. There are a couple of changes that we need to make.

- [x] This should now be added to all pages, even outside of projects.
- [x] When the page is too narrow the right hand side buttons should be fixed. At the moment they disappear. The tabs should still be scrollable.

Image

We now have documentation for the following pages.

- [x] - project - review - https://help.ayon.app/articles/8669165-review-sessions
- [x] - planner (project and studio as it appears in both places) - https://help.ayon.app/articles/8736562-planner-addon
- [x] - Tasks (home) - https://help.ayon.app/articles/2408349-tasks-home-page
- [x] - Project settings - teams - https://help.ayon.app/articles/6300319-teams
- [x] - Project - Browser - https://help.ayon.app/articles/7648006-browser-page
- [x] - Studio settings- Bundles - https://help.ayon.app/articles/4644998-bundles-and-addons
- [x] - Studio settings - permissions - https://help.ayon.app/articles/2635883-user-permissions-and-project-access-groups
- [x] - Project settings - project access - https://help.ayon.app/articles/2635883-user-permissions-and-project-access-groups
- [x] - Studio settings - attributes - https://help.ayon.app/articles/4636995-attributes

Innders (changes requested):
- [x] Remove line introduced under the active tab.

Image

- [ ] When the tab is an addon the featurebase message is missing the addon name.

Image

- [ ] Support article is missing for both project scoped planner addon and studio scoped planner addon.

Image

Image

- [ ] On the home page, if the tab is not recognised then it will still show the dashboard article instead of the generated support message.

Image

- [ ] There is a bug on the project settings page where the article id will get stuck and not update based on the tab. For example if you come from the home page to bundles it will show the fallback message even when there is an article. Refreshing the page will now show the bundles article correctly but navigating to any other tab, it will continue to show bundles article.

- [ ] Missing article for project settings / project access. https://help.ayon.app/articles/2635883-user-permissions-and-project-access-groups

I like the idea of adding those extra pages but because they don't have articles right now lets remove it as it's not consistent with our current design. We are redesigning the navbar so maybe that will be a universal place where this help button can live.

- [ ] Remove from market, events, services

Innders:
> On the /settings/studio I'm not sure if the {page} in chat message is well picked, now using module for example here

The goal is to write an article for every page. So if the fallback support message is rubbish it's even more an incentive to create the article. So don't worry about it.

PR #1391, opened 2025-08-18
Move folder and task to a different folder

filipvnencak:
…PickerDialog portal integration.

## Description of changes 

- Added "Move" context menu item for folders and tasks in project tree table
- Implemented optimistic UI updates for immediate user feedback
- Added EntityPickerDialog integration for destination folder selection


Innders (changes requested):
This is a good start and it works nicely. I have made a lot of notes about how I think it should be structured. I think main logic should be extracted out of useCellContextMenu so that it can be more easily reused. One day we might have drag and drop feature for moving items.

- [x] Extract logic out of context menu hook into own place.

It is possible to move a task to a folder that has the same task name. This is not allowed in AYON. All sibling tasks must have unique names. Doing this should fail and show error toast. In the picker the folder should be greyed out and disabled with suffix message "Already contains {task_name}".

- [x] Prevent moving task into a folder that already has task with identical name and show toast error.
- [ ] Extend EntityPicker to accept disabled ids and disabled message.

There is no way currently to move the folder to the root of the project (no parentId). Only folders can be moved to the root, tasks must have a folder parent.

- [ ] EntityPicker should have passed down footer button {Move to root} for this special use case.

- [ ] We must be able to move multiple folders at the same time.
- [x] When moving an entity expand it's new location so we can see it.
- [ ] Once a folder has published content in it from the pipeline (versions) it should not be movable, talk to Martin about this.

martastain (approved):
Works great. There's a backend a problem in the backend tho: when a child folder has versions, it does not mark the parent folders with `hasVersions:true` so it is possible to move larger branches with published versions. I should fix that before merging

Innders:
@filipvnencak @martastain 

Please note for next time:

1. We try not to use redux slices anymore and moving away from them. Use react context instead where possible.
2. A file in the shared folder cannot import outside of the shared folder. This PR was doing this and it broke the shared package build.
3. I have refactored the code to use react context instead and moved to the shared folder so it can be used inside of ProjectTreeTable.

PR #1403, opened 2025-08-23
1392 - Fix video player

redmonkez12:
## Description of changes 
* Added a polyfill for the requestVideoFrameCallback API to ensure video player functionality works correctly in older browsers that don't support this native API

## Additional context
* This bug solves the problem for older browsers like Firefox 131 and older or IE
* Latest browsers are not affected


martastain (approved):
I can't confirm the fix for older browsers as i don't have any available, but it does not seem to break the newer ones. Let's summon @BigRoy as the master of weird browsers.

PR #1406, opened 2025-08-25
1249 dropdown actions

redmonkez12:

### DIFF
--- a/shared/src/containers/DetailsPanel/DetailsPanel.tsx
+++ b/shared/src/containers/DetailsPanel/DetailsPanel.tsx
@@ -1,14 +1,6 @@
-import { Button, Panel, Dropdown, DefaultValueTemplate } from '@ynput/ayon-react-components'
-import React, { useEffect, useMemo, useState } from 'react'
-import styled from 'styled-components'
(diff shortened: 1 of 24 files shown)


martastain:
It works, but i wouldn't merge it RN because of the mentioned duplicates. not just PIP icon, but i don't think upload thumbnail/version should be at two different places. It's kinda confusing. Let's wait for @Innders

Innders (changes requested):
Cool, it's looking good!

"Add to list" should show a sub menu of lists to add the entity to and the option to create a new list like in the context menu. I _think_ the menu component supports sub menus but I haven't tried it in a while.

Image

The "View Data" item should always be the last item and should have label "View data" only.

Innders (changes requested):
- [ ] Use `more_horiz` icon instead (sorry my bad).
- [ ] When the menu is open keep the dropdown button highlighted.
- [x] Dragging and dropping files to either upload a thumbnail or create a new version is broken. You should be able to drag and drop a file onto the header to do this.
- [x] I can't seem to ever see the add to list item, is this supported?

Innders (changes requested):
Styling completely different now?

Image

PR #1407, opened 2025-08-25
1355: Table: Clickable links

filipvnencak:
## Description

  Added clickable URL support to table cells in ProjectTreeTable

- URL Detection: Automatically detects HTTP/HTTPS URLs in table cell content
- Hyperlink Rendering: URLs are displayed as clickable blue links with hover underline effects
- Mixed Content Support: Handles both pure URLs and text containing URLs (e.g., "Visit https://example.com for details")
- Link Behavior: URLs open in new tab (_blank) with security attributes (noopener noreferrer)
- Text Truncation: Long URLs are truncated with ellipsis (...) to fit cell width

- Cell Editing: Users can still edit cells by clicking on non-URL areas
- Event Handling: Link clicks use stopPropagation() to prevent interference with table selection
- Existing Behavior: Non-URL content remains unchanged with full editing capability

## Technical details

- Modified TextWidget.tsx in shared/src/containers/ProjectTreeTable/widgets/
- Uses regex pattern /(https?:\/\/\S+)/gi for URL detection
- Splits mixed content to render URLs as <a> tags and text as regular content
- Maintains type safety with proper string conversion handling

### Examples:

- "https://example.com" → Clickable blue link
- "Check https://docs.com and https://api.com" → Two clickable links within text
- "Regular text" → Normal editable text (unchanged)

### DIFF
--- a/shared/src/containers/ProjectTreeTable/widgets/TextWidget.tsx
+++ b/shared/src/containers/ProjectTreeTable/widgets/TextWidget.tsx
@@ -1,4 +1,4 @@
-import { forwardRef } from 'react'
+import { forwardRef, useCallback } from 'react'
 import { TextWidgetInput } from './TextWidgetInput'
 import { WidgetBaseProps } from './CellWidget'
 import styled from 'styled-components'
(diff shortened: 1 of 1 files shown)


shared/src/containers/ProjectTreeTable/widgets/TextWidget.tsx
```diff
@@ -15,6 +15,43 @@ const StyledBaseTextWidget = styled.span`
...
+    return url.protocol === 'http:' || url.protocol === 'https:'
+  } catch {
+    return false
+  }
+}
+
+// Function to parse text and extract URLs
+const parseTextWithUrls = (text: string) => {
+  // Regex to match HTTP/HTTPS URLs
+  const urlRegex = /(https?:\/\/\S+)/gi
```
redmonkez12, line 44: Not sure if you want this. If you have `https://example.com.` It will match dot too.

If you don't want to use that dot you can use something like this - `(https?:\/\/[^\s<>"{}|\\^`\[\]]*[^\s<>"{}|\\^`\[\]\.])`

martastain, line 44:
```suggestion
  const urlRegex = /(https?:\/\/[^\s<>"]+[^\s.,!?;:<>\")\]])/gi;
```

Issue #1352, opened 2025-07-30
Table: improved string fields (like description)


Innders:
- Hover over the cell for 600ms to see the full description.
- Double click to open text editing modal.
- A sensible max width and height.
- First iteration just shows text as raw string.
- Second iteration parses the markdown to rich text to be edited using shortcuts like ctrl+b for bold etc.

<img width="981" height="432" alt="Image" src="https://github.com/user-attachments/assets/314c8929-6a90-4fac-a6c7-e9d70c810e00" />

<img width="988" height="447" alt="Image" src="https://github.com/user-attachments/assets/45bd47ba-491d-46f3-a1e8-f22a7bd494c4" />

### Notes
- This should be a new Widget.
- You will probably need to use a portal.
- You can use <CellEditingDialog />
- Look at how LinksWidget and NameWidget a done.

PR #1416, opened 2025-08-28
Entity panel: Description formatting support and improved attributes

redmonkez12:
## Description of changes 
This pull request renames Attributes to Details in the details pane. It has a description, attributes and a details section. It's possible edit/create a comment

## Additional context

Screenshot 2025-08-28 at 9 27 02
Screenshot 2025-08-28 at 9 27 15
Screenshot 2025-08-28 at 9 27 19

Update - Added error handling for user when inserted invalid value

Screenshot 2025-09-01 at 10 10 54


Innders (changes requested):
Great start, functionally it works, just need to make some styling fixes.

- [ ] When editing an input align the text to the left so it matches when it's in read mode.

Image

- [ ] The tags in the header should have the colours flipped so that the colour is the background with white text so that it matches the dropdown.

Image

- [ ] Padding/margin at the top to have a bit of the gap between the description and the top.

Image

- [ ] Remove mentions as they do not work
- [ ] Remove the background for the footer to match the figma design

Image

- [ ] When editing remove the description header and replace with the toolbar.
- [ ] Remove the checklist tool as it won't be used (yet).

Image

- [ ]  unordered list shows are ordered list.

Image

- [ ] There is this weird bug with lists and headers where every time you save and edit them they get more spaced out in editor view. Just in general the markdown is kind of buggy and headers don't work.

https://github.com/user-attachments/assets/4202c3f4-02e9-4338-a8c1-b1827d264ffd

- [ ] min height of the description box when reading and editing should be 300px (or around that but they should be the same).
- [ ] Remove font-family: monospace; from details section.
- [ ] The copy to clipboard button should be square 32px/32px

Innders (changes requested):
Great job on the text editing!

- [ ] Padding on the attributes is now broken.

Image

- [x] There should be a 4px gap between the attribute rows.

Image

- [x] Remove extra hover effect on the dropdown chevron button.

Image

- [x] When the description is active, don't use the hover effect.

Image

- [x] When not editing: hovering the link should underline it and then clicking should open it in another tab.

Image

Innders (approved):
Works really well, thanks for the hard work! 👍 

Because it's such a big PR, I will just get @martastain to give it a quick test.

Innders (changes requested):
I just noticed a small thing.

The table description is rendering the markdown string when it needs to be formatted to markdown. This should be as simple as wrapping string cells with <Markdown>.

You might need to watch out for maintaining support for links though!

Image

First part of this issue, which is rendering string cell content as markdown (don't need to do rest of it though). https://github.com/ynput/ayon-frontend/issues/1352

martastain (commented):
## Tags

When a no tag is defined in the anatomy, a text widget is shown and the patch request tries to update tags field with a string value, but a list of strings is expected. Also the v

You can add a "blind" tag, that is not defined in the anatomy (it won't have a custom color), but the value for the patch request needs to be a list.

Image

Image

Also the error message is quite cryptic: we provide detailed error information in the response:

Image

Btw: would it be possible to get the lightness of the tag color and in the case color is too bright, use a dark foreground color instead of white?

Image

## Deadlocks

When you try to run another operation to update the same field and the previous one was not completed yet, the database cannot get the exclusive lock for the row, which ends up in deadlock. It is a rare case, but can happen in the case of boolean attributes you can click manically. Maybe the widget should be locked until the response from the server is received. Also, it would probably make sense to return a specific error code in that case, so in the frontend, you could show a non-intimidating error message, such as "this field is being updated right now, please try again"

Image

## Nitpicks

The field labels in the details section should probably use proper casing and spaces:
"ID", "Created at".... 

Image

Also it would be nice if datetimes are formated using the browser locale.

Innders (changes requested):
Please address these issues made by Martin

- [x] Tags widget issue with patch request

> When a no tag is defined in the anatomy, a text widget is shown and the patch request tries to update tags field with a string value, but a list of strings is expected. Also the v

> You can add a "blind" tag, that is not defined in the anatomy (it won't have a custom color), but the value for the patch request needs to be a list.

- [ ] Editing a boolean value causes two requests to be made at the same time and a deadlock to happen. (check if this still happens)

- [ ] Attribute and metadata labels should use nice casing `CreatedAt -> Created at`

> The field labels in the details section should probably use proper casing and spaces:
"ID", "Created at"....

Innders (approved):
I think we are finally there. Thank you for the hard work!!

There are a couple small things but I will create seperate issues for them so we can get this merged.

PR #1462, opened 2025-09-07
Overview: Fix styling issues with table column menu

redmonkez12:
## Description of changes 
* This pull request solves a couple of bugs in the overview page
* It fixes the rotating effect when clicking on the more_horiz icon 
* Changed styles of dropdown, like we have in the projects dropdown
* When the column is sorted, we always display the sorting icon

https://github.com/user-attachments/assets/572aba04-9d71-4c82-9d16-411902a8526d

Screenshot 2025-09-07 at 11 46 18
Screenshot 2025-09-07 at 11 46 31

### DIFF
--- a/shared/src/containers/ProjectTreeTable/components/HeaderActionButton.tsx
+++ b/shared/src/containers/ProjectTreeTable/components/HeaderActionButton.tsx
@@ -15,7 +15,7 @@ const ActionButton = styled(Button)`
   }
 
   &:hover {
-    background-color: var(--md-sys-color-surface-container);
+    background-color: var(--md-sys-color-surface-container-hover) !important;
   }
 
   &.selected {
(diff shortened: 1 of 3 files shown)


Innders (changes requested):
Everything that was request was fixed, nice one 👍 Now I'm going to be annoying and ask for some extra things that will make this feature super nice to use 😏 

- [ ] The button sizes of the dropdown menu and sorting menu should match. Both should have 2px padding with a 24x24px size.

Image

- [ ] When the menu is open, maintain the hover effect on the dropdown button.

Image

- [ ] When a column has something active like sorting or pinning then the item should be highlighted with

```css
background: var(--md-sys-color-primary-container);
color: var(--md-sys-color-on-primary-container);
```

Image

- [ ] Menu goes off screen

Image

- [ ] Ascending and descending icons should follow the direction. This also needs fixing in the table header as they currently backwards.
image

PR #1473, opened 2025-09-09
Overview: When copy and paste includes hidden columns

filipvnencak:
## Description of changes 
Pass `visibleColumns` to `ClipboardProvider` and update copy logic to include only currently visible columns in the correct order.

## Technical details

- Copy logic now filters out hidden columns  
- Column order preserved via `visibleColumns.map`  
- Updated `ClipboardProviderProps` to include `visibleColumns`


filipvnencak:
@Innders please check, I tried some ways to copy and paste and its working when some columns are hidden, but I'm not sure if I covered everything.

PR #1479, opened 2025-09-10
Lists: List folders and details panel

Innders:
### List Details panel
- Double click a list to open it details panel for more information.
- List attributes can be updated like the name, tags and category. List metadata details can also more easily be shown.

### List Folders

- Create folders for lists
- Folders can have colors, icons and scope

## Additional context

This is the first step to adding more options to lists and review sessions, like sharing access and review settings.

image


martastain (changes requested):
Maybe i'd change "Value" to  "Name" with a tooltip "unique name of the catetgory". Uniqueness is checked upon saving now, which closes the dialog - it should probably give the user an option to fix the name without filling label, color and icon again.

Image

Otherwise it works really great

PR #1484, opened 2025-09-12
Edit entity label inside of table

filipvnencak:
## Description of changes

  Added inline editing for folder/task names in ProjectTreeTable with improved name validation.

  Features:
  - Edit display label and system name in dual input fields
  - Auto-generates system name from label (e.g., "My Folder" → "my_folder")
  - Manual override available for system names
  - Enhanced name validation with regex pattern: `^[a-zA-Z0-9_]([a-zA-Z0-9_\\.\\-]*[a-zA-Z0-9_])?$`
  - Keyboard shortcuts: Enter to save, Escape to cancel
  - Click outside to save, validates required label field
  - Optimistic updates with rollback on failure

  ## Technical details
  - Added `NameWidget.tsx` and `RenameForm.tsx` for inline renaming of table cells
  - Refactored `LinksManagerDialog.tsx` to `CellEditingDialog.tsx` for consistent naming and portal-based editing
  - Updated context menu and keyboard shortcuts to support "Rename"
  - Enhanced `useCellEditing` with improved state management
  - Updated `checkName.js` regex validator to allow alphanumeric, underscore, dot, and dash characters with proper start/end constraints
  - Enhanced keyboard navigation for inline editing

  The key changes are:
  - LinksManagerDialog was renamed to CellEditingDialog (not a new component)
  - Refactoring of the LinksManager components
  - Improved name validation regex

https://github.com/user-attachments/assets/11e696a3-5b22-44ee-806c-14c04e3ae7d2

### DIFF
--- a/shared/src/api/generated/index.ts
+++ b/shared/src/api/generated/index.ts
@@ -323,6 +323,7 @@ export { api as uRIsApi } from './uRIs'
 // users
 export * from './users'
 export { api as usersApi } from './users'
+export type { Permissions } from './users'
 
 // versions
 export * from './versions'
--- a/shared/src/containers/ProjectTreeTable/context/ProjectDataContext.tsx
+++ b/shared/src/containers/ProjectTreeTable/context/ProjectDataContext.tsx
@@ -55,7 +55,7 @@ export const ProjectDataProvider = ({ children, projectName }: ProjectDataProvid
     writableFields,
     isSuccess: isSuccessAttribs,
     isFetching: isFetchingAttribs,
-  } = useAttributeFields({ projectName, projectPermissions })
+  } = useAttributeFields({ projectPermissions })
 
   // GET USERS
   const { data: usersData = [] } = useGetUsersAssigneeQuery({ projectName }, { skip: !projectName })
--- a/shared/src/containers/ProjectTreeTable/hooks/useAttributesList.ts
+++ b/shared/src/containers/ProjectTreeTable/hooks/useAttributesList.ts
@@ -5,11 +5,10 @@ export interface ProjectTableAttribute extends AttributeModel {
 }
 
 interface UseAttributeFieldsParams {
-  projectName: string
   projectPermissions?: Permissions
 }
 
-const useAttributeFields = ({ projectName, projectPermissions }: UseAttributeFieldsParams) => {
+const useAttributeFields = ({ projectPermissions }: UseAttributeFieldsParams) => {
   const { data: info, isSuccess, isFetching } = useGetSiteInfoQuery({ full: true })
   const { attributes = [] } = info || {}
 


Innders (changes requested):
Looking really good, this is going to be so useful! Some bits of code to clean up and also some functionality fixes.

- [x] The auto generation toggle should be implied by the difference in the label and the name. It does not need to be a separate state.

- [x] Remove "Label is required". If the user tries to save with no label or name then it should just be treated as a cancel edit.

- [ ] When submitting/closing the dialog. The focus needs to return to the table. Try pressing "r" and then "escape". You can no longer now press "r" again to open it.

- [x] Font size of the name input should use `${theme.bodySmall}` to make it smaller.

Innders (changes requested):
Feeling really nice now, great stuff.

I have just made some tiny comments about cleaning some code up.

I also have some useability things I have just discovered... sorry!

- [ ] When typing the label input, I keep finding myself using the down arrow key to try to jump to the name input. Is is possible to add keyboard navigation using the up and down arrow keys to switch between the two inputs?

- [ ] When the editor is open, it is possible to select a different cell. When clicking outside of the dialog, the editing should cancel.

Innders (changes requested):
I'm still facing some issues unfortunately

- [ ] The label input doesn't autofocus or highlight anymore
- [ ] Clicking on another cell selects it but the selection then jumps back to the original one.
- [ ] You can't use cmd/ctrl+R to reload the page as it opens the dialog. The shortcut must only be "r" with no modifiers. 
- [ ] Turn off autocomplete, this could get very annoying.
Image

- [ ] When typing the name, the regex needs to be applied so that is impossible for the user to add something that is not compatible. The regex from the server is `^[a-zA-Z0-9_]([a-zA-Z0-9_\\.\\-]*[a-zA-Z0-9_])?$`

- Start with an alphanumeric character (A-Z, a-z, 0-9) or an underscore (_).
- Contain only alphanumeric characters, underscores, periods (.), or hyphens (-).
- End with an alphanumeric character or an underscore.
- Not be empty.

Image

- [ ] Pressing "Enter" key on the cell still opens the renaming when it should open the details panel.

Innders (commented):
The architure of the editor needs to be rethought and brought in line with other widgets. Look at LinksWidget as a good example.

1. `EditingEntityWidget` is not a widget and should be brought outside of widgets folder. It is a form component and should be put in shared/components and called `RenamingForm.tsx`.
2. A new component called `NameWidget` should be created to be used inside of the table. This component renders the CellEditingDialog and RenamingForm when editing.
3. Add this new widget to CellWidget and a new type `name`.
4. Use CellWidget in buildTreeTableColumns for the name.

shared/src/containers/ProjectTreeTable/widgets/RenameForm.tsx
```diff
@@ -0,0 +1,283 @@
...
+    [cellId, isSaving, handleSave],
+  )
+
+  // Add/remove click outside listener
+  useEffect(() => {
+    document.addEventListener('mousedown', handleClickOutside)
+    return () => {
+      document.removeEventListener('mousedown', handleClickOutside)
+    }
+  }, [handleClickOutside])
```
Innders, line 245: This is not required, because it is handled by CellEditingDialog...

shared/src/containers/ProjectTreeTable/widgets/RenameForm.tsx
```diff
@@ -0,0 +1,283 @@
...
+  rowId,
+  entityType,
+  initialName,
+  initialLabel = '',
+  onClose,
+  valueData,
+}) => {
+  const [label, setLabel] = useState(initialLabel)
+  const [name, setName] = useState(initialName)
+  const [isSaving, setIsSaving] = useState(false) // Add saving state
```
Innders, line 68: Clean up as setIsSaving is never set then isSaving is not required.

shared/src/containers/ProjectTreeTable/widgets/CellWidget.tsx
```diff
@@ -46,7 +47,15 @@ const Cell = styled.div`
...
 
-type WidgetAttributeData = { type: AttributeData['type'] | 'links' }
+type WidgetAttributeData = { type: AttributeData['type'] | 'links' | 'name' }
+export type NameData = {
+  name: string;
+  label: string;
+  meta?: any;
+  type?: string;
+  entityRowId?: string;
+  columnId?: string;
```
Innders, line 57: Do we need this extra other than just name and label? type, row and columnId are already passed as props

shared/src/containers/ProjectTreeTable/widgets/RenameForm.tsx
```diff
@@ -0,0 +1,283 @@
...
+`
+
+interface InlineEditingWidgetProps {
+  cellId: string
+  rowId: string
+  entityType: string
+  initialName?: string
+  initialLabel?: string
+  onClose?: () => void
+  valueData?: any // Contains meta, type, entityRowId, columnId
```
Innders, line 54: valueData type was defined in CellWidget, but it should be defined in this file, used here and then imported into CellWidget.

shared/src/containers/ProjectTreeTable/widgets/RenameForm.tsx
```diff
@@ -0,0 +1,283 @@
...
+          nameInputRef.current.focus()
+          // Position cursor at the end of the name input
+          const length = nameInputRef.current.value.length
+          nameInputRef.current.setSelectionRange(length, length)
+        }
+      } else if (e.key === 'Enter' && (e.ctrlKey || e.metaKey)) {
+        // Ctrl+Enter or Cmd+Enter to save
+        e.preventDefault()
+        handleSave()
+      } else if (e.key === 'Escape') {
```
Innders, line 178: Hitting only "Enter" without modifiers should still submit the form.

Innders (changes requested):
The reformating looks good and has solved a lot of the focusing issues. Now there needs to be some cleanup as commented on.

The previous notes also still need to be addressed.

- [x] You can't use cmd/ctrl+R to reload the page as it opens the dialog. The shortcut must only be "r" with no modifiers.
- [x] Turn off autocomplete, this could get very annoying.
- [x] When typing the name, the regex needs to be applied so that is impossible for the user to add something that is not compatible. The regex from the server is `^[a-zA-Z0-9_]([a-zA-Z0-9_\\.\\-]*[a-zA-Z0-9_])?$`

- Start with an alphanumeric character (A-Z, a-z, 0-9) or an underscore (_).
- Contain only alphanumeric characters, underscores, periods (.), or hyphens (-).
- End with an alphanumeric character or an underscore.
- Not be empty.

- [x] Pressing "Enter" key on the cell still opens the renaming when it should open the details panel.

Innders (changes requested):
Just a couple of tiny things and it will be finished!

Please address the comments. If the comment is unclear or there is an issue making the change, please leave a comment and do not resolve it.

- [x] Hitting "Enter" opens editing by default for all cells accept the name cell, where it opens the details panel instead.
- [x] Clean up some props for NameData
- [x] Escape keyboard shortcut is not needed as it is handled outside.
- [x] The name input is still letting me put in characters that are not allowed. When adding characters that are not allowed it should skip them or add an underscore for spaces. Update generateSystemName instead of checkName.

(this should NOT be allowed)

https://github.com/user-attachments/assets/7b74a12c-3aac-4364-89a0-86169ca49e43

filipvnencak:
I’ve moved RenameForm from` shared/src/containers/ProjectTreeTable/widgets/RenameForm.tsx` to `src/components/RenameForm.tsx` to make use of the checkName function. I’m also considering reusing this form in `src/components/NewEntity/NewEntity.tsx`, so please let me know if this placement makes sense.

Innders (changes requested):
Great thanks for making those changes. Last two small bugs to fix.

- [x] Placeholders are broken

Image

- [x] Hitting Enter whilst focused on name input should submit the changes.

Image

Innders (changes requested):
Awesome everything is working so nicely, super smooth and slick 🙌 

I just remembered one thing we forgot about... **permissions** 🫣

You can set if a user is allowed to update the `label` and `name`.

Image

I think it should be a fairly simple fix though...

If both permissions are disabled:
- Disable R shortcut
- Disable (not hide) the right click menu item
- They cannot open the dialog at all

If either name or label is disabled the user can still open the dialog except the disabled type will have it's input disabled.

Disabled input should be greyed out and in italics.

If the `name` is disabled you will also need to prevent the auto generation from happening.

### Notes

We can use useGetMyProjectPermissions query for this. This is already used in https://github.com/ynput/ayon-frontend/blob/1c6c0f95dd7b4f4f500c13cedfb50ef33abf60cf/shared/src/containers/ProjectTreeTable/hooks/useAttributesList.ts for the attributes.

You will probably need to move useGetMyProjectPermissionsQuery out of this file and into ProjectDataContext so that it can be used for attributes and the naming permissions.

Let me know if you want to go through it.

Innders (changes requested):
The permissions work correctly now thanks. Two things...

- [ ] Address the comment made on ProjectDataContext about duplicate use of useGetMyProjectPermissionsQuery in useAttributesList.
- [x] Bug: When there are multiple folders selected it will try to update all the selected folders. Remove `{ selection: meta?.selection }` from `updateEntities`

PR #1489, opened 2025-09-16
Overview: Better validate cell pasting

filipvnencak:
## Description of changes 
 Improved clipboard validation for the ProjectTreeTable with enhanced type checking and error handling. Key improvements include:

  - Added validation for numeric attribute fields (integer/float types)
  - Enhanced enum validation with case-insensitive matching and auto-correction
  - Added proper assignee validation with multiple assignee support
  - Fixed attribute field validation using schema definitions
  - Improved error messages for better user feedback
  - Added automatic value correction for case mismatches

  The changes ensure pasted data is properly validated against column schemas before applying updates, preventing invalid data entry and improving user experience.

## Additional context

https://github.com/user-attachments/assets/6158cca5-263f-4257-82e8-dda1faca0059

### DIFF
--- a/shared/src/containers/ProjectTreeTable/context/ClipboardContext.tsx
+++ b/shared/src/containers/ProjectTreeTable/context/ClipboardContext.tsx
@@ -49,8 +49,8 @@ export const ClipboardProvider: React.FC<ClipboardProviderProps> = ({
       try {
         // Get visible columns in display order, excluding row selection
         const visibleColumnIds = visibleColumns
-          .map(col => col.id)
-          .filter(id => id !== ROW_SELECTION_COLUMN_ID)
+          .map((col) => col.id)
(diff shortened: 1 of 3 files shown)


Innders (changes requested):
Works well just a couple of edge cases to fix

- [ ] Tags can receive anything, so there are no restrictions for it.
- [x] The case insensitivity is not working properly. If I paste "Normal" it gets accepted but the value is invalid. The value needs to be transformed to the real valid value `normal`. Also "NORMAL" didn't work at all.

Image

filipvnencak:
@Innders You mentioned `Tags can receive anything, so there are no restrictions for it. `Just to clarify, does this mean tags can accept any value? In the anatomy settings I noticed there are defined tags, and when I paste something with validation turned off, the tags appear differently. Could you explain how this is intended to work?

Innders (approved):
Works great thanks, ready for merge!

Issue #1356, opened 2025-07-30
Overview: Creating new entity should set label with autogen name


Innders:
Now that we can change the name after creating a folder/task, we can set the label when creating the entity and then auto generate the name.

- Switch name input to label input.
- Label input can be anything with any characters and format.
- Name should be strict and auto generated from the label.
- Clicking on name should allow to set it to something different.
- Add Type, Label and Name labels to fields for clearer inputs.

<img width="939" height="506" alt="Image" src="https://github.com/user-attachments/assets/7d68ebde-7ac4-4161-a5d8-8657892242ec" />

PR #1494, opened 2025-09-18
Overview: Creating new entity should set label with autogen name #1356

filipvnencak:
## Description of changes 
 Enhanced NewEntity component with smart auto-generation of entity names and improved form reusability across single entity creation and sequence creation workflows.

  Key Features:

- Smart name auto-generation: Names are automatically generated from labels, but stop auto-updating once the user manually edits the name field
- Unified form component: Created reusable NewEntityForm component that works in both single entity creation and folder sequence creation
- Label/Name separation: Clear distinction between display labels and backend names with proper sanitization
- Editable name field: Users can click on the generated name to edit it manually
- Consistent lowercase naming: All generated names are lowercase with proper character sanitization

## Technical details

- NewEntity.tsx: Added nameManuallyEdited state tracking and smart auto-generation logic
- NewEntityForm.tsx: Created flexible, reusable form component with configurable field mappings
- FolderSequence.jsx: Integrated NewEntityForm for consistent UX across creation workflows
- getSequence.ts: Updated to return objects with both name and label properties
- checkName utility: Enhanced to convert to lowercase and handle spaces properly

## Additional context

https://github.com/user-attachments/assets/0065d35b-96a5-4eda-a146-1fe755492565

### DIFF
--- a/src/components/NewEntity/NewEntity.tsx
+++ b/src/components/NewEntity/NewEntity.tsx
@@ -1,7 +1,6 @@
 import React, { useState, useRef, KeyboardEvent } from 'react'
 import { capitalize, isEmpty } from 'lodash'
 import {
-  InputText,
   SaveButton,
   Spacer,
   Toolbar,
@@ -12,7 +11,6 @@ import {
   InputSwitch,
 } from '@ynput/ayon-react-components'
 import styled from 'styled-components'
-import TypeEditor from './TypeEditor'
 import checkName from '@helpers/checkName'
 import ShortcutWidget from '@components/ShortcutWidget'
 import {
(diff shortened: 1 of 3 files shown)


Innders (changes requested):
The functionality works but some styling changes are needed.

Can the name please follow the design in the issue?

- [x] inputs labels should use theme.labelMedium
- [x] name label and input should be on the same line 
- [x] name input value should use white color (default)
- [x] Reduce the gap between the body and the footer by 16px

Image

- [x] Hover effect on the name input value with radius.

```css
  padding: 1px 6px;
  border-radius: 4px;
  &:hover {
    background-color: var(--md-sys-color-surface-container-low);
  }
```

- [ ] When editing the name the input should be styled the same as when hovering over it. This means the text should not jump at all and it should be completely seamless transition.
- [x] Outline style when focusing name input.
- [x] Select all text when focusing name input.

Image

Innders (changes requested):
Getting there but I think we could refine it more.

- [ ] name writable state starts as true which must be a mistake?
- [ ] Increase name padding: 2px 6px;
- [ ] When switching to the input the height is smaller, causing a small jump. Ensure the name input is the same height so there is no jumping.
- [ ] When not focused the width should fit content (remove min-width). When focused it can have the same min width.
- [ ] When hovering add a little edit icon at the end (edit). When focusing the icon is not seen.

Image

Here are the name display style changes required

```css
// min width removed
// font style italic removed
  padding: 2px 6px; // <- changed
  display: inline-flex;
  gap: var(--base-gap-small);
  cursor: pointer;

  .icon {
    display: none;
    font-size: 14px;
  }
  &:hover {
    background-color: var(--md-sys-color-surface-container-low);
    .icon {
      display: flex;
    }
  }
```

Innders (approved):

PR #1506, opened 2025-09-22
Default Project Tab - YN-0059

filipvnencak:
## Description of changes 

- Created useProjectDefaultTab hook that saves/retrieves the last visited tab in localStorage
- Tracks user's tab preferences automatically when they navigate between project tabs
- Remembers the last tab (overview, browser, tasks, etc.) and uses it as default for new project navigation
- Fallback to 'overview' if no preference is stored

## Technical details

Converted window.location.pathname.split() to React Router hooks in projectMenu.jsx:
- Added useLocation hook import
- Replaced window.location.pathname with location.pathname

## Additional context

https://github.com/user-attachments/assets/2106b3ca-bd55-4920-8dbb-17c17415e83d

### DIFF
--- a/src/hooks/useProjectDefaultTab.ts
+++ b/src/hooks/useProjectDefaultTab.ts
@@ -1,4 +1,4 @@
-const STORAGE_KEY = 'ayon-project-default-tab'
+const STORAGE_KEY = 'project-default-tab'
 
 /**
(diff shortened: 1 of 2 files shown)


filipvnencak:
> For standard pages this works perfect, but it does not handle i-frame based addon pages such as node graph or planner

Fixed

Issue #1504, opened 2025-09-19
Details Panel: Built-in attributes should be read only for versions


Innders:
<img width="1371" height="957" alt="Image" src="https://github.com/user-attachments/assets/b417b7b0-ad84-4244-acfe-73b291e5e29c" />

All built-in attributes for versions should be read only.

The only editable fields are:
- status
- tags
- custom attributes with version scope

PR #1508, opened 2025-09-22
Details Panel: Attributes should be read only for versions

redmonkez12:
## Description of changes 
* This pull request disabled attributes for editing in version; only tags and status should be editable

## Notes
I tried to update the Folder Type and got this error. Do we want to disable this too?

Screenshot 2025-09-22 at 13 23 48


Innders (changes requested):
- [ ] Attach issue to the PR.
- [ ] Fix the folderType issue by disabling input when a folder `hasVersions`
- [ ] See updated issue comments regarding built-in vs custom version attributes.
https://github.com/ynput/ayon-frontend/issues/1504#issue-3434714639

> All built-in attributes for versions should be read only.
> The only editable fields are: status, tags,  custom attributes with version scope

PR #1511, opened 2025-09-22
Overview: Move group by context menu to column menu 

redmonkez12:
## Description of changes 
This pull request removes group by from the context menu and adds it to the column menu

https://github.com/user-attachments/assets/eef2e592-f611-490f-9567-36e960600f46


Innders (changes requested):
If I click on the column resize the pages crashes and I get this error?

`Uncaught Error: Rendered fewer hooks than expected. This may be caused by an accidental early return statement.`

redmonkez12:
> If I click on the column resize the pages crashes and I get this error?
> 
> `Uncaught Error: Rendered fewer hooks than expected. This may be caused by an accidental early return statement.`

Wrong hook position, fixed

PR #1520, opened 2025-09-24
Project Access: Very slow updating

filipvnencak:
## Description of changes 
 Changed the "Add" button visibility in AccessGroupsCell from React state-based hover to pure CSS hover for improved responsiveness.

This fixes the issue where rapid mouse movement would cause the Add button to appear delayed or on the wrong row due to React render cycles.

## Technical details

- Removed `showAddButton` prop from AccessGroupsCell component
- Replaced React-based hovering state with CSS `:hover` selector on table rows
- Button now appears instantly on row hover without the "one row behind" lag during quick mouse movements
- Maintained all existing functionality while improving hover performance

## Additional context

https://github.com/user-attachments/assets/9bfa2c5e-8012-404c-ad3e-703b7ed68809

### DIFF
--- a/src/pages/ProjectManagerPage/Users/AccessGroupsCell.tsx
+++ b/src/pages/ProjectManagerPage/Users/AccessGroupsCell.tsx
@@ -7,7 +7,6 @@ import { Spacer } from '@ynput/ayon-react-components'
 type Props = {
   data: $Any
   selected: boolean
-  showAddButton: boolean
   hovering: boolean
   addButtonDisabled: boolean
   readOnly: boolean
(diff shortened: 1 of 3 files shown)


Innders (changes requested):
Even with these changes I'm still feeling a slightly lag. Even the CSS hovering of the row is lagging behind the mouse when moving fast.

Looking deeper it seems that onHoverRow is triggering a state update of hoveredUser that is causing the full table to re-render constantly.

Can we find a way to either not user a hovered state or maybe use a ref that is more performant. Maybe add some throttling as well so the updating isn't so crazy.

Please look deeper into the performance issues.

- [ ] Do the same css trick for `UserCell.tsx`.
- [ ] Improve performance using useRef and throttling

PR #1527, opened 2025-09-26
Overview: Make thumbnails bigger by increasing the column width

filipvnencak:
## Description of changes 
Added dynamic thumbnail resizing functionality to the ProjectTreeTable. When users resize the thumbnail column width, thumbnail images now scale to fit the new width while
  maintaining their 16:9 aspect ratio, and table rows automatically adjust their height to accommodate the scaled thumbnails.

## Technical details

- **ThumbnailWidget**: Updated to use `width: calc(100% - 8px)` with `aspect-ratio: 1.77` for responsive scaling
- **useDynamicRowHeight hook**: New hook that calculates row height based on thumbnail column width using the formula: `(columnWidth - padding) / aspectRatio + padding`
- **Row virtualizer**: Enhanced to use dynamic `estimateSize` function for variable row heights
- **Table components**: Updated to pass calculated row heights to all table cells and drag handles

  **Limitations:**

- All rows resize uniformly (not just rows with actual thumbnails) to maintain consistent table layout
- Uses TanStack Virtual's row measurement for optimal scrolling performance
- Minimum row height enforced to prevent layout issues with very narrow columns

## Additional context

https://github.com/user-attachments/assets/9f9e9ad6-279f-4b7d-96ba-ee3fde49af58

### DIFF
--- a/shared/src/components/ProjectTableSettings/ProjectTableSettings.tsx
+++ b/shared/src/components/ProjectTableSettings/ProjectTableSettings.tsx
@@ -10,6 +10,7 @@ import styled from 'styled-components'
 import { SettingHighlightedId, useSettingsPanel } from '@shared/context'
 import { SettingsPanel, SettingConfig } from '@shared/components/SettingsPanel'
 import ColumnsSettings from './ColumnsSettings'
+import RowHeightSettings from './RowHeightSettings'
 
 const StyledCustomizeButton = styled(Button)`
   min-width: 120px;
@@ -48,7 +49,7 @@ export const ProjectTableSettings: FC<ProjectTableSettingsProps> = ({
   highlighted,
 }) => {
   const { attribFields, projectInfo } = useProjectTableContext()
-  const { columnVisibility } = useColumnSettingsContext()
+  const { columnVisibility, rowHeight = 40 } = useColumnSettingsContext()
 
   const columns = [
     {
@@ -113,6 +114,13 @@ export const ProjectTableSettings: FC<ProjectTableSettingsProps> = ({
       icon: 'view_column',
       preview: `${visibleCount}/${columns.length}`,
       component: <ColumnsSettings columns={columns} highlighted={highlighted} />,
+    },
+    {
+      id: 'row-height',
+      title: 'Row Height',
+      icon: 'height',
+      preview: `${rowHeight}px`,
+      component: <RowHeightSettings />,
     },
   ]
 
--- a/shared/src/components/ProjectTableSettings/ProjectTableSettings.tsx
+++ b/shared/src/components/ProjectTableSettings/ProjectTableSettings.tsx
@@ -115,13 +115,6 @@ export const ProjectTableSettings: FC<ProjectTableSettingsProps> = ({
       preview: `${visibleCount}/${columns.length}`,
       component: <ColumnsSettings columns={columns} highlighted={highlighted} />,
     },
-    {
-      id: 'row-height',
-      title: 'Row Height',
-      icon: 'height',
-      preview: `${rowHeight}px`,
-      component: <RowHeightSettings />,
-    },
   ]
 
   settings.forEach((setting) => defaultSettings.push(setting))
--- a/shared/src/components/ProjectTableSettings/ProjectTableSettings.tsx
+++ b/shared/src/components/ProjectTableSettings/ProjectTableSettings.tsx
@@ -10,7 +10,6 @@ import styled from 'styled-components'
 import { SettingHighlightedId, useSettingsPanel } from '@shared/context'
 import { SettingsPanel, SettingConfig } from '@shared/components/SettingsPanel'
 import ColumnsSettings from './ColumnsSettings'
-import RowHeightSettings from './RowHeightSettings'
 
 const StyledCustomizeButton = styled(Button)`
   min-width: 120px;
@@ -49,7 +48,7 @@ export const ProjectTableSettings: FC<ProjectTableSettingsProps> = ({
   highlighted,
 }) => {
   const { attribFields, projectInfo } = useProjectTableContext()
-  const { columnVisibility, rowHeight = 40 } = useColumnSettingsContext()
+  const { columnVisibility } = useColumnSettingsContext()
 
   const columns = [
     {
--- a/shared/src/containers/ProjectTreeTable/ProjectTreeTable.tsx
+++ b/shared/src/containers/ProjectTreeTable/ProjectTreeTable.tsx
@@ -440,7 +440,7 @@ export const ProjectTreeTable = ({
   const columnSizeVars = useCustomColumnWidthVars(table, columnSizing)
 
   // Calculate dynamic row height based on user setting from Customize panel
-  const { getRowHeight, thumbnailRowHeight, defaultRowHeight } = useDynamicRowHeight()
+  const { getRowHeight, defaultRowHeight } = useDynamicRowHeight()
 
   const attribByField = useMemo(() => {
     return attribFields.reduce((acc: Record<string, AttributeEnumItem[]>, attrib) => {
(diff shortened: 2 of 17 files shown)


Innders (changes requested):
Cool this working really nicely! A couple of points

## General
- [ ] Share some screenshot images and video of the feature.
- [x] Make the default height 34px (I started on the min for some reason)
- [ ] The height is not saved to the view properly so refreshing will reset it.

## Thumbnail

- [x] There's a big gap on the right of the cell preventing the thumbnail from fully filling the horizontal space.

Image

- [ ] The thumbnails should be centered vertically when the height does not fill the space.

Image

- [ ] Some aspect ratios do not properly fill the space.

Image

## Label/Path

- [ ] The last `/` gap size is uneven and too big. It should be 4px either side.

Image

- [ ] Fix double ellipsing.

Image

- [ ] Dropdown button cut off on small row heights

Image

## Settings

- [ ] Move the setting so that is on the root level of the settings.

Image

- [ ] Remove the `px` suffix to the number value.
- [ ] Label should be `Row height`. First word is only capitalized, always for labels.

Innders (changes requested):
- [x] Min width of thumbnail column to 24px.
- [x] Sometimes when increasing the size of the row height the column width will also increase a little bit afterwards?

https://github.com/user-attachments/assets/ed8887ac-943c-44c3-8d1d-fa662757286a

- [x] The slider is pretty glitchy and jumps around.

https://github.com/user-attachments/assets/7ed12e4b-68a3-4b32-a533-c7ef8fdb1e80

- [x] I like the idea that cell content can now overflow onto two lines but this breaks when the row height is too small. We need to set the max number of lines the content can overflow to based on the row height.

Image

- [x] Tags overflowing ellipse broken.

Image

- [x] The dropdown height button should fit content and have min-height 32px.

Image

- [x] Remove this border radius that sometimes shows.

Image

- [x] The padding around the thumbnail doesn't seem even. Should be 6px all the way around.

Image

shared/src/containers/ProjectTreeTable/widgets/EntityNameWidget.tsx
```diff
@@ -56,21 +56,50 @@ const StyledContent = styled.div`
...
+  }
+
+  .divider {
+    ${theme.bodyMedium}
+    font-size: 14px;
+    color: var(--md-sys-color-outline);
+    margin: 0 4px;
+    flex-shrink: 0;
+  }
   }
```
Innders, line 103: invalid bracket here

shared/src/containers/ProjectTreeTable/widgets/EntityNameWidget.tsx
```diff
@@ -56,21 +56,50 @@ const StyledContent = styled.div`
...
   overflow: hidden;
+  flex: 1;
+  min-width: 0;
 
   .path {
-    ${theme.labelSmall}
-    margin-bottom: -4px;
+    ${theme.bodyMedium}
+    font-size: 14px;
+    margin-bottom: ${props => props.$isCompact ? '0' : '-4px'};
```
Innders, line 71: Instead of using styled props could we please classNames. For example here we would have a .compact className that overrides the styles for compact. We can use clsx function to add these variable classNames in the component.

```css
&.compact {
 flex-direction: row;
 align-items: center
 ...rest
}
```

shared/src/containers/ProjectTreeTable/widgets/EntityNameWidget.tsx
```diff
@@ -121,12 +160,13 @@ export const EntityNameWidget = ({
           <div style={{ display: 'inline-block', minWidth: 24 }} />
         )
       ) : null}
-      <StyledContentWrapper style={{ height: path ? 32 : 24 }}>
+      <StyledContentWrapper style={{ height: contentHeight }}>
         <StyledContentAbsolute>
           <StyledContent>
             {icon && <Icon icon={icon} />}
-            <StyledTextContent>
+            <StyledTextContent $isCompact={isCompact}>
```
Innders, line 167:
```suggestion
            <StyledTextContent className={clsx({ compact: isCompact })}>
```

shared/src/components/ProjectTableSettings/ProjectTableSettings.tsx
```diff
@@ -10,6 +10,7 @@ import styled from 'styled-components'
 import { SettingHighlightedId, useSettingsPanel } from '@shared/context'
 import { SettingsPanel, SettingConfig } from '@shared/components/SettingsPanel'
 import ColumnsSettings from './ColumnsSettings'
+import RowHeightSettings from './RowHeightSettings'
```
Innders, line 13: Unused import

shared/src/components/ProjectTableSettings/ProjectTableSettings.tsx
```diff
@@ -48,7 +49,7 @@ export const ProjectTableSettings: FC<ProjectTableSettingsProps> = ({
   highlighted,
 }) => {
   const { attribFields, projectInfo } = useProjectTableContext()
-  const { columnVisibility } = useColumnSettingsContext()
+  const { columnVisibility, rowHeight = 40 } = useColumnSettingsContext()
```
Innders, line 52: unused import

shared/src/containers/ProjectTreeTable/widgets/EnumCellValue.tsx
```diff
@@ -32,34 +34,34 @@ const StyledWidget = styled.div`
...
-const StyledValuesContainer = styled.div`
+const StyledValuesContainer = styled.div<{ $allowWrap?: boolean }>`
   flex: 1;
   display: flex;
   gap: var(--base-gap-small);
   align-items: center;
   overflow: hidden;
   border-radius: var(--border-radius-m);
   padding: 0px 2px;
+  flex-wrap: ${({ $allowWrap }) => ($allowWrap ? 'wrap' : 'nowrap')};
```
Innders, line 45: same thing with using className instead of prop :)

shared/src/containers/ProjectTreeTable/widgets/ThumbnailWidget.tsx
```diff
@@ -6,17 +6,17 @@ import styled from 'styled-components'
 const Wrapper = styled.div`
   position: absolute;
   inset: 0;
-  padding: 4px;
+  padding: 6px;
```
Innders, line 9:
```suggestion
  padding: 4px;
```

shared/src/containers/ProjectTreeTable/ProjectTreeTable.tsx
```diff
@@ -438,6 +439,9 @@ export const ProjectTreeTable = ({
 
   const columnSizeVars = useCustomColumnWidthVars(table, columnSizing)
 
+  // Calculate dynamic row height based on user setting from Customize panel
+  const { getRowHeight, thumbnailRowHeight, defaultRowHeight } = useDynamicRowHeight()
```
Innders, line 443: unused variable

Innders (changes requested):
Looking really good, just need to do some clean up in the files and some small things.

- [ ] If I refresh the page the row height is not saved and it goes back to default.
- [ ] When in hierarchy mode there is a jumping of gaps when the row gets small. We should try to keep the gap the same here. The gap should be the small size (when in the larger row height)

https://github.com/user-attachments/assets/7b5a5e44-5079-4a32-9b8b-722e02ac27d5

- [ ] I can use the arrow keys to change the range (which is great) but it does not update the row height.

https://github.com/user-attachments/assets/c9d0826b-0753-4963-a63b-aa472f386d63

I made some style changes to the slider FYI :)

Image

Innders (approved):
Great stuff, approved! (please do not merge until after 1.12.1 release)

PR #1535, opened 2025-09-30
Task Progress: Folder status icon does not show

filipvnencak:
## Description of changes 

1. Include both 'task' and 'folder' scopes when fetching statuses for the Task Progress page
2. Changed the filtering logic from .every() to .some() so statuses with ANY of the specified entity types are included, not requiring ALL of them

## Technical details

Update filtering logic to use `some` instead of `every` and include 'folder' in task statuses.


Innders (changes requested):
The original reason for the scoping was to ensure that the status select for tasks only showed statuses scoped to tasks. These changes break that.

Image

Innders (approved):
Works now thanks.

PR #1537, opened 2025-09-30
Task Progress: Expanded selected folders glitch

filipvnencak:
## Description of changes 
  When a folder row is both selected and expanded, the Path component (which displays the selected background color) will now be hidden, preventing it from overflowing
  behind the thumbnail image.

## Technical details

Fixed the background overflow issue. Here's what I changed:

1. FolderBody.tsx:80 - Added expanded class to the Path component
2. FolderBody.styled.ts:103-107 - Added CSS rule to hide the Path component when expanded

## Additional context
Before:
Screenshot 2025-09-30 at 14 48 05
After fix:
Screenshot 2025-09-30 at 14 47 27


Innders (approved):
Thanks!

PR #1541, opened 2025-10-01
Tooltips: Breaking words in half

filipvnencak:
## Description of changes 

Replace `word-break: break-all` with `overflow-wrap: break-word` for better text wrapping behavior


Innders (approved):
LGTM just waiting for @redmonkez12 to review the code.

PR #1551, opened 2025-10-02
Increased colours readability

martastain:
This pull request enhances the color handling for enum cell values in the `EnumCellValue` component. The main improvement is the introduction of a utility to ensure text color remains readable against various background colors by dynamically determining whether to use black or white text.

### Before

image

### After

image


BigRoy:
> Missing in details header
> 
> Image

For what it's worth - in this case I find the white on pink more readable than the black on pink.
I wonder if the threshold is too low - or that maybe this just an odd one out?

Innders:
>  I find the white on pink more readable than the black on pink

I agree, I think the thresholds need to be way higher for turning to black. And maybe the black should be more of an off black so it's not as harsh.

shared/src/containers/DetailsPanel/DetailsPanelFloating/DetailsPanelFloating.tsx
```diff
@@ -126,8 +126,8 @@ export const DetailsPanelFloating: FC<DetailsPanelFloatingProps> = () => {
           </Styled.Content>
         </Styled.Header>
         <Styled.Row>
-          <Styled.Status style={{ backgroundColor: statusAnatomy?.color }}>
-            {<Icon icon={statusAnatomy.icon || 'question_mark'} />}
+          <Styled.Status style={{ backgroundColor: statusAnatomy?.color, color: getTextColor(statusAnatomy?.color || '#000') }}>
```
Innders: Why is there a fallback of black here. The argument can be undefined as the function handles the fallback colour.

shared/src/containers/Feed/components/ActivityCategorySelect/CategoryTag.tsx
```diff
@@ -81,7 +82,7 @@ export const CategoryTag = forwardRef<HTMLDivElement, CategoryTagProps>(
           power: isPower,
           disabled: isDisabled,
         })}
-        style={{ ...style, backgroundColor: color, borderColor: color }}
+        style={{ ...style, backgroundColor: color, borderColor: color, color: getTextColor(color|| "#2E3133") }}
```
Innders: Why is there a random fallback colour here when the getTextColor handles undefined color arguments.

Innders (changes requested):
Status dropdown needs to be handled properly inside of https://github.com/ynput/ayon-react-components Create a new branch off of main for it. Copy the getTextColor function over to be used there. You will be looking at the `invert` prop on the Dropdown component. Use `yarn storybook` to start the development server.

- [x] Update Dropdown invert styling in ayon-react-components
- [x] Fix Dropdown item color contrast.
- [x] Fix category submit button text not changing

Image

shared/src/hooks/useTagStyling.ts
```diff
@@ -34,7 +35,7 @@ export const useTagStyling = ({
               if (tagColor) {
                 const htmlElement = tagElement as HTMLElement
                 htmlElement.style.backgroundColor = tagColor
-                htmlElement.style.color = 'var(--color-tag-text)'
+                htmlElement.style.color = getTextColor(tagColor)
```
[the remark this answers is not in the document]
filipvnencak: @Innders, ss it possible to change the text color in this hook? I checked and TagsSelect comes from the @ynput/ayon-react-components package.

Innders (changes requested):
- [ ] Update ayon-react-components library
- [ ] Now use `getTextColor` function from ARC and remove util file from ayon-frontend
- [ ] Category dropdown item and submit button not using correct color [previous issue] 

image

Innders (changes requested):
- [x] Tags are not correct in the details panel
- [x] Dropdown selected item colour not correct (3rd time asking)

Image

Image

PR #1552, opened 2025-10-02
Lists: Filter by tags

filipvnencak:
## Description of changes 
Adds dynamic tag-based filtering capability to the Lists Filters Dialog. Users can now filter lists by tags, with the filter options dynamically generated based on tags
  present in existing lists.

## Technical details

- Dynamically generates tag filter options from all tags found across lists in listsData
- Integrates with project tag anatomy to display tag colors from projectInfo
- Shows tag counts (number of lists with each tag) and sorts by count (most used first)
- Tag filter only appears when at least one list has tags
- Uses OR operator for tag filtering (shows lists with any of the selected tags)
- Supports custom tag values in the filter

## Additional context

Screenshot 2025-10-02 at 12 36 14

Screenshot 2025-10-02 at 12 36 37

### DIFF
--- a/src/pages/ProjectListsPage/components/ListsFiltersDialog/ListsFiltersDialog.tsx
+++ b/src/pages/ProjectListsPage/components/ListsFiltersDialog/ListsFiltersDialog.tsx
@@ -58,21 +58,11 @@ const ListsFiltersDialog: FC<ListsFiltersDialogProps> = ({}) => {
       },
     ]
 
-    // Add tags option if there are any tags in the lists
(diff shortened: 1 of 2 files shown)


Innders (changes requested):
Works great, just a couple of small things.

- [x] The tags options should be based on project anatomy. It does not matter what tags are on current lists, just show all of the tags even if no list has it. You can still sort by the count though.

- [ ] When filtering could you please make the menu item active so it gets a check mark. This is so that the user knows there is some filtering going on.

Image

PR #1556, opened 2025-10-03
Lists Attributes

filipvnencak:
… enhancing attribute options and value handling

## Description of changes 

This PR adds support for list-scoped attributes throughout the application, enabling attributes to be specifically assigned to and displayed within entity lists. This brings
  list entities to feature parity with other entity types (folders, tasks, versions) in terms of attribute handling.

  Key changes:

- Added 'list' as a new scope option in the Attribute Editor
- List attribute forms now fetch and display list-scoped attributes from the global attribute definitions
- List filter dialog now includes list-scoped attributes as filterable options
- Attributes with enum values are properly supported in list filters

## Technical details

- Fetches all attributes via useGetAttributeListQuery()
- Filters by scope?.includes('list') (matching the pattern in useEntityFields.ts:128 for folders/tasks)
- Merges list-scoped attributes with passed attributes in form fields
- Stores scoped attribute values in the attrib object

## Additional context

The approach maintains consistency with how attributes are handled in folder/task detail panels (useEntityFields.ts, NewEntityContext.tsx), ensuring uniform
  behavior across all entity types.

Screenshot 2025-10-03 at 10 32 47

### DIFF
--- a/shared/src/components/AttributeEditor/AttributeEditor.tsx
+++ b/shared/src/components/AttributeEditor/AttributeEditor.tsx
@@ -24,6 +24,7 @@ const SCOPE_OPTIONS = [
   { value: 'version', label: 'Version' },
   { value: 'representation', label: 'Representation' },
   { value: 'user', label: 'User' },
+  { value: 'list', label: 'List' },
 ]
 
 // Define types for constants
--- a/shared/src/components/ListAttributeForm/ListAttributeForm.tsx
+++ b/shared/src/components/ListAttributeForm/ListAttributeForm.tsx
@@ -5,6 +5,7 @@ import {
   EntityListModel,
   useGetProjectQuery,
   useUpdateEntityListMutation,
+  useGetAttributeListQuery,
 } from '@shared/api'
 import { toast } from 'react-toastify'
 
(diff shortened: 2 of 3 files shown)


Innders (changes requested):
This is a pretty awesome update!

- [x] Inherit flag will never make sense for lists, hide checkbox for lists.

Image

- [x] Use correct attribute icons for the filtering. [getAttributeIcon.ts](https://github.com/ynput/ayon-frontend/blob/fa6ee856e071214852cecc4d1a869c11b81730f4/shared/src/util/getAttributeIcon.ts) [useBuildFilterOptions](https://github.com/ynput/ayon-frontend/blob/fa6ee856e071214852cecc4d1a869c11b81730f4/shared/src/components/SearchFilter/useBuildFilterOptions.tsx) I think you can probably reuse this function?

Image

- [x] Filtering by number or text should allow for custom values to added. I think this might be solved by useBuildFilterOptions maybe.

Image

- [ ] Number fields should show the aggregate of the numbers set on the lists. If list 1 has 100 and list 2 has 200 then we should see 100 and 200 as options for the attribute.

Innders (changes requested):
- [ ] Boolean attributes should always have Yes/No options like in the overview page.

Image

- [ ] Filters are still stored on the [user object](https://github.com/ynput/ayon-frontend/blob/afa4ff62d77773bcfb486d565c24761f80d07dd1/src/pages/ProjectListsPage/context/ListsDataContext.tsx#L67) but need to be stored on the view. Look at overview logic for how  this is done as it requires some converting from queryFilters to filters that work with the component. Try to use the same logic/component.

Innders (approved):
Amazing, thanks!

PR #1572, opened 2025-10-08
Links: Show real time updates for link changes in the table

filipvnencak:
## Description of changes 

Added real-time WebSocket subscriptions to link queries so that when one user creates or deletes a link, other users viewing the same entities will see the updates
  immediately without needing to refresh the page.

## Technical details

- **Main Query (`getEntityLinks`)**: Added `onCacheEntryAdded` lifecycle hook that:
- Subscribes to `link.created` and `link.deleted` WebSocket events
-  Uses batched updates (500ms debounce, max 100 entities) to efficiently handle multiple link changes
- Extracts both `inputId` and `outputId` from event messages since both entities in a link relationship need cache updates
- Fetches fresh link data for affected entities and updates RTK Query cache in real-time
- Filters events by project name to prevent cross-project updates

## Additional context

Follows the same implementation pattern as existing real-time features (e.g., `getOverview.ts` task updates):
- Uses `PubSub.subscribe` with topic prefix matching
- Implements debounced batch processing
- Uses `updateCachedData` for in-place cache modifications
- Properly handles subscription cleanup on cache entry removal

### DIFF
--- a/shared/src/api/queries/links/getLinks.ts
+++ b/shared/src/api/queries/links/getLinks.ts
@@ -70,6 +70,7 @@ const injectedQueries = gqlLinksApi.injectEndpoints({
           return { cursor: pageInfo.endCursor }
         },
       },
+      providesTags: [{ type: 'linkSearchItem', id: 'LIST' }],
       queryFn: async ({ queryArg, pageParam }, api) => {
         try {
           const { projectName, entityType, search, parentIds } = queryArg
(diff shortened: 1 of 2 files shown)


shared/src/api/queries/links/getLinks.ts
```diff
@@ -239,6 +240,47 @@ const injectedQueries = gqlLinksApi.injectEndpoints({
...
+            const outputId = message?.summary?.outputId
+            if (!inputId && !outputId) return
+
+            // Invalidate the cache when a link is created or deleted
+            // This is simpler than updating the infinite query cache structure
+            const cacheEntry = getCacheEntry()
+            if (cacheEntry.status === 'fulfilled') {
+              // Force a refetch by invalidating the cache
+              // The user will see updated results on next page load
+              // Note: Currently just tracking the event, actual invalidation could be implemented if needed
```
Innders, line 269: I'm guessing the invalidation still needs to be added here?

Innders (changes requested):
The realtime updates seem to work nicely! Just need to invalidate the search query/s now when we know a change has been made to ensure we don't hit any stale caches.

PR #1575, opened 2025-10-09
Overview: Field updates are not reflected in filtering

filipvnencak:
## Description of changes 
  Implemented filter-aware cache updates for Overview tasks and Lists to ensure entities are correctly shown/hidden when filtered fields are updated.

  Problem: When filtering tasks/list items and updating a field that affects the active filter (e.g., changing status from "Pending" to "Approved" while filtering for
  "Pending"), the optimistic cache update would keep the item visible when it should disappear. This happened because we can't reliably determine client-side if an item still
   matches the filter after an update.

  Solution: After optimistic updates, the code now evaluates whether the updated entity still matches the active filter. If it doesn't match, it's removed from the cache
  immediately. If an item now matches a filter it previously didn't, it's added to the filtered cache.

  Key Features:

- Tasks/items disappear instantly when they no longer match the filter
- Tasks/items appear when they now match the filter
- No visible table refresh or loading UI
- Works for all fields (status, assignees, tags, custom attributes)
- Maintains snappy, responsive UX

## Technical details
 New Utility: shared/src/containers/ProjectTreeTable/utils/filterEvaluator.ts

- `doesEntityMatchFilter()`: Evaluates if an entity matches a QueryFilter
- Supports all filter operators: eq, ne, in, notin, includes, excludes, includesall, excludesall, includesany, excludesany, gt, gte, lt, lte, like, isnull, notnull
- Handles nested fields (e.g., attrib.fieldName)
- Handles array fields (assignees, tags)

Modified Files:
  1. shared/src/api/queries/overview/updateOverview.ts
- Updates all cache entries (including those with different filters) to ensure consistency
- Evaluates updated tasks against active filters
- Removes tasks that no longer match or adds tasks that now match
- Finds full task data from no-filter cache when adding to filtered views

2. shared/src/api/queries/entityLists/updateLists.ts
- Similar implementation for task items
- Handles both updateEntityListItems (bulk) and updateEntityListItem (single)

## Additional context

https://github.com/user-attachments/assets/27265e63-2eca-4c87-8b01-50dce94cde30

### DIFF
--- a/shared/src/api/queries/overview/refetchFilteredEntities.ts
+++ b/shared/src/api/queries/overview/refetchFilteredEntities.ts
@@ -83,52 +83,79 @@ export const refetchTasksForCacheEntry = async ({
   }
 }
 
-interface UpdateOverviewTasksWithDataArgs {
+interface RefetchOverviewTasksForCacheEntryArgs {
   dispatch: ThunkDispatch<any, any, UnknownAction>
-  overviewTasksEntries: any[]
-  allFetchedTasks: any[]
+  projectName: string
+  updatedTaskIds: string[]
+  cacheEntry: any
 }
 
 /**
- * Updates the overview tasks cache with pre-fetched task data.
- * This avoids redundant API calls by reusing data fetched from filtered queries.
+ * Refetches overview tasks with the cache entry's specific filter and updates the cache.
+ * This ensures that tasks are correctly filtered and removed if they no longer match.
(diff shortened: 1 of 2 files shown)


Innders (approved):
Works really well now thanks! 

The next step is to do the same for the lists page :) https://github.com/ynput/ayon-frontend/issues/1583

PR #1582, opened 2025-10-15
Lists: Adding to list gets stuck on first item

filipvnencak:
## Description of changes 
Refactored menu item handling by introducing a recursive `rebindCommands` function.  
This ensures that command-based list items are correctly rebound with the current selection context

## Technical details

- Replaced the previous shallow `.map()` call with a recursive traversal.
- Rebuilds only items with a `command`, preserving other properties.
- Maintains nested submenu structure using recursion.
- Fixes incorrect command bindings caused by captured references.


Innders (changes requested):
Original issue fixed 👍 

But now the add to list menu item appears twice?

Image

Innders (approved):

PR #1585, opened 2025-10-15
Overview: Multi-select items should fill row height

filipvnencak:
## Description of changes 
Fixed multi-select enum tags stretching to full width when wrapping is enabled in multi-row cells. Tags now maintain their natural width.

## Technical details

- Added $maxHeight prop to StyledValuesContainer to constrain height when wrapping
- Applied max-height only when .wrap class is active (row height > 60px)
- Subtracted 10px from $maxHeight to account for padding and prevent overflow
- Changed wrap threshold from 40px to 60px row height for better UX

## Additional context

The flex-wrap: wrap; property enables the multi-row layout, but there is an issue with tag widths — the CSS doesn’t know it should shrink the tags to a min-width of 20px and display ellipses.

Screenshot 2025-10-15 at 13 00 57
Screenshot 2025-10-15 at 13 56 11

### DIFF
--- a/shared/src/containers/ProjectTreeTable/widgets/EnumCellValue.tsx
+++ b/shared/src/containers/ProjectTreeTable/widgets/EnumCellValue.tsx
@@ -32,9 +32,22 @@ const StyledWidget = styled.div`
       background-color: var(--md-sys-color-primary-container-hover);
     }
   }
+
+  /* when value container is higher than 55px, wrap the contents */
+  &:not(.item) {
+    container-type: size;
+
+    @container (min-height: 55px) {
+      .values {
+        flex-wrap: wrap;
+        align-items: flex-start;
+        max-height: 100%;
+      }
+    }
+  }
 `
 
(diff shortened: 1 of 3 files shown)


shared/src/containers/ProjectTreeTable/widgets/EnumCellValue.tsx
```diff
@@ -34,18 +34,21 @@ const StyledWidget = styled.div`
...
   overflow: hidden;
   border-radius: var(--border-radius-m);
   padding: 0px 2px;
   flex-wrap: nowrap;
-
+  
   &.wrap {
     flex-wrap: wrap;
+    align-items: flex-start;
+    max-height: ${({ $maxHeight }) => ($maxHeight ? `${$maxHeight- 10}px` : 'unset')};
```
Innders: Don't over complicate, this can just be `max-height: 100%;`?

filipvnencak:
Tested and working. Looks great! ✅

PR #1586, opened 2025-10-16
Lists: Updates should trigger background refresh of item

filipvnencak:
## Description of changes 
Implements filter-aware cache updates and calculated attribute refetching for Lists page, matching PR  #1575 behavior.

**Problem**
Calculated attributes weren't refreshing and stale filtered items remained in cache.

**Solution**

- Added background refetch after list updates
- Refetches data with original filters
- Removes items no longer matching filters
- Smart loading states prevent unnecessary indicators

## Technical details

- refetchFilteredListItems.ts (new) - Core refetch logic
- updateOverview.ts - List cache refetch on entity operations
- updateLists.ts - Refetch after bulk updates
- useGetListItemsData.ts - Smart loading state hook

## Additional context

- Updated items show fresh calculated values
- Items not matching filters disappear
- Loading indicators only show on filter changes

### DIFF
--- a/shared/src/api/queries/entityLists/updateLists.ts
+++ b/shared/src/api/queries/entityLists/updateLists.ts
@@ -241,7 +241,6 @@ const updateListsEnhancedApi = entityListsApi.enhanceEndpoints({
         }
       },
       invalidatesTags: (_s, _e, { listId, listItemId }) => [
-        { type: 'entityList', id: listId },
         { type: 'entityListItem', id: listId },
         { type: 'entityListItem', id: listItemId },
       ],


Innders (changes requested):
The code technically works but I think it's overcomplicated.

Let's break down exactly what is happening...

1. Create tags for invalidation
2. Find caches that _would_ be invalidated by those tags
3. Refetch data for each one of those caches
4. Patch the new data into the cache so it's up to date.

This is basically exactly how `invalidateTags` works under the hood so we should just use that. `invalidateTags` does a full re-fretch and then updates the cache, just like we are doing here manually.

- [ ] Use invalidateTags instead to trigger cache refresh

The reason we do it manually in the overview is that we only re-fetch data for a specific folder/task, not the full query so it's more efficient. Here we are re-fetching the full request anyway so there's no advantage of doing it manually.

shared/src/api/queries/entityLists/updateLists.ts
```diff
@@ -240,6 +240,11 @@ const updateListsEnhancedApi = entityListsApi.enhanceEndpoints({
           })
         }
       },
+      invalidatesTags: (_s, _e, { listId, listItemId }) => [
+        { type: 'entityList', id: listId },
```
Innders: Why does the list itself need updating?

PR #1597, opened 2025-10-22
Project Access: Filtering broken

filipvnencak:
## Description of changes 
Fixed an issue where user filters in the Project Manager > Users page were not updating in real-time within the running application. Filters would only persist and display correctly after a page refresh.

## Technical details

 The `useUserPageFilters` hook was reading `frontendPreferences` from the Redux state slice using `useAppSelector()`. When filters were updated, the
  `setFrontendPreferences` mutation performed an optimistic update to the RTK Query cache, but this update wasn't reflected in the Redux state selector until the next page
  load/refetch.

  ### Solution
  Changed the data source from Redux state selector to RTK Query's `useGetCurrentUserQuery()` hook. This allows the component to:

1. Subscribe directly to the RTK Query cache
2. Receive optimistic updates immediately when preferences are saved
3. Follow the same pattern used elsewhere in the codebase (e.g., `useProjectListUserPreferences`)

### DIFF
--- a/src/pages/ProjectManagerPage/Users/hooks.ts
+++ b/src/pages/ProjectManagerPage/Users/hooks.ts
@@ -1,11 +1,10 @@
 import { $Any } from '@types'
 import { useState } from 'react'
 import { SelectionStatus } from './types'
-import { Filter } from '@components/SearchFilter/types'
-import { useAppSelector } from '@state/store'
 import { useSetFrontendPreferencesMutation, useGetCurrentUserQuery } from '@shared/api'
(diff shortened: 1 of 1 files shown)


Innders (changes requested):
Adding the filters works now but there is another annoying bug. When you click the access group header the filters disappear for some reason?

- [x] Fix access groups not opening/closing
- [x] Fix clicking access groups removing filters

https://github.com/user-attachments/assets/f8c78bd9-253e-41d2-aee9-fcc3afee0ead

PR #1599, opened 2025-10-22
Bundle Page / Lists of Addons - should be aligned for all circumstances

filipvnencak:
## Description of changes 
 Fixed addon list alignment issues in bundle comparison view. When comparing bundles with different addons (e.g., Bundle A has AfterEffects + Maya, Bundle B has Maya +
  Nuke), the lists were misaligned making it difficult to compare addon versions side-by-side.

  **Changes:**
  - Modified `BundleCompare.tsx` to create a unified addon list across all bundles being compared
  - Pre-process bundle data to ensure each bundle has entries for all addons (with `null` for missing ones)
  - Updated `BundlesAddonList.tsx` to use consistent title-based sorting in `readOnly` mode (comparison view)
  - Added logic to synchronize sorting across comparison bundles

  **Result:** All bundle columns now show the same addons in the same order, with missing addons displayed as "NONE", making version comparison significantly easier.

## Technical details

  - In comparison mode (`readOnly=true`), all addon lists now sort by title only, regardless of bundle type (project vs studio)
  - In edit mode, project bundles retain their original sorting behavior (by `projectCanOverrideAddonVersion` first, then title)
  - The normalization process preserves all original bundle properties while extending the addons object
  -  Added `handleSort` logic to synchronize sorting across bundles

### DIFF
--- a/src/pages/SettingsPage/Bundles/BundleCompare.tsx
+++ b/src/pages/SettingsPage/Bundles/BundleCompare.tsx
@@ -2,6 +2,7 @@ import { Panel, Section } from '@ynput/ayon-react-components'
 import React, { useRef, useState } from 'react'
 import BundlesAddonList from './BundlesAddonList'
 import useScrollSync from '@hooks/useScrollSync'
+import { DataTableSortEvent } from 'primereact/datatable'
 
 type Bundle = {
   name: string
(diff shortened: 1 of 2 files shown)


Innders (changes requested):
- [x] Sorting needs to happen for all tables to avoid misalignment.

Image

Innders (approved):
Works great thanks

PR #1601, opened 2025-10-23
NewServiceDialog: configurable service environment

tweak-wtf:
## Description of changes 
This PR adds a new FormRow that lets users set environment variables for the containers spawned by ash.

## Additional context
This is the modified Form
image
And here's an example of the sg-leecher running with debug logging enabled
Screenshot 2025-10-23 at 17 14 43


MustafaJafar (commented):
I can confirm that the env is added to service data.. 
But, I can't confirm the env is propagated to the service container as expected as I'm facing some issues in spawning services (not related to this PR).
image

Innders (approved):
Looks good to me!

PR #1603, opened 2025-10-24
Bundles: Addon search filter

filipvnencak:
## Description of changes 
 **Key Features:**
  - Search input field to filter addons by name or title (case-insensitive)
  - Persistent search state across sessions using localStorage
  - Automatic filtering on page load if previous search exists
  - Search functionality available in both creation and view modes

## Technical details

  **Implementation:**
  - Custom React hook (`useAddonSearch.ts`) handles all search logic and state management
  - Uses `useLocalStorage` hook to persist search across sessions
  - Filters addons by `name`field
  - Search state stored under key `bundles-search` in localStorage
  - Race condition prevention using `useRef` to ensure initial filter only runs once

  **State Management:**
  - Search value persisted to localStorage automatically
  - Search filter applied on component mount if value exists
  - Manual addon selection preserved during addon updates (no overwrites)

## Additional context

  **Before:**
  Users had to manually scroll through entire addon list to find specific addons, especially problematic with 20+ addons.

  **After:**
  Type a few characters to instantly filter matching addons. Search persists across page reloads and navigation.

### DIFF
--- a/src/pages/SettingsPage/Bundles/AddonSearchContext.tsx
+++ b/src/pages/SettingsPage/Bundles/AddonSearchContext.tsx
@@ -1,4 +1,11 @@
-import React, { createContext, useContext, ReactNode, useMemo, useCallback, ChangeEvent } from 'react'
+import React, {
+  createContext,
+  useContext,
+  ReactNode,
+  useMemo,
+  useCallback,
+  ChangeEvent,
+} from 'react'
 import { useLocalStorage } from '@shared/hooks'
 import type { Addon } from './types'
 import { matchSorter } from 'match-sorter'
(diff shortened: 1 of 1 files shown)


Innders (changes requested):
- [x] Filtering doesn't work, I can still see all the addons.
- [x] Remove auto selection of addons, that is not filtering.
- [ ] Increase width of input when viewing bundle.
- [ ] When filtering, add a little message underneath the last addon like "3 addons filtered out, remove search filter". In faded outline color text. This is just so that users don't get confused when they see less addons.

Image

Innders (approved):
Nice one thanks!

PR #1610, opened 2025-10-29
Slideout: Detach feed selection state

filipvnencak:
## Description of changes 
Fixed state sharing issue between main DetailsPanel and PiP (picture-in-picture) window where switching tabs (activity, versions, etc.) in one window would affect the other window.

## Technical details

- Changed `currentTab` state to be prop-based instead of context-based
- `FeedWrapper` now accepts `currentTab` and `setCurrentTab` as props from parent components
- `DetailsPanel` passes its scoped `currentTab` state from `useScopedDetailsPanel`
- `DetailsPanelFloating` creates its own independent local `currentTab` state
- Fixed TypeScript types to correctly reflect that `editingId` is local state, not a prop

## Additional context

 This fix ensures that when users open a PiP window and switch between tabs (activity/versions/etc.) in either window, the other window maintains its own independent tab selection.

https://github.com/user-attachments/assets/8ea2b585-f9b8-45bd-a6eb-a40dfc8f0ac6


Innders (changes requested):
- [x] Weirdness when going to the details tab

Image

Innders (approved):

PR #1617, opened 2025-10-29
Table: Thumbnail column width scale with row height change

filipvnencak:
## Description of changes 
Thumbnail column now maintains its aspect ratio when row height is adjusted. As row height increases/decreases, the thumbnail width automatically adjusts proportionally to keep images properly sized.

## Technical details

- Locks aspect ratio at start of row height adjustment session using `lockedAspectRatioRef`
- `updateRowHeight`: Provides instant visual feedback during slider drag
- Thumbnail width is calculated based on current ratio: `thumbnailWidth / rowHeight`
- Clears internal state and locked ratio after persistence completes

## Additional context

https://github.com/user-attachments/assets/461c740b-f98b-4528-b44c-c62e36bbc20e


Innders (approved):
Works great thanks!

PR #1625, opened 2025-10-30
Inbox: Publishing Grouping

filipvnencak:
## Description of changes 
This PR implements smart grouping for `version.publish` and `reviewable` inbox messages to prevent inbox flooding when multiple product versions are published simultaneously under the same folder.

  ### Problem
 When publishing multiple product versions at once (e.g., reviewQrcode, reviewImage1, reviewImage2), the inbox was flooded with individual messages making it unusable. Each published version created separate inbox entries, overwhelming users with redundant notifications.

  ### Solution
 Messages are now grouped using a two-tier strategy:

1. **Time Window Grouping**: Messages within 5-minute intervals are considered related
2. **Folder Hierarchy Grouping**: Within each time window, messages are grouped by their parent folder

When multiple versions are published under the same folder within a 5-minute window:
- They appear as **one row** showing the parent folder name (e.g., "01 Schnaeryouds")
- Clicking opens the detail panel with the parent folder context
- All individual version messages remain accessible via the `messages` array

When only a single version is published:
- Falls back to normal behavior showing the specific version details
- No grouping is applied to avoid confusion

## Technical details

1. Filter version.publish + reviewable messages
2. Group by 5-minute time windows (rounded down)
3. Within each window, group by parent folder ID
4. Mark messages as visited to prevent duplicate processing
5. Use parent folder as entity for grouped rows

 ### Key Implementation Details

- **Time Window**: `Math.floor(date.getMinutes() / 5) * 5` creates consistent 5-minute buckets
- **Parent Folder Detection**: Uses `activityData.parents[0]` to identify folder hierarchy
- **Fallback Behavior**: Single messages skip special grouping and use standard inbox display
- **WebSocket Compatible**: Works with cache invalidation + refetch pattern in RTK Query

  ### Limitations

- Groups are recalculated on full message array updates (efficient due to `useMemo`)
- 5-minute window is hardcoded (not configurable)
- Messages without parent folder metadata fall back to individual display

## Additional context

Before:
Screenshot 2025-10-30 at 14 50 40
After:
Screenshot 2025-10-30 at 14 49 18

Screenshot 2025-11-06 at 16 00 18

### DIFF
--- a/src/pages/InboxPage/hooks/useGroupMessages.js
+++ b/src/pages/InboxPage/hooks/useGroupMessages.js
@@ -399,8 +399,8 @@ const transformGroups = (groups = []) => {
         // Collect product names from the group
         const productNames = []
         group.forEach(msg => {
-          // Get product name from activityData
-          const productName = msg.activityData?.product?.name || msg.activityData?.productName
+          // Get product name from activityData.context
+          const productName = msg.activityData?.context?.productName
           if (productName && !productNames.includes(productName)) {
             productNames.push(productName)
           }
@@ -466,6 +466,7 @@ const transformGroups = (groups = []) => {
 }
 
 const useGroupMessages = ({ messages, currentUser }) => {
+  console.log(messages)
   const grouped = useMemo(() => {
     // const simpleGroups = messages.map((message) => [message])
     const simpleGroups = groupMessages(messages, currentUser)


Innders (approved):
I haven't tested this yet but I just found another scenario where it would be useful to group messages together. When there are a lot of assignment changes in a short space of time we could group them together so that it just shows from -> to.

For example here Luke was removed and then Filip added.

Image

It would be great if this could be one message `James: reassigned development from Luke to Filip`

`James: reassigned {entity_name} from {first_removed} to {last_added}`.

Give it a one minute grouping threshold. It should also support multiple reassignments.

Innders (changes requested):
Cool, I like the swap icon.

- [x] Missing author at the start `Tom: Reassigned...`

I found a case where it doesn't make sense. When someone was added and then removed.

- [x] Hide messages where the final result is the same as the starting assignees.

Image

I think we could handle these situations better, where multiple assignments are happening. Do you think it's possible to group all assignments into that time frame and then resolve them all into one message.

- [x] Group all remove/add messages in that time frame and should final resolve, all removed to all added.

This example would end up as `Tom: Reassigned layout from Ava Williams, Sophia Miller to Jackson Wilson, Isabella Davis`

Image

This example would end up as `Tom: Reassigned layout from Ronja Hakala to Ava Williams, Luke, Becky`.

Image

BigRoy:
> When there are more than 4 it should be `Tom: Published 5 new versions`.

Could this be:
-> Tom published versions to X products?

Or if it may even group multiple versions of the same product as well if it's within the time constraint, then perhaps:

-> Tom published X new versions to Y products.

Innders (changes requested):
- [ ] Use product names not versions. `published versions to {product_name}, {product_name}`.
image

PR #1626, opened 2025-10-30
Launcher/dependencies manager: Inability to easily cancel/close the window

filipvnencak:
## Description of changes 
 Replace `hideCancelButton` with `onClose` prop

### DIFF
--- a/src/components/AddonDialog/AddonDialog.jsx
+++ b/src/components/AddonDialog/AddonDialog.jsx
@@ -55,7 +55,8 @@ const AddonDialog = ({ uploadOpen, setUploadOpen, uploadHeader, manager }) => {
       header={uploadHeader || 'Upload addon'}
       size="md"
       tabIndex={-1}
-      onClose={()=>setUploadOpen('')}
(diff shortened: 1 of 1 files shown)


src/components/AddonDialog/AddonDialog.jsx
```diff
@@ -55,7 +55,7 @@ const AddonDialog = ({ uploadOpen, setUploadOpen, uploadHeader, manager }) => {
       header={uploadHeader || 'Upload addon'}
       size="md"
       tabIndex={-1}
-      hideCancelButton
+      onClose={()=>setUploadOpen('')}
```
Innders: Looking at the blame martin added hideCancelButton to "prevent closing upload dialog by clicking outside". So there was a reason why we decided to hide the close button 2 months ago.

For a change like this we always need to check what the original reason was and ensure we maintain that original fix.

Your change reverses the fix for "prevent closing upload dialog by clicking outside". You will need to now also fix this again. (Use prop `enableBackdropClose={false}`)

Innders (changes requested):

PR #1640, opened 2025-11-05
Projects: Ensure projects cannot be selected in the background (archived)

filipvnencak:
## Description of changes 
 Fixed issue where archived projects could be deleted when hidden from the list after toggling "Show archived" off while the project remained selected.

## Technical details

 Added `!singleProject` check to delete button disabled condition in `useProjectsListMenuItems.ts:165`. This ensures the selected project exists in the currently visible projects list before allowing deletion.


Innders (changes requested):
The core issue here has nothing to do with the delete button; it's the fact the archived project is still selected in the background. So you might fix the delete button but I can still perform actions on a project that is not visible.

- [ ] When deactivating a project and "show archived" is off, ensure the selection of the archived project is removed.
- [ ] When toggling "show archived" to **off** ensure the selection of the archived project is removed.

https://github.com/user-attachments/assets/d5193ad0-4674-4807-bcc5-853690d3e228

PR #1649, opened 2025-11-10
Views: Default views

filipvnencak:
## Description of changes 
- Introduced a new button to allow admin users to set a default project view.
- Added confirmation dialog and API integration for setting and sharing default views.
- Updated styles and context to support new "Make Default View" action.

### DIFF
--- a/shared/src/containers/Views/ViewsMenuContainer/ViewsMenuContainer.tsx
+++ b/shared/src/containers/Views/ViewsMenuContainer/ViewsMenuContainer.tsx
@@ -35,7 +35,7 @@ export const ViewsMenuContainer: FC = () => {
   const modalRef = useRef<HTMLDivElement>(null)
 
   const { user } = useGlobalContext()
-  const isAdmin = (user?.uiExposureLevel || 0) >= 700
+  const isAdmin = (user?.uiExposureLevel || 0) >= 900
   // Modal position calculation
   const portalContainer = getViewsPortalContainer(viewType)
   const buttonRect = portalContainer?.getBoundingClientRect()


Innders (changes requested):
Getting a lot closer, the idea definitely works and is initiative to use. The actual technical stuff is working. Now just to polish the UI and UX. 💅 

- [x] Use figma design for buttons and section.

Image

- [x] Buttons should not be text selectable.

Image

- [x] To update a base view you must first remove it. This dialog should just ask if you want to remove base the view. Once it's removed you can click it again to update.

Image

- [x] If there is a base view set already the chip needs to have the filled background with no dashed border because we have "filled" that slot.

Image

- [ ] You should be able to click the whole chip not just the icons, the whole thing is a button.

- [ ] Reset button always visible. At the moment it's only on hover.

Image

- [ ] Reset button should use the nice blue background if the working view is selected. See the more button for how it works and looks.

Image

Innders (changes requested):
- [x] The buttons need to go above the "Create new view" button as per design.

Image

The **whole** thing is just a normal regular button with some different styling and altering icons.

- [x] Use Button component
- [x] Remove icon colours
- [x] When hovering an "empty" slot the border and icon should be white and the background `--md-sys-color-surface-container-high-hover` (like a normal button).

Image

_hovering a empty slot_
image


- [ ] When removing an base view the main dropdown of the views panel should stay open (currently it closes on confirming).

Image

- [ ] When a view is selected and then you press reset to default button it will prompt you to save that view you used to be on, this doesn't make sense. It should just put you on the working view and reset it to default and that's it.

https://github.com/user-attachments/assets/4846be9e-acae-49e4-87e2-becf57c1b179

- [ ] When setting a default view, the order of the views moves about for a bit?

https://github.com/user-attachments/assets/c5290c80-0d97-4fb4-b392-3daf41a6bafc

Innders (changes requested):
- [x] Remove divider above the default views.

Image

shared/src/containers/Views/ViewsMenuContainer/BaseViewsTags.tsx
```diff
@@ -0,0 +1,99 @@
...
+    </Styled.BaseViewsContainer>
+  )
+}
+export default BaseViewsTagContainer
+
+type ScopeIconProps = {
+  existingView: boolean
+  onClick: () => void
+  label: string
+  poweLicense?: boolean
```
Innders: typo

Innders (changes requested):
- [x] Default Views alignment off

Image

- [x] When adding a new default view, the optimistic updates are not working and you have to wait for the request to finish before it turns the correct colour. This creates a slow experience. Users expect things to happen instantly.

https://github.com/user-attachments/assets/096293c0-2947-41f8-be07-c5bdc7ac981a

- [x] Powerpack issues https://discord.com/channels/517362899170230292/1437708986412568618/1445791724504088778

Innders (approved):
This one turned out really nice, great job!

The optimistic update are still not working but we can fix that another time, it's not a big deal.

PR #1658, opened 2025-11-17
Shareable entity links

Innders:
### URL Entity Syncing

Now when opening the entity panel, the url is synced with the entity. This makes it easy to share entities with other users just by sending them the link.

You need three URL search params to link to to a specific entity.

`project={project_name}`
`type={entity_type}`
`id={entity_id}`

_example_
`.../projects/Quantum_Demo/products?project=Quantum_Demo&type=version&id=ece66a4b39ff11ef8dca14ac6003f3c1`

### URI Syncing

`uri` search params are also supported and will open the entity when navigated to.

image

### Activity Links
Now that we can directly open any entity, we can also create links to activities. This is done through the search param `activity`.

To get a link to an entity, use the more menu on a comment.

image

The copied link will always either point to the **overview** or **products** page based on the entity type.

### Testing

Open an entity, copy the link and then paste again to see if you return to the same entity.

**Pages to test**
- Home page
- Overview (folders and tasks)
- Task progress
- Products
- Lists
- Reviews

### DIFF
--- a/shared/src/containers/DetailsPanel/DetailsPanel.tsx
+++ b/shared/src/containers/DetailsPanel/DetailsPanel.tsx
@@ -416,6 +416,7 @@ DetailsPanelProps) => {
               exportAnnotationComposite={exportAnnotationComposite}
               currentTab={currentTab}
               setCurrentTab={setTab}
+              isSlideOut={isSlideOut}
             />
           )}
           {currentTab === 'files' && (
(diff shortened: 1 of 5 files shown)


martastain:
Originally, when multiple entities were selected, the link contained the last selected one. Now the link points to the first selected one. Is that intentional? The original behavior felt more logical to me.

filipvnencak (commented):
When two Detail panels are open at the same time and I click the more "(…)” button inside a comment, the dropdown menu is rendered in both Detail panels instead of just the one I interacted with.

Image


In the PIP container, the Copy button is rendered below the PIP container instead of inside it. It also appears three times, and it doesn’t follow the container when it moves.

https://github.com/user-attachments/assets/84358e89-0a86-47cd-903e-8d525d5a60d0

PR #1665, opened 2025-11-21
YN-0212: product name filter on Products page

filipvnencak:
## Description of changes 
Added a product name filter to the Products page (VersionsProductsPage) allowing users to filter products by typing their names.

## Technical details

- Added new `'productName'` filter type to `FilterFieldType` union in `useBuildFilterOptions.tsx`
- Created filter option with `allowsCustomValues: true` to enable free text input with suggestions
- Filter is scoped to the `'product'` scope only
- Product names are extracted from `productsMap` in `VPSearchFilter.tsx` and passed as suggestions
- Filter ID is set to `'name'` (maps to the product's `name` field in GraphQL)

### DIFF
--- a/shared/src/components/SearchFilter/useBuildFilterOptions.tsx
+++ b/shared/src/components/SearchFilter/useBuildFilterOptions.tsx
@@ -683,10 +683,10 @@ const getOptionRoot = (
       break
     case 'productName':
       rootOption = {
-        id: getRootIdWithPrefix(`name`),
+        id: getRootIdWithPrefix(`productNames`),
         type: 'string',
         label: formatLabelWithScope(`Product Name`),
-        icon: getAttributeIcon('product'),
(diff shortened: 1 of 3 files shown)


shared/src/components/SearchFilter/useBuildFilterOptions.tsx
```diff
@@ -659,6 +681,22 @@ const getOptionRoot = (
         operatorChangeable: false,
       }
       break
+    case 'productName':
+      rootOption = {
+        id: getRootIdWithPrefix(`name`),
+        type: 'string',
+        label: formatLabelWithScope(`Product Name`),
+        icon: getAttributeIcon('product'),
```
Innders: This will give it the "text" icon which I think it more appropriate.

```suggestion
        icon: getAttributeIcon('productName', 'string'),
```

Innders (approved):
Works well, great job!

PR #1671, opened 2025-11-27
Links: Show links the user does not have access to

filipvnencak:
## Description of changes

  This PR adds support for displaying restricted links in the ProjectTreeTable. Previously, when a user had a link to an entity they didn't have permission to view (node is
   null from the API), those links were silently filtered out. Now, users can see that restricted links exist, similar to how other parts of the application show restricted
   items.

  **Changes:**
  - Restricted links are now included in the link list instead of being filtered out
  - Added `isRestricted` flag to `EntityLink` and `LinkEntity` types to track restricted links
  - Updated LinksWidget to display lock icons for restricted links in table cells
  - Updated LinkManagerItem to show "You don't have permission to view this link" message with visual styling

  ## Technical details

  **Implementation:**
  1. Modified `getEntityLinks.ts` to include links with `null` nodes instead of filtering them out
  2. Updated `EntityLink` type to allow `node: null` and added optional `isRestricted` flag
  3. Updated `linksToTableData.ts` to handle restricted links by creating LinkEntity objects with empty labels
  4. Modified `ChipValue` type to accept `React.ReactElement` for icons (was incorrectly typed as DOM `Element`)
  5. Updated LinksWidget and LinkManagerItem components to handle and display restricted links appropriately

  **Limitations:**
  - Restricted links cannot be clicked or removed (by design, since users don't have access)
  - The label is empty for restricted links since no node data is available from the API

  ## Additional context

  This change aligns with the design pattern used elsewhere in the application where restricted items are shown but clearly marked as inaccessible. Console logging has been
   added at multiple points to aid in debugging and identifying which specific links are restricted (including linkId, linkType, direction, and entityType).

### DIFF
--- a/shared/src/components/Chips/Chips.tsx
+++ b/shared/src/components/Chips/Chips.tsx
@@ -20,6 +20,10 @@ const Chip = styled.div`
   text-overflow: ellipsis;
   cursor: pointer;
 
+  display: flex;
+  align-items: center;
+  gap: 4px;
+
   flex-shrink: 0;
   min-width: auto;
   &.last {
@@ -65,7 +69,7 @@ const AddIcon = styled(Icon)`
 export type ChipValue = {
   label: string
   tooltip: string
-  icon?: React.ReactElement
+  icon?: string
 }
 
 interface ChipsProps {
@@ -174,7 +178,7 @@ export const Chips: FC<ChipsProps> = ({ values, disabled, pt }) => {
             pt?.chip?.className,
           )}
         >
-          {chip?.icon}  {chip.label}
+          {chip?.icon && <Icon icon={chip.icon}/>}  {chip.label}
         </Chip>
       ))}
       {hiddenCount > 0 && (
--- a/shared/src/components/LinksManager/LinkManagerItem.tsx
+++ b/shared/src/components/LinksManager/LinkManagerItem.tsx
@@ -28,21 +28,21 @@ export const LinkManagerItem: FC<LinkManagerItemProps> = ({
       onClick={() => isClickable && onEntityClick?.(link.entityId, link.entityType)}
       data-tooltip={
         link.isRestricted
-          ? ""
+          ? "Access Restricted - Insufficient Permissions to Entity"
           : link.parents.join('/') + '/' + link.label
       }
       className={clsx({
         clickable: isClickable,
         selected: isSelected,
+        restricted: link.isRestricted,
       })}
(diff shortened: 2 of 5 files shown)


shared/src/components/LinksManager/LinkManagerItem.tsx
```diff
@@ -20,35 +20,52 @@ export const LinkManagerItem: FC<LinkManagerItemProps> = ({
...
+        link.isRestricted
+          ? ""
+          : link.parents.join('/') + '/' + link.label
+      }
       className={clsx({
-        clickable: entityTypeSupported,
+        clickable: isClickable,
         selected: isSelected,
       })}
+      style={link.isRestricted ? { opacity: 0.5, fontStyle: 'italic' } : undefined}
```
Innders: Small thing but could we use classNames to assign the restricted styles instead. `restricted: link.isRestricted`. There are two reason for this.

1. We keep all the CSS in the same file so it's less fragmented.
2. The restricted className is now on the DOM element so it makes it easier to debug. We can see these styles are applied because of the restricted className without the need to look at the code.

shared/src/components/LinksManager/LinkManagerItem.tsx
```diff
@@ -20,35 +20,52 @@ export const LinkManagerItem: FC<LinkManagerItemProps> = ({
...
 
   return (
     <Styled.LinkItem
       key={link.linkId}
-      onClick={() => entityTypeSupported && onEntityClick?.(link.entityId, link.entityType)}
-      data-tooltip={link.parents.join('/') + '/' + link.label}
+      onClick={() => isClickable && onEntityClick?.(link.entityId, link.entityType)}
+      data-tooltip={
+        link.isRestricted
+          ? ""
```
Innders:
```suggestion
          ? "Access Restricted - Insufficient Permissions to Entity"
```

shared/src/components/Chips/Chips.tsx
```diff
@@ -173,7 +174,7 @@ export const Chips: FC<ChipsProps> = ({ values, disabled, pt }) => {
             pt?.chip?.className,
           )}
         >
-          {chip.label}
+          {chip?.icon}  {chip.label}
```
Innders: 1. Use <Icon /> component in here instead of LinksWidget.
2. Ensure the icon is small enough so that it does not increase the base height of the chip. At the moment this creates a bigger chip and the heights are misaligned with links that are not restricted. Current chip height is 24px.

PR #1673, opened 2025-11-27
Feed: Show new reviewables in realtime

filipvnencak:
## Description of changes 

Fixed real-time activity updates not appearing for related entities (e.g., version publish activities in task feeds).

## Technical details

WebSocket activity handler now checks all reference types (origin, relation, mention) to match query behavior, previously only checked origin references.


Innders (approved):
Seems to work!

Issue #1598, opened 2025-10-22
Project Settings: wrong dev bundle


antirotor:
## Story
This happens when you have more than 1 dev bundle on the server. When you go to Project settings, there's a name of the bundle displayed under variants and the list of addons in the bundle:

<img width="1012" height="313" alt="Image" src="https://github.com/user-attachments/assets/22baff54-c41e-4ff0-b522-c3ca66cb0e5e" />

In my case, I have two dev bundles named `dev-02` and `dev-03` - the current one assigned to me is `dev-03`, yet when I go to the Project settings, it will show `dev-02` (along with its addons) - from the reports I have it looks like it will always default to second dev bundle it finds.

Only when you click on the `[DEV]` variant button again, it will show current dev bundle (in my case `dev-03`)

### To reproduce

1. Have at least two dev bundles on the server
2. Go to the project settings
3. Change bundle name and addon list
4. Click on **DEV** variant button and you should see the correct bundle.

### Expected behavior
Current dev bundle (the one associated with current user) should be displayed right away.

## Environment

- **Server version**:  1.12.4+DEVELOP
- **Server host OS**: Docker
- **Browser**: Version 141.0.7390.77 (Official Build) (64-bit)

Issue #831, opened 2024-09-30
AY-6854_GUI: Settings tab always defaults to Production


jm22dogs:
## Story
Opening the settings tabs will always load the production variant, instead of keeping the one you were working on previously. This may be dangerous in cases where you are fiddling around with the staging settings, and at some point you change tabs for whatever reason, and when you come back to the settings tab you may think you are still where you left it at (staging settings), but you are now actually happily modifying production settings unaware.

![settings_reset_to_production_tab](https://github.com/user-attachments/assets/3ed4f270-40d2-4e2f-ba0b-787073110bed)

### To reproduce

1. Click on settings tab, select staging
2. Click on some other tab
3. Go back again to the settings tab, you are now in production instead of staging.

### Expected behavior
I'd expect the settings variant to stick through sessions.

## Environment

- **1.4.2+202409231943n**: 
- **Linux**: 
- **Firefox**: 

[CU Ticket](https://app.clickup.com/t/6658547/AY-6854)

Innders:
@martastain This would probably be quite simple but would require moving the state to the redux store?

Innders:
@filipvnencak please just store the state in local storage so it is maintained across page loads. Do not worry about the difference between studio and project levels, it can just be the same stored state for both.

PR #1674, opened 2025-11-28
AY-6854_GUI: Settings tab always defaults to Production

filipvnencak:
## Description of changes 

 - Replaced `useState` with `useLocalStorage` for persisting selected bundle variant (`production`/`staging`/`dev`) across sessions
  - Removed unnecessary `selectedBundle` and `selectedType` logic
  - Simplified button active state to directly compare with `variant` prop

## Technical details

The `selectedType` calculation checked bundle flags in order (`isProduction` → `isStaging` → `isDev`), which could incorrectly return `'production'` even when
   staging was selected.

  **After:** Button highlighting now directly compares `variantType === variant`, which is simpler and always correct.

  **localStorage key:** `'variant-type'` - stores the selected variant for Studio Settings page

## Additional context

This PR fixing the #831 issue

### DIFF
--- a/src/containers/AddonSettings/AddonSettings.jsx
+++ b/src/containers/AddonSettings/AddonSettings.jsx
@@ -727,7 +727,6 @@ const AddonSettings = ({ projectName, showSites = false, bypassPermissions = fal
     setCurrentSelection(null)
   }
 
-
   const onUpdateAddonSchema = (addonName, schema) => {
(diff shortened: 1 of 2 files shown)


BigRoy:
Does this by any chance also influence or fix https://github.com/ynput/ayon-frontend/issues/1598 ?

filipvnencak:
@BigRoy No, this is separate bug.

Innders (approved):

PR #1677, opened 2025-11-28
Permissions: Dashboard all users switch for List all users permission

### LINKED ISSUE #958: Permissions: Dashboard all users switch for List all users permission
This should appear when a regular user has the `List all users` permission.

Image

filipvnencak:
## Description of changes 
 Add support for `studio.list_all_users` permission to allow non-admin users to view all users' tasks in the Dashboard.

  - Added `canListAllUsers()` method in `useUserProjectPermissions` hook
  - Updated `DashboardTasksToolbar` to show "Me/Users" switch for users with this permission

## Technical details

 The permission check follows the existing pattern:
  1. Admins/managers bypass all permission checks (`hasElevatedPrivileges = true`)
  2. Regular users with `list_all_users = true` can view all users
  3. Regular users without this permission are restricted to viewing only their own tasks

## Additional context

  Closes #958


Innders (changes requested):
You can see the switch but it doesn't seem to affect the actual kanban tasks when used.

Image

PR #1687, opened 2025-12-03
Links: When linking to a specific comment ensure "all activity" tab selected

filipvnencak:
## Description of changes 

Fixed syntax error and implemented automatic 'activity' tab selection when DetailPanel opens from shared URLs. When users share entity/comment links (introduced in #1658 ), clicking these links now automatically opens the DetailPanel on the 'activity' tab, ensuring the shared content is immediately visible.

- Added currentScope state tracking in DetailsPanelContext to identify which scope is active
- Updated URL parameter handling to set tab to 'activity' for the current scope when ?project=...&type=...&id=... params are present

## Technical details

  - Uses currentScope state registered by useScopedDetailsPanel hook to determine which scope's tab to update
  - Only updates tab setting for the active scope, preserving tab preferences for other scopes
  - Changes persist in localStorage via details/tabs-by-scope key

### DIFF
--- a/shared/src/context/DetailsPanelContext.tsx
+++ b/shared/src/context/DetailsPanelContext.tsx
@@ -1,16 +1,15 @@
 import React, {
   createContext,
-  useContext,
-  useCallback,
   ReactNode,
-  useState,
+  useCallback,
+  useContext,
   useEffect,
+  useState,
 } from 'react'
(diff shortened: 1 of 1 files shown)


Innders (changes requested):
I think this needs a slight re-think :)

Innders (approved):

PR #1695, opened 2025-12-05
YN-0177: add expand/collapse functionality to the left pane in all tabs

### LINKED ISSUE #1641: YN-0177: add expand/collapse functionality to the left pane in all tabs

The left pane in many tabs (Overview, Products) is missing the same functionality of expanding or collapsing children available in the middle pane. 

### Additional context:

_This issue was automatically created from AYON ticket [YN-0177](https://support.ayon.app/projects/Active_Clients/overview)_

You can hold `option` when clicking the expand button to expand/collapse all it's children.

Image

The "left pane" they are talking about is the "slicer". Add this same behaviour to the `SimpleTable` component (slicer uses it).

Image

When simple table is "expandable":
- [ ] Add context menu items for expanding and collapsing all children
- [ ] Add key modifier `option` to clicking the expand button to expand/collapse all children.

filipvnencak:
## Description of changes 
This PR adds missing expand/collapse functionality to the left pane (Slicer) to match the behavior already available in the middle pane (e.g., in Overview or Products).
Users will now be able to expand/collapse items directly in the Slicer using:
	•	Context menu actions (Expand All / Collapse All)
	•	Key-modifier click on the expand toggle (⌥ Option key) to expand/collapse all children recursively

This brings feature parity between the Slicer and other tree-based views.

## Technical details

Implemented context menu support in Slicer using useCreateContextMenu.
Added new menu builder hook useSectionMenuItems for:
Expand All
Collapse All
(Placeholder) Delete action
Integrated context menu with SimpleTable via pt.row.onContextMenu.
Added recursive expand/collapse logic in handleExpand and handleCollapse.
Prepared structure for further Slicer actions.


Innders (changes requested):
The feature works but I think we could better architure the code to make it more reusable.

Expanding and collapsing all rows is not just a slicer specific action but is useful for all hierarchical tables.

Can this be built into the SimpleTable component so that all tables that use this component can have this functionality.

I think it could work the same way that it does in ProjectTreeTable context menu https://github.com/ynput/ayon-frontend/blob/8c5ca2a15245f3c88dcfe02d05d4922f53176f00/shared/src/containers/ProjectTreeTable/hooks/useCellContextMenu.ts. There are some built-in context menu items like these two and then you can add more custom ones yourself. 

The context menu items are also functions. This means that to component can provide useful data to the context menu inside of the component, making it far easier for context menu items to be written outside of the component.

Innders (changes requested):
- [x] `option+click` should be a modifier on the expand icon, not the whole row.
- [x] `option+click` on the middle table folders expand icon does not work now.

_How it currently works and how it should work on the simple table as well_

https://github.com/user-attachments/assets/cede85a9-ef6a-4493-a45f-c17d595ee75e

Innders:
Added in #2189

PR #1696, opened 2025-12-08
YN-0228: copying text breaks when cells selected

filipvnencak:
## Description of changes 
This PR significantly **improves the reliability** of the `copyToClipboard` functionality across various browsers and contexts.

The implementation now utilizes a **robust, multi-stage copying mechanism**:
1.  **Primary Attempt:** Use the modern, asynchronous **`navigator.clipboard.writeText`** API.
2.  **Automatic Fallback:** If the modern API fails (e.g., due to permission issues, secure context requirements, or focus problems), it automatically reverts to the proven **legacy method** using a hidden `<textarea>` and **`document.execCommand('copy')`**.

This ensures that clipboard operations are successful even in environments or scenarios where the modern API is restricted or unreliable.


Innders (changes requested):
This has not addressed the issue. Please follow discord link on original issue to see what is happening.

It is not that the copying function doesn't work but the cell copying function is blocking the copy shortcut of other parts of the app.

Innders (approved):

Issue #1693, opened 2025-12-05
YN-0287: Viewer Minimal Mode


ynbot:
### Please describe the feature you have in mind and explain what the current shortcomings are?

When opening the viewer using the spacebar there should be a new mode that hides all the extra UI and shows the video content as big as possible.

  

When selecting this mode it should be saved to localstorage. 

### Additional context:

_This issue was automatically created from AYON ticket [YN-0287]( https://support.ayon.app/projects/Active_Clients/overview?project=Active_Clients&type=folder&id=4749ffc0d13111f09ab3d973c9b949a8)

PR #1717, opened 2025-12-15
YN-0287: Viewer Minimal Mode

filipvnencak:
## Description of changes 

Implements a minimal/maximized mode toggle for the Viewer component, allowing users to switch between a streamlined viewing experience and a full-featured interface with all controls and panels visible.

## Technical details

 - Modified `src/containers/Viewer/Viewer.tsx`:
    - Added `useLocalStorage` hook to manage minimized state (`viewer-minimized-window`)
    - Implemented conditional rendering based on `minimizedWindow` state
    - Created button group with Details toggle and Close button
    - Wrapped conditional UI elements to only render in full mode

  - Modified `src/containers/Viewer/Viewer.styled.ts`:
    - Refactored grid layout to support both minimized and full states
    - Added `.minimized` class that switches container to flexbox column layout
    - Created `ButtonGroup` styled component for action buttons
    - Added `RightContainer` component for future layout flexibility
    - Adjusted positioning and alignment for responsive behavior

## Additional context

  This feature addresses issue #1693 (YN-0287) and provides users with flexibility in how they interact with the Viewer. The minimal mode is ideal for focus-driven review workflows where users primarily need to view media without distraction from UI controls.

### DIFF
--- a/src/components/Canvas.jsx
+++ b/src/components/Canvas.jsx
@@ -37,7 +37,7 @@ const Canvas = forwardRef(({style, onDraw, ...props}, ref) => {
 
     return () => resizeObserver.unobserve(parentElement);
 
-  }, [ref]);
+  }, [ref, onDraw]);
 
 
   return (
--- a/src/containers/VideoPlayer/Trackbar.jsx
+++ b/src/containers/VideoPlayer/Trackbar.jsx
@@ -206,13 +206,13 @@ const Trackbar = ({
     // ctx.moveTo(markInX, height - 1)
     // ctx.lineTo(markOutX, height - 1)
     // ctx.stroke()
-  }, [currentFrame, frameCount, markIn, markOut, isPlaying, highlighted])
+  }, [currentFrame, frameCount, markIn, markOut, isPlaying, highlighted, bufferedRanges, primaryColor, primaryContainer, onPrimaryContainer, containerLow, containerLowest])
 
   // Events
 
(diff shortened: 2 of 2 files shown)


Innders (changes requested):
Looking good, a couple of small things.

- [ ] Just noticed the tootips now go off the screen because of the tighter padding. Would it be possible for the tooltip component to detect this and flip to go under the button?

Image

- [x] Use correct icon `dock_to_left`.
- [x] The button should be highlighted when the details panel is open (opposite to now)
- [ ] Add a shortcut key for the state toggle `c`. This should also have a tooltip.

Image

- [ ] Title should use full path not just folder path. a.k.a

#### Title variants for each type
Folder - `{folder parent} / {folder_label}`
Task - `{folder parent} / {task_label}`
Version - `{folder parent} / {product_name} / {version}`

Innders (changes requested):
- [x] The button should be highlighted when the details panel is open (opposite to now)
- [x] Add a gap between the button and the tooltip when in bellow position.

Image

- [x] Sometimes when viewing a version through a product, the version number does not show in the title. (try using the spacebar shortcut on the products grid)

Innders (changes requested):
The folder name no longer shows twice but there is still no product name.



It should look like this:

Image

- [ ] Fix title to include `folder label (fallback to name) / product name / version`

PR #1718, opened 2025-12-15
YN-0305: Cells applying overrides when the values have not changed

filipvnencak:
## Description of changes 
 Fixed cell editing behavior in ProjectTreeTable to ensure data is saved correctly based on user interaction:

  **Behavior:**
  - **Enter key**: Always saves and moves to next row (even if value unchanged)
  - **Click outside / Blur**: Only saves if value actually changed
  - **Escape key**: Cancels edit without saving, restores original value
  - **Click on another cell**: Properly saves changes before switching focus

## Technical details

 1. **`ProjectTreeTable.tsx`** 
     - Added logic to find and blur the active input before switching cells
     - Uses `setTimeout(..., 0)` to ensure blur event completes before selection

  2. **`TextWidgetInput.tsx`** 
     - Removed "no change" check for Enter key (always triggers onChange)
     - Blur only saves if value changed

  3. **`EnumWidget.tsx`** 
     - Added value comparison for both single-select and multi-select
     - Only calls onChange if selection actually changed
     - Calls `onCancelEdit` if no change detected

  4. **`DateWidgetInput.tsx`** 
     - Enter: always saves
     - Click/Blur: checks if date changed before saving

  5. **`BooleanWidget.tsx`**
     - Added value change check 
     - Passes `'Click'` key for consistency

  6. **`CellWidget.tsx`** 
     - Refactored handleOnChange with explicit Enter/Click logic
     - Clearer separation of concerns

### DIFF
--- a/shared/src/containers/ProjectTreeTable/widgets/CellWidget.tsx
+++ b/shared/src/containers/ProjectTreeTable/widgets/CellWidget.tsx
@@ -134,7 +134,7 @@ export const CellWidget: FC<EditorCellProps> = ({
     if (key === 'Enter') {
       moveToNextRow()
       onChange?.(newValue, key)
-    } else if (key === 'Click' && newValue !== value) {
+    } else if (key === 'Click' && newValue != value) {
       onChange?.(newValue, key)
     }
   }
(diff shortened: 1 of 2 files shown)


Innders (approved):
Looks good, merging

PR #1728, opened 2025-12-22
Products: Details Panel Does Not Update Table

Innders:
## Description of changes 
When making changes to the version through the details panel, ensure the table data is updated as well.

image


Innders:
> When I look at the Overview page, the updates appear to be optimistic. In the Product page, however, the update happens only after a refetch triggered by invalidating tags.

For this issue, that's good enough, we can create a new issue to address that delay. This is a critical problem so I wanted to get a fix out ASAP.

PR #1739, opened 2026-01-12
YN-0345: Link Manager Colours

filipvnencak:
## Description of changes 

  Added entity-specific color support to the LinksManager widget. Folder and Task links now display colors from their respective type anatomies on both icons and labels, making it easier to visually identify entity types at a glance.

  Changes:
  - LinksManager icons and labels now use colors from folder/task/product type anatomies
  - Colors are applied via inline styles: style={{ color: link.color || undefined }}
  - Only affects the LinksManager widget display - the EntityPickerDialog remains unchanged (no colors in the selection dialog)

## Technical details

  Implementation:
  - Modified LinkManagerItem.tsx to apply link.color to both icon and label elements
  - The color data is already extracted in linksToTableData.ts via getEntityColor() helper
  - Colors default to undefined when not available, falling back to default styling

### DIFF
--- a/shared/src/api/generated/addons.ts
+++ b/shared/src/api/generated/addons.ts
@@ -672,8 +672,12 @@ export type AddonSettingsItemModel = {
   reason?: {
     [key: string]: string
   }
+  /** Indicates if the addon is part of the project bundle */
+  isProjectBundle?: boolean
 }
 export type AllSettingsResponseModel = {
+  /** If specified, indicates that the settings are for a project */
+  projectName?: string
   bundleName: string
   addons?: AddonSettingsItemModel[]
   /** If a project bundle is used, this field contains alist of addons that are inherited from the studio bundle */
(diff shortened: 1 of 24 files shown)


Innders (changes requested):
Because we want to ensure that the folder and task icons are consistent throughout create an `EntityIcon` component that resolves the colour of the icon based on entityType and subType. The component could probably hook into the project context so you don't need to prop drill the anatomy everywhere.

Missing from:

- [ ] Links Manager Pick dialog. The icon also needs to be added as well for tasks.

Image

- [ ] Type column enum (folder and tasks)

Image

- [ ] Entity card icons

Image

- [ ] Filter options (icon and text)

Image

- [ ] Task mentions

Image

- [ ] Slicer task type

Image

- [ ] Create new folder/task

Image

- [ ] Workfiles has lost the icon, maybe this is unrelated though?

Image

mkolar:
The only thing I'm hitting is that filtering task types applies the colour to icon and text. I'd only do it for the icon to keep the text legible
image

PR #1768, opened 2026-01-23
Project bundles

martastain:
This pull request introduces a major refactor and feature expansion for project bundle management, focusing on enabling "frozen" (per-project) bundles and improving related UI/UX in the Addon Settings. The changes add a new dialog for freezing/unfreezing bundles per project and variant, enhance the InfoMessage component, and update the AddonSettings container to support and display these new capabilities.


BigRoy (approved):
✅ Looking real good.

---

Small annoyances:

In case of `dev` bundle - can we show a different tooltip over the icon.

image

(my bundle here is named TEMP) but preferably it says "Can't freeze dev bundles"


When clicking freeze on a project production bundle whilst being in developer mode switches it directly to "Dev" mode just after. Can we stick that to "Production" to avoid confusion?  (same happens when unfreezing)


Trying to remove an override on a frozen bundle should perhaps not be shown as an option in the context menu - but admittedly not super important. Or better perhaps would be the error message being more descriptive as to why it's not allowed.

image
image

PR #1781, opened 2026-01-28
Products: Ctrl click does not work

filipvnencak:
## Description of changes 

Fixed cmd/ctrl multi-select not working in the Products ProjectTreeTable.    

When using cmd/ctrl + click to select multiple cells, any slight mouse movement during the click would clear the previous selections, making multi-select effectively broken.  

## Technical details

Added `isAdditiveSelectionRef` to track whether the selection was started in additive mode (cmd/ctrl held). This ref is set in `startSelection` and used in `extendSelection` to preserve the additive behavior throughout the drag/extend operation.

## Additional context

This bug only affected the Products page because the Overview page uses different data maps (`foldersMap`, `tasksMap`) that are properly memoized, resulting in fewer re-renders and less chance of triggering the `extendSelection` during clicks.

### DIFF
--- a/shared/src/containers/ProjectTreeTable/ProjectTreeTable.tsx
+++ b/shared/src/containers/ProjectTreeTable/ProjectTreeTable.tsx
@@ -1351,11 +1351,6 @@ const TD = ({
           return
         }
 
-        // if cell is already selected and it's not the row selection column, do nothing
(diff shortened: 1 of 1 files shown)


Innders (approved):
Works now thanks

PR #1798, opened 2026-01-31
DetailsDialog: Add simple JSON syntax highlighting + Copy to Clipboard button on hover

BigRoy:
## Description of changes 

Add syntax highlighting for the JSON data in the entity details panels.

Before:

image

After:

image

---

Every time I checked the Details panel I really just wanted to quickly investigate and read the data and often copy the contents elsewhere. This makes that much easier.

## Technical details

I'm sure the code is bad. 

What I looked for initially was whether we just had a Code component (or Markdown component) that'd come with this copy to clipboard button and allow to just specify a syntax, e.g. `<CodeBlock syntax="json">{rawJson}</CodeBlock>` but I couldn't find anything of the sort. We do have react markdown, but it doesn't seem to do any syntax highlighting for code blocks that specify syntax unfortunately. Because that would've made this a tiny PR.

For now I just 'vibe-coded' what I kind of expected the result to be just to learn a bit more about React - but it kept changing half the codebase so I just took the syntax highlighting snippet and copy-to-clipboard logic from it. I'm sure they are really bad.

We may want to decide on the colors of the syntax highlighting too - but admittedly, likely better to use a dedicated syntax highlighting library at that point?

### DIFF
--- a/shared/src/components/DetailsDialog/DetailsDialog.tsx
+++ b/shared/src/components/DetailsDialog/DetailsDialog.tsx
@@ -1,13 +1,9 @@
-import { useMemo, useState, useCallback } from 'react'
+import { useMemo } from 'react'
 import { toast } from 'react-toastify'
 import { useGetEntityQuery } from '@shared/api'
-import {Dialog, Icon} from '@ynput/ayon-react-components'
+import { Dialog, Icon } from '@ynput/ayon-react-components'
 import CodeEditor from '@uiw/react-textarea-code-editor'
-// TODO: Probably not import cross-component like this
-import {
-  BlockCode
-} from "@shared/containers/Feed/components/ActivityComment/ActivityComment.styled";
-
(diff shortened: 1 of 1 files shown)


Innders (changes requested):
Could you please extract this Syntax Highlighter to a separate component so that it can be reused in other places.

martastain:
What about CodeEditor we use in settings?

https://github.com/ynput/ayon-frontend/blob/develop/src/containers/SettingsEditor/Widgets/TextWidget.tsx#L23
https://github.com/ynput/ayon-frontend/blob/develop/src/containers/RawSettingsDialog/RawSettingsDialog.jsx#L163

BigRoy:
Sorry, just really want to try this once.

---

> Could you please extract this Syntax Highlighter to a separate component so that it can be reused in other places.

@copilot go be a good bot and make the Syntax Highlighter is a separate reusable component.

PR #1821, opened 2026-02-13
YN-0497: Task picker when uploading a version

filipvnencak:
## Description of changes 
  Adds a "Task" field to the version upload dialog so users can link a version to a task when uploading via the web. Previously, versions uploaded through the dialog could not be linked to tasks, making them easy to lose.
 - New "Task" row below "Version" with a "Pick linked task" button
  - Opens the entity picker dialog with the current folder preselected, showing all tasks within it
  - Selected task can be removed via a clear button
  - When uploading a new version (e.g. v003) for an existing product, the linked task from the previous version (v002) is auto-prefilled
## Technical details

 - Added `taskId` to the `GetLatestProductVersion` GraphQL query to support auto-prefill from previous version
  - Exposed `setTaskId` from `VersionUploadContext` so the form can update it
  - Used a `hasPrefilledTaskRef` to prevent re-prefilling after the user intentionally clears the task
  - Created `getTasks.ts` query file to properly export `useGetTaskQuery` following the existing codebase pattern

### DIFF
--- a/shared/src/components/VersionUploader/components/UploadVersionForm.tsx
+++ b/shared/src/components/VersionUploader/components/UploadVersionForm.tsx
@@ -72,33 +72,8 @@ const TaskFieldContainer = styled.div`
   width: 100%;
 `
 
-const TaskButton = styled.button`
-  display: flex;
-  align-items: center;
-  gap: var(--base-gap-small);
+const TaskButton = styled(Button)`
   flex: 1;
-  min-width: 0;
-  padding: 4px 8px;
-  cursor: pointer;
-  color: inherit;
-  font-size: inherit;
-  font-family: inherit;
-  background: none;
-  border: none;
-  text-align: left;
-
-  &:hover {
-    background: var(--md-sys-color-surface-container);
-    border-radius: var(--border-radius-m);
-  }
-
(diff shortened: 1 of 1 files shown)


shared/src/api/queries/tasks/getTasks.ts
```diff
@@ -0,0 +1,3 @@
+import { api } from '@shared/api/generated/tasks'
+
+export const { useGetTaskQuery } = api
```
Innders: You will want to enhance the getTask query before exporting the hook to at least provide some tags. This will ensure that other task mutations will keep this query updated with invalidateTags for tasks.

Innders (changes requested):
- [ ] Remove placeholder text and just have link button.

Image

- [ ] When linked the task itself should become the button to open the picker dialog.
- [x] Add folder path of the task
- [x] Add icon of the task

Image

- [ ] When opening the picker from a product or version the versions is not pre-filled
- [ ] When opening the picker from a product or version the folder is not pre-selected for the picker.

https://github.com/user-attachments/assets/7a0579f4-d19e-4b23-a92e-cdd8aab8e19c

Issue #860, opened 2024-10-08
URI: Refreshing homepage with task selected redirects to browser


Innders:
I understand why, but it's extremely annoying and confusing.

https://github.com/user-attachments/assets/d68f8198-6c46-4de3-8f1f-fbff78df7c98

PR #1832, opened 2026-02-19
Permissions: Dashboard all users switch for List all users permission

filipvnencak:
## Description of changes 

Show the Me/Users switch on the Dashboard tasks kanban for all users, not just managers/admins. The users dropdown is populated with project-scoped users, so each user only sees other users who have access to the same selected projects.  
           

## Technical details

  - Removed `isManager` permission gate from `DashboardTasksToolbar` — the switch now renders for all users
  - Removed the force-to-"me" filter logic that reset non-managers back to their own tasks
  - Replaced the global `useGetUsersQuery` (unfiltered `users(last: 2000)`) in `MeOrUserSwitch` with a `users` prop fed by the existing `useGetKanbanProjectUsersQuery` — this returns only users scoped to the selected projects, so no `list_all_users` permission is required
  - `MeOrUserSwitch` no longer fetches users internally; it receives them as a prop from the parent via `UserDashboardKanBan` → `DashboardTasksToolbar`
  
  
## Additional context

  - Previous attempt in  #1677  was closed due to bugs (switch appeared but didn't affect kanban tasks). This approach is simpler — always show the switch, use already-available project-scoped users
  - Regular users see only users from their assigned projects. Admins/managers continue to work as before
  - The `list_all_users` studio permission is not involved — it controls the global unfiltered user list, which we no longer use here

PR Beckend: [#860](https://github.com/ynput/ayon-backend/pull/860)

### DIFF
--- a/src/components/MeOrUserSwitch/MeOrUserSwitch.styled.js
+++ b/src/components/MeOrUserSwitch/MeOrUserSwitch.styled.js
@@ -31,6 +31,11 @@ export const MeOrUserSwitchContainer = styled.div`
     border-radius: var(--border-radius) 0 0 var(--border-radius);
     border: 1px solid var(--md-sys-color-outline-variant);
     border-right: none;
+
+    &:only-child {
+      border-radius: var(--border-radius);
+      border-right: 1px solid var(--md-sys-color-outline-variant);
+    }
   }
 
   .dropdown {


Innders (changes requested):
Users can now only see other users tasks in the kanban if they have 0 **read** restrictions (`read.enabled: false`).

- Always show the "me" part of the switch.
- Only show the other users part of the switch switch when every selected project has this `read.enabled: false` state for regular users.
- Enhance the "me" tooltip to explain why a user cannot access other users tasks. `My tasks. Remove read restrictions to see all tasks.`

Innders (changes requested):
I was unable to load the page because of this error.

__mf__virtual_host__….js?v=a3b6e7ab:9177 Uncaught TypeError: Cannot read properties of undefined (reading 'wing_it')
    at DashboardTasksToolbar.jsx:42:52
    at Array.every (<anonymous>)
    at DashboardTasksToolbar.jsx:41:33
    at DashboardTasksToolbar (DashboardTasksToolbar.jsx:35:29)

Issue #858, opened 2024-10-07
AY-6730_Handle products with no version


ynbot:
### Additional context:
link to discussion on [Discord](https://discord.com/channels/517362899170230292/1283418730205806632)
_(might be a private channel)_

_This issue was automatically created from Clickup ticket [AY-6730](https://app.clickup.com/t/86c09zqcj)_

PR #1833, opened 2026-02-19
Comments: Allow Admins to delete other users comments

filipvnencak:
## Description of changes 

Allow Admins to delete other users' comments for moderation purposes. Previously, only the comment author could delete their own comment. Edit remains owner-only. 

## Technical details

- `shared/src/containers/Feed/components/ActivityComment/ActivityComment.tsx` - Decoupled delete permission from edit permission. Delete now uses `canDelete = isOwner || isAdmin (via`useDetailsPanelContext()`), while edit stays gated on `isOwner` only.
## Additional context

Beckend PR: [#858](https://github.com/ynput/ayon-backend/pull/858)


Innders (changes requested):
Works for admins but the server blocks it for managers.

- [ ] Remove option for managers as they do not have that permission

(I pushed an update to shared package so it can be published and these changes can be used with the review addon)

Innders (approved):

PR #1839, opened 2026-02-23
YN-0515: Products: Double clicking on grid version card does not open panel

filipvnencak:
## Description of changes 

Grid view details panel does not open when double-clicking a version card (or using the context menu "Show details") while "Show Products" is toggled off.

## Technical details

The `gridMap.rowIdToIndex` (used by `SelectedRowsProvider` to validate selected rows) was only populated by `ProjectTreeTable` (table view). In grid view, `registerGrid` was never called, so version entity IDs were filtered out when deriving `selectedRows` — causing `showVersionDetails` to remain `false`.

  Fix: call `registerGrid` from `VPGrid` with the visible grid entity IDs so `SelectedRowsProvider` recognizes them.


Innders (approved):
Works thanks

PR #1844, opened 2026-02-25
Review: Integrate the remote ReviewSessionCards component on the Review list page

jakubfiala:
## Description of changes 

In the upcoming version of the Review addon, we introduce a new view on a particular Review session, which lets the user rearrange clips in the session as well as grouping them and organising into sections. The current table view is not suited to this, so this PR adds a remotely loaded ReviewSessionCards view directly from the addon. This view becomes the default and the table view can still be accessed using a switch.

### DIFF
--- a/shared/src/containers/ProjectTreeTable/context/SelectedRowsProvider.tsx
+++ b/shared/src/containers/ProjectTreeTable/context/SelectedRowsProvider.tsx
@@ -33,7 +33,10 @@ export const SelectedRowsProvider = ({ children }: SelectedRowsProviderProps) =>
             parseCellId(cellId)?.colId === ROW_SELECTION_COLUMN_ID && parseCellId(cellId)?.rowId,
         )
         .map((cellId) => parseCellId(cellId)?.rowId)
-        .filter((rowId): rowId is string => rowId !== undefined && gridMap.rowIdToIndex.has(rowId)),
+        .filter((rowId): rowId is string =>
+          rowId !== undefined
+          // in case the grid map hasn't been initialised, we ignore it
+          && (gridMap.rowIdToIndex.size === 0 || gridMap.rowIdToIndex.has(rowId))),
     [selectedCells, gridMap],
   )
 
--- a/src/pages/ProjectListsPage/ProjectListsPage.tsx
+++ b/src/pages/ProjectListsPage/ProjectListsPage.tsx
@@ -213,6 +213,7 @@ const ProjectLists: FC<ProjectListsProps> = ({
   const { listItemsData, deleteListItemAction } = useListItemsDataContext()
 
   const {
+    selectedCells,
     setSelectedCells,
     setFocusedCellId,
     setAnchorCell,
@@ -278,8 +279,13 @@ const ProjectLists: FC<ProjectListsProps> = ({
               onSelectionChange={(versionIds) => {
                 if (versionIds.length === 0) return clearSelection()
 
+                const areRowsSelected = Array.from(selectedCells)
+                  .some((cellId) => parseCellId(cellId)?.colId === ROW_SELECTION_COLUMN_ID)
+
+                const columnToSelect = areRowsSelected ? ROW_SELECTION_COLUMN_ID : "name"
+
                 const cellIds = versionIds
-                  .map((versionId) => getCellIdForColumn(listItemsData, versionId, "name"))
+                  .map((versionId) => getCellIdForColumn(listItemsData, versionId, columnToSelect))
                   .filter((id) => id !== null)
 
                 setSelectedCells(new Set(cellIds))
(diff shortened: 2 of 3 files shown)


Innders (changes requested):
## Toolbar

The toolbar can stay consistent across table and grid view, we want to add sections to table in the future anyway. For any functions that do not work in the table view they can be hidden.

- `Add clips` table and grid
- `Add section` grid (table in the future)
- `Filtering` table and grid
- `Actions` table and grid
- `Grouping` grid (table in the future)
- `Tab switch`
- `Customize` mostly for table view but grid will still have settings like `grid size`. Look at products page for that example.

Image

## Grid

The grid should dynamically fill the available width of the panel and be based on the user set size, look at the products page for how this should be done.

Multi Selection doesn't seem to work and is buggy. The details panel should work with the multi-selection along with dnd as well.

## DnD

Single selection of the item picks it up immediately, there should be a drag pixel threshold set for the card starts dragging. Look at the kanban page for an example.

I would also hide the card item that is showing where the card will be dropped. Currently it is low opacity but it should just be hidden, leaving an empty gap.

## Sections

It's very difficult to drag items between sections. I would also add a drop hover effect to the whole section, rather than just the header.

When editing a section name I would remove the checkmark icon button and just use Enter key shortcut. Look at how it is done in Planner addon scenarios.

When grouping by an attribute that is not section, you can still edit the name.

Use the icon `chevron_right` for the header dropdown arrow.

When adding a new section, create the new section and then auto focus the name to change it. The dialog feels kind of clunky.

It took me a while to figure out how to delete the section. Maybe hovering over the section header could show a delete button on the right?

### Tasks by priority

## Toolbar
- [x] **(Critical)** Make the toolbar consistent across both table and grid views.
- [x] **(Critical)** Auto-hide any toolbar functions that are currently incompatible with the table view.
- [x] **(Critical)** Ensure `Actions` work for both table and grid.
- [ ] **(High)** Ensure `Filtering` works for both table and grid.
- [ ] **(High)** Implement `Customize` menu (tailor mostly for table view, but include `grid size` for grid view—reference the products page).
- [ ] **(Medium)** Ensure `Add clips` works for both table and grid.

## Grid View
- [x] **(High)** Update grid to dynamically fill the available panel width based on the user-set size (reference the products page).
- [ ] **(High)** Fix multi-selection bugs so it functions reliably.
- [ ] **(Medium)** Connect multi-selection to the details panel so it updates correctly when multiple items are selected.
- [ ] **(Medium)** Ensure Drag and Drop (DnD) works seamlessly with multi-selected items.

## Drag and Drop (DnD) Mechanics
- [ ] **(High)** Implement a drag pixel threshold so single-selected items don't pick up immediately on a regular click (reference the kanban page).
- [ ] **(Medium)** Update the drag placeholder: entirely hide the original card item where the card will be dropped (leaving an empty gap), rather than setting it to low opacity.

## Sections
- [ ] **(Critical)** Improve the sensitivity/hitboxes to make it easier to drag items between sections.
- [ ] **(Critical)** Expand the drop hover effect to cover the entire section, rather than just the header.
- [ ] **(High)** Disable section name editing when the view is grouped by an attribute other than "section".
- [ ] **(Medium)** Replace the header dropdown arrow icon with `chevron_right`.
- [ ] **(Medium)** Add a section delete button that appears on the right side of the section header only on hover.
- [ ] **(Medium)** Update section name editing: remove the checkmark icon button and bind save to the `Enter` key (reference Planner addon scenarios).
- [ ] **(Low)** Revamp the "Add Section" UX: remove the dialog prompt, immediately create the new section in the UI, and auto-focus the name field for quick typing.

Innders (changes requested):
- [ ] Ensure the details panel can be closed and the selection of the cards is cleared. [LOW]

Image

## Dragging
- [ ] A little dot appears top left when dragging the card. [LOW]
- [x] When dragging the card it should rotate slightly. Look at kanban page, it should match how it looks there.
- [x] When dragging the card use the dragging cursor.

## Selection
- [x] You should be able to select multiple versions using ctrl/cmd click.
- [x] You should be able to drag and move multiple versions at once. Look at kanban how it stacks. [LOW PRIORITY]

## Context menu
- [x] "Remove" rename to "Remove from list" with "remove" icon to match table context.
- [ ] Add "Show details" context item to match table view. [LOW]
- [x] Add "Open in viewer" to match table view.
- [ ] Add "Add to section" that lets you pick a section to move the versions to [LOW]

## Player
- [x] Add support for showing the ViewerPlayer so you can quickly preview the version. This should work with the context menu item and the spacebar shortcut. [CRITICAL PRIORITY]

## Grouping
- [x] When grouping is active the height is more than 32px, shifting the height of the toolbar. It should not shift height and stay at 32px.

Image

- [x] When grouping by emun that has an icon, could we show this icon + color in the group header. [LOW]

Image
 
- [x] Move the counter to the top so that is in line with the expand all button.

Image

Innders (changes requested):
## Critical

- [x] When the wrong addon version is installed, the grid tab shows nothing. Use `EmptyPlaceholder` to tell the user they need x version of the addon for the grid.

- [x] The details panel should open when double clicking the item, it should work exactly the same as the products page.

PR #1851, opened 2026-03-02
Bundles: Show warning higher version icon for server bundles

filipvnencak:
## Description of changes 

Show the "i" info icon (latest version indicator) next to the version dropdown selector in bundles. Previously, the icon was only visible in read-only mode for pipeline addons and would disappear when the version dropdown was shown. Now the icon appears alongside the dropdown, so users can still see when a higher version of an addon is available while editing.

## Technical details

- Added `latestVersion` prop to `AddonListItem` component
- Renders `LatestIcon` next to `VersionSelect` when the selected version is not the latest installed version
- Reuses the same `LatestIcon` component and tooltip ("Latest installed version: X") as the read-only `AddonItem` for consistency

### DIFF
--- a/src/pages/SettingsPage/Bundles/BundlesAddonList.tsx
+++ b/src/pages/SettingsPage/Bundles/BundlesAddonList.tsx
@@ -115,14 +115,12 @@ const AddonListItem: React.FC<{
         placeholder="NONE"
         onChange={(e: string[]) => setVersion(e[0] || null)}
       />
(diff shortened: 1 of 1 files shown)


Innders (changes requested):
The icon should still take up space even when not showing so that the dropdowns are always the same width.

- [ ] Ensure dropdowns stay the same width regardless of icon showing or not.

Image

PR #1856, opened 2026-03-03
 feat: add defaultData prop to AttributeEditor for duplicate support

filipvnencak:
## Description of changes 
Add `defaultData` prop to `AttributeEditor` component, allowing consumers to pre-fill the form when creating a new attribute. This enables duplicate functionality — the editor opens in "new" mode (with name auto-generation) but with existing attribute data pre-populated.

## Technical details

 - New optional `defaultData?: Partial<AttributeForm>` prop on `AttributeEditorProps`
  - When `attribute` is `null` (new mode) and `defaultData` is provided, the form initializes with merged default values
  - Name is auto-generated via `camelCase(defaultData.data.title)` after `buildInitFormData` returns, since the merge logic skips excluded keys like `name`

## Additional context

 Needed by powerpack's `ListsAttributesSettings` to support duplicating custom list attributes.
 
 
[#63](https://github.com/ynput/ayon-power-pack/pull/63)

### DIFF
--- a/shared/package.json
+++ b/shared/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@ynput/ayon-frontend-shared",
-  "version": "0.3.15",
+  "version": "0.3.16",
   "description": "Shared React components for AYON frontend",
   "main": "dist/index.cjs.js",
   "module": "dist/index.es.js",


shared/src/components/AttributeEditor/AttributeEditor.tsx
```diff
@@ -186,7 +188,16 @@ export const AttributeEditor: FC<AttributeEditorProps> = ({
   onEdit,
   onDelete,
 }) => {
-  const initForm = buildInitFormData(excludes, { position: existingNames.length })
+  const initForm = buildInitFormData(excludes, {
+    position: existingNames.length,
+    ...defaultData,
+  })
+  // Set name after buildInitFormData since it skips excluded keys during merge
```
Innders: This a bit of a hack, it would be better to resolve this properly inside of buildInitFormData

PR #1865, opened 2026-03-11
YN-0114: Create multiple links of the same entity

filipvnencak:
## Description of changes 

 Adds support for creating and managing multiple links of the same entity. Instead of displaying duplicate links individually, they are now grouped into a single visual item with a count badge (e.g. `x5`).

- **Grouped display**: Duplicate links to the same entity are collapsed into one row with an `x{count}` badge in both the LinksManager dialog and the table cell chips (LinksWidget).
- **Inline count editing**: Clicking the count badge opens an inline number input. Changing the value automatically creates or removes the appropriate number of links
  to match the desired count.
- **Optimistic UI**: Count changes reflect immediately in the UI while the actual link operations happen in the background.
- **Bulk remove**: The unlink button on a grouped link removes all links in the group at once, with an updated tooltip (`Remove all 5 links`).
- **Restricted links**: Links the user can't access are never grouped (since we can't verify they point to the same entity).

## Technical details

- New `groupLinksByEntity()` utility groups links by `entityId` preserving original sort order. Restricted links are excluded from grouping (keyed by `linkId` instead).
- Count changes work by diffing: increasing the count adds N new links via `linksUpdater.add()`, decreasing removes N links from the end of the group via `linksUpdater.remove()`.
- Optimistic count state (`optimisticCounts`) provides instant feedback — it auto-clears once the real data from the cache matches.
- `CellEditingDialog` now blurs the active element before closing so that in-progress count edits commit their value.

## Additional context

Screenshot 2026-03-11 at 08 26 37

links_count

### DIFF
--- a/shared/src/components/LinksManager/LinkManagerItem.tsx
+++ b/shared/src/components/LinksManager/LinkManagerItem.tsx
@@ -103,12 +103,13 @@ export const LinkManagerItem: FC<LinkManagerItemProps> = ({
             onBlur={commitCount}
             onKeyDown={handleKeyDown}
             onClick={(e) => e.stopPropagation()}
+            data-tooltip="Number of links to this entity."
           />
         ) : (
           <Styled.CountBadge
             onClick={handleBadgeClick}
             tabIndex={0}
-            data-tooltip="Click to edit count"
+            data-tooltip="Number of links to this entity."
           >
             x{count}
           </Styled.CountBadge>


Innders (changes requested):
- [x] The opening of the count input is different height. The height of the count button and input should be exactly the same. The span button height should be a bit bigger and the input smaller so they match. Ensure these changes do not affect the height of the overall link item.
- [x] Remove hover effect when used in the chip

https://github.com/user-attachments/assets/7a379dee-5148-421a-8dfa-ab2a49301b72

Innders (approved):
Works great thanks

PR #1890, opened 2026-03-25
YN-0559: Merge "show hierarchy" with group by and add folder grouping

### LINKED ISSUE #1868: YN-0559: Show flat lists of shots or assets in the overview

 

We need to add a mode to the overview page that is in-between full hierarchy and `tasks` only. 

  

When a producer needs to see only "shots" for example without the sequences, it is not really possible at the moment. 

  

The best approach might actually be what we've done in the new review page and the scheduler, where this is just one of the options in the grouping settings. 

  

With that we could promote the "Group by" button to the top menu and give it more options and expand down the line.

  

Starting with these:

- Hierarchy
- Folder
- Status
- Assignee
- Task Type
- Task Attributes (Dropdown with all attributes)

### Additional context:

_This issue was automatically created from AYON ticket [YN-0559]( https://support.ayon.app/projects/Active_Clients/overview?project=Active_Clients&type=folder&id=0383ac901bce11f19e3b21c0f875334b)

filipvnencak:
<!-- PR TODO: -->
<!-- 1: Set assignee to you -->
<!-- 2: Set reviewer to: Luke Inderwick or Martin Wacker -->
<!-- 3: Set label -->
<!-- 4: Automations will set any required projects -->

## Description of changes 
 Replace the binary "Show hierarchy" toggle on the Overview page with a SortingDropdown that lets users group tasks by Hierarchy, Folder (type), Status, Assignee, Task Type, Tags, or any groupable attribute. All non-hierarchy options render flat task lists in collapsible group sections. Default grouping is "Folder".

  - Add top-level "Group by" dropdown in the Overview toolbar synced bidirectionally with the Customize panel's group-by setting
  - Fix folderType grouping pipeline: correct entity type for group queries (folder), GraphQL fragment (folder { folderType }), client-side entity matching, and filter routing (folder-level conditions go to folderFilter, not task filter)
  - Add "Hierarchy" and "Folder" options to the Customize panel's group-by list


## Technical details
<!-- Please state any technical details such as limitations -->

- viewGroupBy is stored in both localStorage (for fast local reads) and server view settings. When the user changes the dropdown, updateGroupBy syncs the selection to the server via onUpdateColumns. The localStorage copy is used to derive showHierarchy (viewGroupBy === null) without waiting for the server round-trip.        

- Bidirectional sync between the dropdown and Customize panel uses a useRef to track the previous panel value, preventing circular update loops.
- folderType grouping requires entityType: 'folder' for the groups API query (backend doesn't support folderType as a task grouping key). The group filter conditions containing folderType are separated from task-level conditions and routed to folderFilter instead of the task filter parameter.                                                                                          
- Default grouping is folderType for new users. Existing users who had hierarchy enabled will see their persisted localStorage value.                                                                                                                                                                                                                                                        
- Grouped expand loading has no per-group loading indicator — all expanded groups refetch together via forceRefetch: true. This is a known limitation, not addressed in this PR.

## Additional context
<!-- Add any other context or screenshots here. -->
Also fixed UI glitch in ARC [#223](https://github.com/ynput/ayon-react-components/pull/223)


Innders (changes requested):
Overall working really well 👍 

- [x] Clicking none in side panel does not work.
- [x] Side panel is missing `Folder` option.

Image

- [x] Group by dropdown should use the `x` and `v` actions to remove the grouping and change grouping direction. Look at homepage group by dropdown.

Image

Innders (changes requested):
- [x] Grouping by "None" does not work, making it impossible to see a flat list of only tasks. "Hierarchy" is not no grouping, it is grouping by Hierarchy.
- [x] Removing grouping from the dropdown defaults to "hierarchy". It should default to nothing, we are not grouping by anything.

Innders (changes requested):
- [x] Hierarchy chip doesn't have option to remove (`x` button)

Image

- [x] Ascending/Descending should always be visible in the settings, even when not grouping. When not grouping it's disabled.
- [x] Ascending/Descending should be disabled for hierarchy and folder

Image

- [x] When grouping by folder do not use the same table UI as with normal groups. It should use the same table layout as Hierarchy. This means we do not force the Folder/Task column to be pinned and the column should be sortable.

Image

- [x] Switching to group by product in setting panel is super glitchy.

https://github.com/user-attachments/assets/ca490607-62d0-484f-b9e4-8ee458c889a7

Innders (changes requested):
Working nicely now!

- [x] Toggle desc on the chip for products page doesn't work.

Image

Innders (changes requested):
- [x] sort groups value seems to make no difference to the sorting of the groups. (products page)

Image

Innders (changes requested):
It's working way smoother now and functionally everything is correct.

A couple of cosmetic things

- [x] Can the dropdown of the group fields include the icons?

Image

- [x] Use the same arrow icon from the dropdown for the settings panel.

Image



- [x] Ensure Ascending/Descending row persists even when not grouping but with disabled state. (Do not hide it, just disabled it).

Image

Image

Innders (approved):
Works and looks great! Thanks!

PR #1904, opened 2026-03-31
Attributes: Delete attribute at top of page does nothing

BigRoy:
## Description of changes 

Fix Delete Attributes button on the top of the Attributes tab doing nothing at all

Note: The context menu "Delete" action was working fine

## Technical details

- `onClick={onDelete}` passes the React synthetic click event as the argument instead of `selectedAttribute`. The `onDelete` guard if `(!selected?.name)` catches it silently and returns. The context menu path worked fine because it explicitly calls `() => onDelete(selected)`.

## Additional context

Found this out here: https://github.com/ynput/ayon-backend/pull/898#issuecomment-4165360371

### DIFF
--- a/src/pages/SettingsPage/Attributes.jsx
+++ b/src/pages/SettingsPage/Attributes.jsx
@@ -158,7 +158,7 @@ const Attributes = () => {
               label="Delete attribute"
               icon="delete"
               disabled={selectedAttribute?.builtin}
-              onClick={onDelete}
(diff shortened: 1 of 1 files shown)


Innders (approved):
Didn't test but looks good and make sense. Assuming you tested it works.

BigRoy:
> Didn't test but looks good and make sense. Assuming you tested it works.

Yup, worked for me.

Issue #905, opened 2024-10-28
Reset the input name when we adding shots


krishnaavril:
## Story
When we click on "Add" the entered value should get reset, so its easy for entering the shots numbers manually
![image](https://github.com/user-attachments/assets/5d24dc74-e2ad-4f13-b204-2d9cd8427821)

### To reproduce

1. On any project, create shots
2. when wanted to enter multiple numbers, a cleared number will be handy

## Environment

- **Server version**: 
- **Server host OS**: 
- **Browser**: 
- **Client version**:

Innders:
I understand the idea but I think this specific use case doesn't fit for everyone.

At the moment the name input focusses and selects all the text allowing you to quickly type the next name.

If you are looking to create many folders where the number at the end changes then the "Add folder sequence" tool might be a better fit.

Nevertheless thanks for the feedback! We may review this again in the future if we get further feedback 😄

PR #1922, opened 2026-04-10
Deleting Entities: Improve message of what will be deleted

### LINKED ISSUE #1882: Deleting Entities: Improve message of what will be deleted
## Story

Deleting entities is a dangerous operation - and the user should be clearly informed about the results of their operations. Otherwise, it's easy to overlook something that may also get deleted that you did not anticipate.

### Problems

Take for example these cases:

Image

Image

The second one will have a massive impact, deleting essentially ALL assets but the delete prompt does not highlight this massive distinction because both say "will delete 2 entities".

### Proposal

A summary of what will be deleted will help a lot:
```
This operation will delete:
- 2 Folders
- 2 Tasks
- 20 products
- 80 versions
```
But it'd be great if the dialog could perhaps have a small scrollable section highlighting all the entities that will be deleted - just to confirm?
```
/path/to/x/y/z    - FOLDER  -  35 children entities
/path/to/a/b/c    - TASK    -  3 children entities
```
Hovering the children entities could perhaps tell me more accurately what they are, e.g. 2 tasks, 3 products, 30 versions.

Possibly if I have a multiselection like the second image above that it condenses it to the root one because the other one is a child anyway and will be removed with it regardless. That way it'll be very clear you're deleting that entity and everything underneath 

## TODO

- [ ] ...

filipvnencak:
## Description of changes 
Improves the delete confirmation dialog in ProjectTreeTable to show a detailed breakdown of what will be deleted, instead of a generic "X entities" message.

## Technical details

- Counts selected entities by type (folder, task, product, version) and builds a human-readable summary (e.g. "2 folders, 1 task")
- For folders with children, appends a detailed list showing which folders contain child folders or tasks
- Uses `React.createElement` to render a structured message (paragraph + scrollable list) inside the confirmation dialog
- Updated `confirmDelete` message type from `string` to `string | React.ReactNode` to support rich content

## Additional context

Single entity deletions still show the entity name in quotes (e.g. `"MyFolder"`). The children details list has a max-height of 200px with scroll for cases with
  many nested items.

 Beckend [#905](https://github.com/ynput/ayon-backend/pull/905)


BigRoy (commented):
Pretty good it seems:

Image

Image

(Can we get this to display the amount of child folders?)


When selecting nested entries:

Image

It'd be great if it could collate to the 'top level' of what it'd remove, so the `substance` entry is unrelated here because it's part of a parent being deleted? however, it must be much clearer from the parent selection how big the size of children being removed will be?

Preferably the folder children entry would read:

- `film` contains `4` child folders with `123` tasks

(also that child number should be the 'recursive children' and should take into account the amount of children WITHOUT the current view filter applied (not sure if that's relevant with how you'd be querying that) because deleting a parent would delete children regardless of applied filter - so what may look like an empty folder due to filters may actually be deleting lots of stuff)

filipvnencak:
@BigRoy what do you think now 😃?
Screenshot 2026-04-13 at 14 28 22

shared/src/containers/ProjectTreeTable/hooks/useDeleteEntities.ts
Innders, line 1: What is this? If there is jsx in it then it needs to be a `.tsx` file.

Innders (changes requested):

shared/src/api/generated/graphql.ts
Innders, line 1: Using wrong old graphql generation. Merge develop to get latest generation

shared/src/containers/ProjectTreeTable/components/DeleteConfirmContent.tsx
```diff
@@ -0,0 +1,137 @@
...
+
+export const DeleteConfirmContent = ({
+  entityLabel,
+  topLevelFolders,
+  fetchDeleteInfo,
+}: DeleteConfirmContentProps) => {
+  const [loading, setLoading] = useState(topLevelFolders.length > 0)
+  const [childrenDetails, setChildrenDetails] = useState<string[]>([])
+
+  useEffect(() => {
```
Innders: If the data has already been loaded outside of the component then this useEffect is not needed.

shared/src/containers/ProjectTreeTable/hooks/useDeleteEntities.tsx
```diff
@@ -0,0 +1,173 @@
...
+            })
+          }
+        }
+      }
+    } catch (error) {
+      console.warn('Failed to fetch folder delete info, falling back to local data', error)
+    }
+
+    return map
+  }
```
Innders: If you use rtk query hooks then this fetching logic is not required. If you need to transform that data do that with .enhanceEndpoints in a services file like normal.

Innders (changes requested):
This is not using the best practices of RTK query, please look to do it properly using rtk query hooks.

Innders (approved):
Looks better now thanks

PR #1926, opened 2026-04-13
Add a project-wide search for folders, tasks or products

BigRoy:
## Description of changes 

When in a project add a Project wide search that is a quick search for a folder, task or product.

## Technical details

⚠️ This is currently just a UX draft to discuss the UI - the implementation will surely need heavy cleanup.

- [ ] It seems quite finnicky to get the right Product selected on the Products page automatically when choosing it from the search.  We seem to be lacking an easy way to ensure a page will display the entity if we redirect the user to it. So that we can enforce e.g. "highlight this item"

## Additional context

https://github.com/user-attachments/assets/525b9276-a135-4fa9-b54a-09b2e5d2c476

image

image

image

### DIFF
--- a/shared/src/api/generated/graphqlLinks.ts
+++ b/shared/src/api/generated/graphqlLinks.ts
@@ -1575,7 +1575,7 @@ export type GetFolderLinkDataQueryVariables = Exact<{
 }>;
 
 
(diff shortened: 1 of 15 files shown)


BigRoy:
Closing this since it's merely presenting a UX idea but needs more discussion implementation side to be worth pursuing further.

PR #1931, opened 2026-04-14
Settings: Support multi-selection with manual re-ordering

### LINKED ISSUE #1730: AY-7981: preferred apps

 

We need different widget for list[str] settings model where it is possible to change order of strings.

### Additional context:

_This issue was automatically created from AYON ticket [AY-7981]( https://support.ayon.app/projects/Active_Clients/overview?project=Active_Clients&type=folder&id=34cef4917ea011f083fb3003c82d0673)

filipvnencak:
## Description of changes 
Add sortable multiselect widget for preferred apps (project applications profiles). Users can reorder application priority via drag-and-drop using a new OrderedListWidget component integrated into the settings editor. 

## Technical details

- New OrderedListWidget component using @dnd-kit for drag-and-drop reordering with dual-panel UI (selected + available items)
- SelectWidget detects sortable multiselect fields via widget="sortable_multiselect" schema prop or ID pattern match (applications_profiles_*_applications)
- Uses isEqual (order-sensitive) instead of equiv (order-insensitive) for change detection so reordering is correctly tracked and persisted

## Additional context

Screenshot 2026-04-15 at 10 36 57
Screenshot 2026-04-15 at 10 31 38

### DIFF
--- a/src/containers/SettingsEditor/Widgets/OrderedListWidget.tsx
+++ b/src/containers/SettingsEditor/Widgets/OrderedListWidget.tsx
@@ -111,6 +111,33 @@ const SectionLabel = styled.div`
   letter-spacing: 0.5px;
 `
 
+const ButtonGroup = styled.div`
+  display: flex;
+  gap: 6px;
+  padding: 6px 0;
+`
+
+const ActionButton = styled.button`
+  padding: 4px 10px;
+  font-size: 12px;
+  border: 1px solid var(--md-sys-color-outline-variant);
+  border-radius: var(--border-radius-s);
+  background: var(--md-sys-color-surface-container-high);
+  color: var(--md-sys-color-on-surface);
+  cursor: pointer;
+  transition: all 0.2s;
+
+  &:hover:not(:disabled) {
+    background: var(--md-sys-color-surface-container-highest);
+    border-color: var(--md-sys-color-outline);
+  }
+
+  &:disabled {
+    opacity: 0.5;
+    cursor: not-allowed;
+  }
+`
+
 // Sortable item for selected list
 const SortableItem = ({
   id,
@@ -172,6 +199,15 @@ const OrderedListWidget = ({ value, options, onChange }: OrderedListWidgetProps)
 
   const removeItem = (item: string) => {
     onChange(value.filter((v) => v !== item))
+  }
+
+  const selectAll = () => {
+    const allValues = options.map((opt) => opt.value)
+    onChange(allValues)
+  }
+
+  const deselectAll = () => {
+    onChange([])
   }
 
   // Close dropdown on outside click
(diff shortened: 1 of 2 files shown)


Innders (changes requested):
I think we could make this simpler.

For the search dropdown component just use a regular multi-select dropdown. All of the extras are built in. When selecting a value keep it in the dropdown so we can see it selected. Then underneath we have the re-ordering list.

So this is basically what we used to have with the previous dropdown but just with the extra re-ordering underneath it. Then this because a "MultiSelectReordering" widget.

1. Bring back the previous dropdown that was used.
2. Keep the the re-ordering UI underneath.
3. Rename "Selected" to "Order selection" and remove the select/deselect all buttons.

Image

## Actions

Order set in the settings doesn't match order in the actions dropdown?

image
image


Ensure that these applications are sorted above any other applications.

image

filipvnencak:
@Luke the ordering is beckend issue, we have talked with Martin and Kuba about that, they need to fix that, now beckend returning list randomly, here is beckend issue https://github.com/ynput/ayon-backend/issues/917

Innders:
Thanks for the issue creating and link @filipvnencak. Could we please look at the UI feedback?

martastain:
I think the widget should be named just `sortable`. The fact it is "multiselect" is already infered from the datatype and presence of an enum.

iLLiCiTiT:
It is an issue of applications addon, not of a backend. This PR looks like works, if [this](https://github.com/ynput/ayon-frontend/pull/1931#issuecomment-4286536821) is resolved and it really works, then it can be merged.

NOTE: We also have to start showing dynamic actions in frontend to be able to change applications addon, this PR has to be already merged and released when we start to do it in applications addon.

iLLiCiTiT:
PR that changes the applications addon https://github.com/ynput/ayon-applications/pull/145 (you won't see them in frontend because dynamic actions are not showed there).

src/containers/SettingsEditor/Widgets/OrderedListWidget.tsx
```diff
@@ -0,0 +1,205 @@
...
+
+  return (
+    <Container data-tooltip="">
+      <Dropdown
+        widthExpand
+        options={options}
+        value={value}
+        onSelectionChange={handleSelectionChange}
+        multiSelect
+        placeholder="Select applications"
```
Innders: Missing lots props that would normally be on select dropdown, Like `onSelectAll`.

Innders (changes requested):
Almost there, just a couple things to clean up. Also I pushed a small change to hide the item when ordering so that only the overlay shows.

- [ ] Set a max width for select and ordered list dropdowns so that they don't overflow like this.

Image

Innders (approved):
Works well now thanks

Innders (approved):

Issue #1913, opened 2026-04-08
YN-0642: Filters: Allow filtering by date range on products page


ynbot:
### Please describe the feature you have in mind and explain what the current shortcomings are?

Allow filtering by a date range (start-end) in the products page, so that you can e.g. list all products that were updated between certain dates.

<br>

**Overview**

<br>

This is currently implemented on the overview page but:

1.  Seems to be buggy and doesn't always work.
1.  Only has start and end dates. createdAt and updatedAt fields need to be added.
1. Does not support custom date range.

**Products**

<br>

Needs to support date filtering on versions like the overview page.

1.  createdAt and updatedAt fields + any custom date attributes.
1.  Same options are overview Today, Yesterday etc + newly added custom date range.

**Notes**

<br>

- The dates look quote bad when added to the filter bar, format them nicely.
- Remember this is a powerpack only feature. You might need to make changes in the powerpack. 

### Additional context:

_This issue was automatically created from AYON ticket [YN-0642]( https://support.ayon.app/projects/Active_Clients/overview?project=Active_Clients&type=folder&id=ccfea1e0327811f19ff44548948c24f9)

PR #1932, opened 2026-04-15
YN-0648: Search for multiple folders at the same time

### LINKED ISSUE #1930: YN-0648: Search for multiple folders at the same time

Can we copy and paste shots into the search bar and find only those shots. 

Pasting a list of strings into the search field in Overview and products should break it up using a "," delimiter and create them as a list of separate search terms.

### Additional context:

_This issue was automatically created from AYON ticket [YN-0648]( https://support.ayon.app/projects/Active_Clients/overview?project=Active_Clients&type=folder&id=c3e4b3d0329411f1916bcf676d7f83ef)

filipvnencak:
## Description of changes 
**Search chip edit prefill:** Clicking an existing global search chip pre-fills the dropdown search input with the chip's current value for quick editing, hiding redundant preselected rows

## Technical details

- Search auto-fill intercepts `pt.dropdown.pt.item.onClick` and returns `false` to prevent SearchFilter from reopening the values panel after creating the filter
- Search chip prefill uses the native `HTMLInputElement.value` setter + dispatched `input` event to programmatically update the controlled React input inside SearchFilterDropdown
- `buildFilterId` is imported from `ayon-react-components` to generate unique filter IDs when creating filters from the wrapper

## Additional context

Relates to GitHub issue #1930  (search text auto-fill) and #1913  (date range filtering on products page).


Innders (changes requested):
Please provide demo of using the new feature, I'm not sure how it's supposed to work. I've tried many ways to search for multiple folders but none of them work.

Innders:
@filipvnencak could we take a look at this please with high priority

filipvnencak:
https://github.com/user-attachments/assets/ed0d253c-e584-47a1-891c-29810ebe02b0

Innders (changes requested):
I cannot find a way to search for multiple things at once.

I have tried: 
- `010_0010 010_0020`
- `010_0010, 010_0020` (with commas specified in the issue)
- Adding them one by one

The video above doesn't show searching for multiple folders either.

Innders (changes requested):
- [x] Editing the string whilst adding commas does not work `shot010 shot020` -> `shot010, shot020`

shared/src/containers/ProjectTreeTable/utils/expandRelativeDates.ts
```diff
@@ -1,4 +1,4 @@
-import { QueryFilter, QueryCondition } from '@shared/containers'
```
Innders, line 1: related change?

shared/src/api/queries/overview/getOverview.ts
```diff
@@ -317,6 +317,34 @@ const injectedApi = enhancedApi.injectEndpoints({
         }
       },
     }),
+    // Derive folder IDs from tasks matching a search string. Reuses the existing
+    // GetTasksList query — GraphQL tasks resolver splits search on commas for OR
+    // semantics, unlike the REST searchFolders endpoint which ANDs all terms.
+    getFolderIdsByTaskSearch: build.query<
```
Innders: The search folders query has been updated to work with commas so I don't think this is needed now?

https://github.com/ynput/ayon-backend/pull/923

Innders (commented):
It works but the code needs to be looked at, especially for the data fetching. Using `first: 10000` on tasks to get folders is not good practice.

Innders (changes requested):
Changes above

PR #1946, opened 2026-04-21
Overview: Group by folder and hierarchy broken

### LINKED ISSUE #1942: Overview: Group by folder and hierarchy broken
It's impossible to set the grouping to `folder` or `hierarchy`.

https://github.com/user-attachments/assets/4e5f34a1-360b-4dea-abcb-5a0711d861bc

filipvnencak:
## Description of changes 
Fixes the Overview page grouping bug where selecting a Group By option (Status, Folder, etc.) either failed to apply or flickered back to the previous mode. Also removes the "Hierarchy" flicker during initial view-settings load.

## Technical details

- viewUpdateHelper.ts is shared by all view types (Overview, Lists, TaskProgress, VersionsProducts). Safe fallback: if cache-read returns undefined, we use the old closure path — same behavior as before.
- The fix assumes RTK's onQueryStarted runs synchronously before updateView(...) returns. Standard RTK behavior.                                                    
- The "cache is fresh this tick" flag is module-level so two different useViewUpdateHelper instances (data provider + context) can see each other's writes in the same tick. A per-instance ref wouldn't work.                                                                                                                        
- store.getState() is cast to any — RootState lives in the app, not @shared. Selector only reads the api slice, which every app has.                                
- Type: viewGroupBy: string | null → string | null | undefined.

### DIFF
--- a/shared/src/containers/ProjectTreeTable/context/ProjectTableProvider.tsx
+++ b/shared/src/containers/ProjectTreeTable/context/ProjectTableProvider.tsx
@@ -212,8 +212,10 @@ export const ProjectTableProvider = ({
     isLoadingMore,
   })
 
-  // overrideGroupBy (from view dropdown) takes priority over Customize panel's groupBy
-  const effectiveGroupBy = overrideGroupBy || columnSettingsGroupBy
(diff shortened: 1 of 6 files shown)


Innders:
Seems to work for me if this is ready for review

PR #1955, opened 2026-04-28
Views: Changing view permissions does not move view to correct section

### LINKED ISSUE #1938: Views: Changing view permissions does not move view to correct section
When adding or removing sharing from a view, the view does not move to the correct "My/Shared" views section until the page reloads.

https://github.com/user-attachments/assets/5d290d61-0c55-48d6-b1a9-f447acb0b6fd

filipvnencak:
## Description of changes 
Fixes [YN-1938] — when a view's permissions change (e.g. from private → studio, or studio → project), the view did not move to the correct section in the views dropdown until a hard refresh.      

## Technical details

Fix: extend `providesTags` in `getViews.ts` so each list also provides a per-view tag `{ type: 'view', id: v.id }` for every view in the result. Now any mutation that invalidates `view:<id>` will cause the affected `listViews` queries to refetch and the view will surface in the correct section automatically.

### DIFF
--- a/shared/src/api/queries/views/getViews.ts
+++ b/shared/src/api/queries/views/getViews.ts
@@ -1,10 +1,5 @@
 import { ListViewsApiResponse, viewsApi } from '@shared/api/generated'
-import {
-  DefinitionsFromApi,
-  OverrideResultType,
-  TagDescription,
-  TagTypesFromApi,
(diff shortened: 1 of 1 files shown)


Innders (approved):
Works now thanks

PR #1964, opened 2026-05-04
Overview: Folder grouping should not pin name column

### LINKED ISSUE #1962: Overview: Folder grouping should not pin name column
Historically the name column has been pinned as the first column when grouping the tasks.

When grouping by `folder` remove this restriction so the column ordering and pinning can be anything.

Image

filipvnencak:
## Description of changes 
Removes the forced name-column pinning + first-position when grouping by Folder, so column order and pinning are free. Other groupings (status, taskType, etc.) keep the existing behavior.

## Technical details

- ColumnSettingsProvider.tsx — gated the name-pin / name-first block on groupBy.id !== 'folder'.


Innders (approved):
Works thanks

PR #1967, opened 2026-05-05
Project: Add label alongside project name

### LINKED ISSUE #1940: Project settings: Add label alongside project name
## Story
As a **studio admin**, I want to set a human-readable label on a project so my team sees a friendly display name without me having to rename the project's machine identifier.

As an **artist / end user**, I want to see meaningful project names in the sidebar, project manager, and breadcrumbs instead of technical slugs like `prj_client_2026_v02`.

### Problems

- Project `name` is the machine identifier: primary key, routing slug, referenced in logs and URLs. It has to stay stable and regex-constrained.
- That leaves no room for a nice human-readable display name (spaces, punctuation, proper casing).
- Renaming a project to make it prettier is expensive and risks breaking integrations and deep links.

### Proposal
Use the optional `label` field being added on the backend (ynput/ayon-backend#915) as a display-only layer on top of `name`. When a label is set, show it; when it isn't, fall back to `name`. Machine ID / routing / state keys stay on `name` — only the visible text changes.

- Render `label ?? name` everywhere the project is shown to end users (sidebar, Project Manager table, breadcrumbs / page titles).
- Keep `name` visible wherever it's used technically (URLs, logs, copy-pasteable IDs).
- Introduce a single `getProjectDisplayName(project)` helper in `shared/src/util/` so the fallback isn't scattered across components.
- **Edit label via right-click context menu** on a project row in the Project Manager list — "Edit label" (or "Rename") opens a small inline prompt / dialog, writes via PATCH

## TODO
- [ ] Add `getProjectDisplayName(project)` helper in `shared/src/util/`
- [ ] Swap `ProjectMenu/projectMenu.jsx` (sidebar) to use the helper
- [ ] Swap `ProjectsList/ProjectsListRow.tsx` + `buildProjectsTableData.ts` to use the helper
- [ ] Swap breadcrumb / page-title components that render project name to use the helper
- [ ] Add **right-click context menu** on project rows in `ProjectsList` with an "Edit label" action
- [ ] Context menu action opens an inline edit / small dialog prefilled with the current label (or empty)
- [ ] Wire the edit to the project PATCH mutation — save on Enter / confirm, cancel on Esc
- [ ] Do NOT change any state selector or routing key from `name` to `label` — label is display-only
- [ ] Verify deep links using `name` still resolve (no routing regressions)
- [ ] Verify project selection, switching, and recent-projects memory still work unchanged

### Attachments
- Backend PR: https://github.com/ynput/ayon-backend/pull/915

filipvnencak:
## Description of changes 

Show optional project.label as a friendly display name (falls back to name) in sidebar, Project Manager list, page title, delete confirm, and user-access filter. Inline-edit label via right-click → "Edit label" (or R) on project rows.

## Technical details

- New getProjectDisplayName(project) helper in shared/src/util/.
- buildProjectsTableData swaps row display label only; routing, selection, pinned memory still keyed on project.name.
- Sort + sidebar search use display name; SimpleTable fuzzy filter already covered both fields.
- Label PATCH via useUpdateProjectMutation with optimistic update on listProjects cache (rollback on failure).
- Manager+ permission gate, mirroring archive/delete.
- Breadcrumbs left untouched — TODO added (needs per-scope crumb detection).

### DIFF
--- a/shared/src/api/queries/project/updateProject.ts
+++ b/shared/src/api/queries/project/updateProject.ts
@@ -29,6 +29,33 @@ const enhancedProjectApi = projectApi.enhanceEndpoints({
               { type: 'projects', id: 'LIST' },
             ],
       transformErrorResponse: (error: any) => error.data.detail,
+      // Optimistically patch listProjects cache for fields exposed in the list view
+      // so the UI updates instantly without waiting for refetch.
+      async onQueryStarted({ projectName, projectPatchModel }, { dispatch, queryFulfilled }) {
+        if (!('label' in projectPatchModel)) return
+
+        const newLabel = projectPatchModel.label ?? undefined
+        const patches = [
+          dispatch(
+            projectApi.util.updateQueryData('listProjects', { active: true }, (draft) => {
+              const project = draft.find((p) => p.name === projectName)
+              if (project) project.label = newLabel
+            }),
+          ),
+          dispatch(
+            projectApi.util.updateQueryData('listProjects', { active: undefined }, (draft) => {
+              const project = draft.find((p) => p.name === projectName)
+              if (project) project.label = newLabel
+            }),
+          ),
+        ]
+
+        try {
+          await queryFulfilled
+        } catch {
+          patches.forEach((patch) => patch.undo())
+        }
+      },
     },
     setProjectBundle: {
       invalidatesTags: (_r, _e, { projectName }) => [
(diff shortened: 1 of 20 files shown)


Innders (changes requested):
Works nicely, just need to add it to one more place!

- [x] Add project label to project details panel on dashboard page

Image

Innders (changes requested):
Sorry I wasn't clear enough.

Name and Label need to be fields with label being editable.

Image

Image

Innders (approved):
Works great!

PR #1972, opened 2026-05-06
YN-0717: Date: Setting date can set it to day before

### LINKED ISSUE #1969: YN-0717: Date: Setting date can set it to day before

When setting the start/end date for a task/folder the date that gets set to the date before what was selected.

&nbsp;

it's only been replicated on MacOS so far. 

_No response_

### Additional context:

_This issue was automatically created from AYON ticket [YN-0717]( https://support.ayon.app/projects/Active_Clients/overview?project=Active_Clients&type=folder&id=1a5f60d0488311f1b160af2f86a7cbda)

filipvnencak:
## Description of changes 

Fixes off-by-one day when setting Start/End dates on entities in timezones west of UTC (e.g. US Pacific). Reported on macOS but the underlying cause is timezone, not OS.

## Technical details

Root cause: date-only fields are stored as UTC midnight (YYYY-MM-DDT00:00:00.000Z) but displayed via date-fns format(), which renders in the host's local timezone. In any TZ with a negative UTC offset, UTC midnight resolves to the previous day local — so picking May 10 displayed as May 9, every reopen-and-save drifted another day.

### DIFF
--- a/shared/src/components/DateRangePicker/DateRangePicker.tsx
+++ b/shared/src/components/DateRangePicker/DateRangePicker.tsx
@@ -12,6 +12,7 @@ import {
 } from 'react'
 import { createPortal } from 'react-dom'
 import { format, isValid, parseISO } from 'date-fns'
+import { formatUTCDate } from '../../util/formatUTCDate'
 import * as Styled from './DateRangePicker.styled'
 import clsx from 'clsx'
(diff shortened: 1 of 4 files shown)


Innders (approved):
Seems to work for me and makes sense. Thanks.

PR #1973, opened 2026-05-07
Details panel: Add version upload dialog with product picker and task linking

### LINKED ISSUE #1918: Details panel: Add version upload dialog with product picker and task linking
## Story
As a user, I want the same version upload dialog (with product picker, version calculation, and task linking) when uploading a version from the details panel, so I have consistent control over the upload, regardless of the entry point.

### Problems

- The overview page version upload now provides a full dialog with product selection, automatic version calculation, and the ability to link a task to the uploaded version.
- - The details panel version upload does **not** show this dialog — it bypasses product picking, version calculation, and task linking entirely.
 - This inconsistency means users uploading from the details panel lose control over which product the version targets, what version number is assigned, and
   cannot link a task.

### Proposal
- Reuse the existing version upload dialog (product picker, version calculation, task linking) when uploading a version from the details panel.
 - Ensure behavior is consistent with the overview page upload flow.

## TODO

 - [ ] Integrate the version upload dialog into the details panel upload flow
 - [ ] Ensure product picker, version calculation, and task linking all work in this context

- [ ] ...

filipvnencak:
## Description of changes 
Details Panel "Upload version" now opens the same rich dialog (product picker, version calc, task linking) used on Overview/VP pages. Before: simple inline dialog that bypassed picker/calc/linking.

Works from both triggers:                                                                                                                                                                                      
  - Drag files onto the panel → "Upload version" drop zone                                                                                              
  - Right-click thumbnail → "Upload version" 

## Technical details

§- Added `useOptionalVersionUploadContext()` — non-throwing variant of the hook for shared components.  
- `EntityPanelUploader.handleVersionUpload` now intercepts when the provider is mounted: queues files into context `pendingFiles`, opens dialog with prefilled product/folder/task/latestVersion. Existing dialog flow does the rest.         
- Inbox page now wraps its Details Panel with `VersionUploadProvider` + `UploadVersionDialog` so the new flow works there too. Tasks Progress was already covered (it renders inside ProjectPage which already mounts the provider).

### DIFF
--- a/shared/src/components/EntityPanelUploader/EntityPanelUploader.tsx
+++ b/shared/src/components/EntityPanelUploader/EntityPanelUploader.tsx
@@ -15,6 +15,7 @@ import { toast } from 'react-toastify'
 import { useReviewablesUpload } from '../ReviewablesList'
 import { useDetailsPanelContext } from '@shared/context'
 import EntityPanelUploaderDialog from './EntityPanelUploaderDialog'
+import { useOptionalVersionUploadContext } from '@shared/components'
 import {
   sanitizeProductName,
   createProductHelper,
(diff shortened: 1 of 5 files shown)


Innders (changes requested):
Works nicely!

Could you please just take a look at this bug as well.

The No versions published yet message is showing even when I have just published a new version.

Image

Innders (approved):

PR #1985, opened 2026-05-12
Group by: lost when having additional view opened when refreshing the page

### LINKED ISSUE #1980: Group by: lost when having additional view opened  when refreshing the page
## Story
When having opened e.g. Details panel and any sort of Groupping and refresh the page via F5 it looses the groupping and get disabled.

Image

### To reproduce

1. Navigate to Products page and enable any sort of groupping
2. Select some product and open the Details panel
3. Refresh the page
4. Groupping is lost

### Expected behavior

Group by should stay enabled and configured when refreshing the pages

## Environment

Latest server 1.15.1
WIN 11
MS Edge (also spotted in Chrome)

filipvnencak:
## Description of changes 

Fixes Group by  being wiped when refreshing a page that has the Details panel open. Affects Products, Overview, and Tasks Progress.

## Technical details

- `onUriOpen` now receives `source: 'uri' | 'url'`. Page-level handlers only reset filters / groupBy / hierarchy when `source === 'uri'` (true URI navigation); plain URL refresh (`?project&type&id`) leaves view state alone.
- Fixed a race in `VPToolbar.handleViewGroupByChange`: picking a new option fired both a sort-only `onUpdateColumns` PATCH and the grouping PATCH in parallel, and the stale columns payload could land last and overwrite `showProducts: true`. Sort-only PATCH is now skipped when the grouping selection itself is changing, and the grouping update is awaited.
- Same gate applied in `ProjectOverviewPage` and `TaskProgressDetailsPanel`.
## Additional context

 Closes #1980

### DIFF
--- a/src/pages/VersionsProductsPage/components/VPToolbar/VPToolbar.tsx
+++ b/src/pages/VersionsProductsPage/components/VPToolbar/VPToolbar.tsx
@@ -74,14 +74,15 @@ const VPToolbar: FC = () => {
   }, [viewGroupBy, viewGroupByOptions, columns.groupBy, sorting])
 
   const handleViewGroupByChange = useCallback(
-    (values: { id: string; sortOrder?: boolean }[]) => {
+    async (values: { id: string; sortOrder?: boolean }[]) => {
       const value = values[0]
 
       // Get the grouping selection from the dropdown (what the user selected)
       const selectedGrouping = !value ? undefined : value.id === 'product' ? 'hierarchy' : value.id
+      const isGroupingChange = selectedGrouping !== viewGroupBy
 
       // Handle sort order changes by updating the columns config
-      if (value && value.sortOrder !== undefined) {
+      if (value && value.sortOrder !== undefined && !isGroupingChange) {
         const newDesc = !value.sortOrder // invert sortOrder to desc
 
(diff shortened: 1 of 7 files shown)


src/pages/VersionsProductsPage/components/VPToolbar/VPToolbar.tsx
```diff
@@ -105,14 +106,14 @@ const VPToolbar: FC = () => {
       }
 
       // Handle grouping selection change (only if selection actually changed)
-      if (selectedGrouping !== viewGroupBy) {
+      if (isGroupingChange) {
         if (!value) {
           // X clicked — flat list (no grouping, no products)
-          onUpdateViewGroupBy(undefined)
+          await onUpdateViewGroupBy(undefined)
         } else if (value.id === 'product') {
```
Innders, line 113: `onUpdateViewGroupBy` is not async so await does nothing here. Is it even required?

Innders (changes requested):

Innders (approved):

src/pages/VersionsProductsPage/components/VPToolbar/VPToolbar.tsx
filipvnencak: Hmmm, Not really, but from my point of view it's safe to keep with await for future.

PR #1997, opened 2026-05-15
Comments: fix copy/paste of mentions

### LINKED ISSUE #1968: Comments: Copy and pasting mentions does not work correctly
When mentioning a user like `@luke` and then copying and pasting that text the mention tag is lost.

Ensure that when coping a mention the url is copied `(Luke)[user:luke]` and then when pasting it's converted into a mention tag.

filipvnencak:
## Description of changes 

Fixes copy/paste of mentions in the comment input so pasted mentions stay clickable and keep the linked entity, instead of degrading to plain `@name` text. Also fixes a related bug where pressing space right after a mention deleted it.

## Technical details

- New `MentionClipboard` Quill module (`shared/src/containers/Feed/components/CommentInput/modules/MentionClipboard.ts`) registers capture-phase copy/paste listeners on `quill.root`. On copy it serialises `<MENTION>` blots to markdown link form `[label](type:id)`; on paste it parses the same form and re-inserts mention-attributed deltas. Falls back to the Quill `A` clipboard matcher for external HTML pastes.
- Wired through `modules/index.ts` (`getModules`) with `mentionTypeOptions` passed from `CommentInput` so prefix mapping (`@` / `@@` / `@@@`) is shared with the rest of the input.
- `CommentInput.handleChange`: the auto-delete-whole-mention branch is now gated by `isDelete`. It was firing on any change at a mention-formatted cursor position, so a space typed right after a pasted (non-trailing-space) mention would inherit the mention format and trigger deletion of the whole mention.
- Pins `react-quill-ayon` to `2.0.2` to fix an unrelated regression introduced in `2.0.3`.

### DIFF
--- a/shared/package.json
+++ b/shared/package.json
@@ -139,7 +139,7 @@
     "quill-magic-url": "4.2.0",
     "react-markdown": "10.1.0",
     "react-perfect-scrollbar": "1.5.8",
-    "react-quill-ayon": "2.0.3",
+    "react-quill-ayon": "2.0.2",
     "react-toastify": "9.1.3",
     "rehype-raw": "7.0.0",
     "remark-directive": "3.0.1",
--- a/shared/src/containers/Feed/components/ActivityComment/ActivityComment.styled.ts
+++ b/shared/src/containers/Feed/components/ActivityComment/ActivityComment.styled.ts
@@ -88,6 +88,7 @@ export const Body = styled.div`
     .reference {
       margin-right: -2px;
       top: 5px;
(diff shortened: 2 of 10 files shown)


Innders (changes requested):
It works but the first thing I tried to do was copy a message not in editing mode.

We will need to be able to copy in both editing and non editng modes.

Innders (approved):
Some glitchy-ness but it works well enough and not worth spending more time on. Thanks!

PR #1998, opened 2026-05-15
Pipeline installer: launcher select shows fallback platforms

### LINKED ISSUE #1978: Update Pipeline: Launcher not giving any options
When running Update Pipeline there are no options to select Launcher OS

Image

filipvnencak:
## Description of changes 

Fixes blank "Select launcher platforms" step in the Install pipeline release dialog when the release manifest returns no installers, and adds a loading skeleton so the user no longer sees an empty list mid-fetch.

## Technical details

- `PlatformSelect` now accepts `isLoading` + `placeholderCount` and renders skeleton `AddonCard`s while loading, matching the `AddonsSelectGrid` pattern.
- `ReleaseInstallerPlatforms` falls back to `['windows', 'linux', 'darwin']` when `releaseInstallers` is empty; `getReleaseInstallUrls` already filters by the manifest, so unmatched selections install nothing harmful.

## Additional context

Closes #1978

### DIFF
--- a/src/components/PlatformSelect/PlatformSelect.tsx
+++ b/src/components/PlatformSelect/PlatformSelect.tsx
@@ -1,4 +1,5 @@
 import { forwardRef } from 'react'
+import clsx from 'clsx'
 import styled from 'styled-components'
 import AddonCard from '@components/AddonCard/AddonCard'
 import { getPlatformIcon, getPlatformLabel } from '@pages/AccountPage/DownloadsPage/DownloadsPage'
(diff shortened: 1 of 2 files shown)


MustafaJafar:
Before 
image

After
image

I believe it fixes the mentioned issue but I don't think it solves the underlying problem. 

It could be just me, but the current result for me is the update pipeline progress bar doesn't move and there are no installers brought by the tool.

Innders:
If the problem is with the release manifest then these changes would have made it harder to spot. @filipvnencak in these cases we need to notify the team of the underlying problem. Closing is it's not a fix on frontend.

PR #2000, opened 2026-05-18
Feed: Filter activities by search

### LINKED ISSUE #525: Feed: Filter activities by search
## Story

I want to filter the activity feed using a search so I can find a specific or group of comments.

### Problems

Can you search through everything `comment`, `status`, `assignee`, `publish`?
Will search work with author as well? So you can filter by author name.
Could this be advanced like github and have after-date filtering etc?

This will need to done server side.

#### Search string
- Search by string that matches against body.
#### Filters
- Author
- Has attachments
- Categories
- Type: comment, status, version etc
- Review session - filter out comments from review sessions, or pick specific review sessions.
- Guests - filter out or show only guest comments (maybe this could be done with categories?)
- Tags (not supported in UI yet) but should support on server

### UI
- Dedicated tab switch to filtering mode
- Filtering bar across the top.
- Compact version of SearchFilter bar

Image

filipvnencak:
Feed: filter activities by search and user

## Description of changes 

Adds a search bar plus user and attachment filters to the activity feed so users can find specific comments and events fast. Builds on the existing type chips; combines with them using AND.

## Technical details

- New `$activityFilter: String` variable on `GetActivities`; `buildBackendFilter` translates the UI `feedFilter` (body, author, has_attachments) into the backend `QueryFilter` JSON, strips UI-only chip keys, and OR-combines `activity_data.author` with `activity_data.assignee` so picking a user surfaces both their actions and assignment events involving them.
- Search input is debounced 300ms via a new shared `useDebouncedValue` hook to avoid one request per keystroke.
- Mutation cache invalidation in `updateActivities.ts` now `JSON.stringify`s object filters via a `serializeFilter` helper so optimistic patches reach the right cache entry when a filter is active (was coercing to `[object Object]` and missing every match).
- Depends on the new `filter` arg landing in `activities()` (ynput/ayon-backend#933, including `activity_data` / `reference_data` in the whitelist).

## Additional context

Closes #525


Innders (changes requested):
Great start!

- [x] The assignees dropdown needs to have images.
- [x] Versions: Missing the ability to filter by if the comment was made in a review session or not. It should use the review session icon (`subscriptions`) and when enabled only show comments made within a review session.
- [x] Missing the ability to filter the activities by category. You should be able to pick one or multiple categories. For example: filtering activities by the category `director` and `guest`. "No category" should also be an option.

Innders (changes requested):
- [x] Keep users as the last filtering
- [x] Remove `x` icon from No category (have no icon)
- [x] No category does not work
- [x] Add review session filtering to folders as well

Image

Innders (commented):
The idea and concept is great. Now this slight re-design needs to implemented into the SearchFilter component so that it can be used everywhere.

https://github.com/ynput/ayon-react-components/issues/211

Innders:
@filipvnencak is it possible to implement a version that uses current SearchFilter but more compact?

Then we can look at improved search filter at another time.

I would like to get this over the line ASAP

PR #2007, opened 2026-05-20
Home: Details panel can become 0 width

### LINKED ISSUE #2005: Home: Details panel can become 0 width
It's hard to replicate but randomly clicking around the tasks page, switching between projects can sometimes cause the details panel to have width 0.

Image

https://discord.com/channels/517362899170230292/1001811244535259207/1506574516602671174

filipvnencak:
## Description of changes 

Fixes the Home details panel occasionally collapsing to 0 width after clicking or dragging task cards on the Kanban view.

## Technical details

- Removes the `isDragging ? 0 : ...` width collapse on the details `SplitterPanel` in `UserTasksContainer.jsx`. The 2024 cursor-alignment workaround is obsolete now that `DragOverlay` (via `KanBanCardOverlay`) tracks the cursor itself. Any drag-start without a paired drag-end (double-click race, view switch, dnd-kit edge case) used to leave the panel stranded at width 0.
- Drops the orphan `draggingIds` state, `onDraggingStart` / `onDraggingEnd` reducers, and their dispatches in `UserDashboardKanBan.jsx`'s drag handlers — no remaining consumers.
- Cleans up the dead `StyledSplitter` block that targeted the removed `.dragging` className.

## Additional context

Closes #2005

### DIFF
--- a/src/features/dashboard.js
+++ b/src/features/dashboard.js
@@ -16,7 +16,6 @@ const dashboardSlice = createSlice({
       assignees: getInitialStateLocalStorage('dashboard-tasks-assignees', []),
       assigneesFilter: getInitialStateLocalStorage('dashboard-tasks-assigneesFilter', 'me'),
       collapsedColumns: getInitialStateLocalStorage('dashboard-tasks-collapsedColumns', []),
-      draggingIds: [],
     },
   },
   reducers: {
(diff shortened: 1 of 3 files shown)


Innders (changes requested):
We need a solution that maintains the original behaviour. When dragging a task the details panel should be hidden .

Innders (changes requested):
Moving events on a far right column is super broken and it was broken before. There is no easy fix for that. Sorry can you please go back one commit so the details panel is NOT hidden when dragging, that actually works better.

Image

PR #2010, opened 2026-05-21
Projects page redesign

Innders:
# Projects Page
### A page dedicated to viewing and managing projects.

- Replaces the old Dashboard page
- Moved navbar Projects button to left, removed label and changed icon.
- Supports project folders and all project folder management
- Improve project details panel that's more in line with the entity details panel.
- Support editing project attributes inline
- All project actions on right click are supported

# ListTable
### A new reusable table component that functions more like a list of items with adjustable columns.

- Keyboard support
- Supports column resizing, reordering, sorting and grouping
- Grouping (by multiple levels support) through tree structured data
- Right click context menu API
- Re-ordering items support 
- Uses tanstack table
- Virtualized rows with lazy loading support
- Inline cell editing
- Selection and multi selection of rows support

image


martastain (changes requested):
- [x] - It is not possible to access skeleton projects?
- [x] - One of the column is "Folder" which show the folder id. It should be a label
- [x] - Attribute column headers show the attribute name instead of title
- [x] - Clicking a project requests the project thumbnail 6x every time
- [x] - Project folders no longer require powerpack installed?
- [ ] - When editing a first project in the group, it is easy to misclick and hit the "group row" and collapse the whole group. especially when you scroll right, it is very confusing. 
- [ ] - Could we pin the project label column so it is always visible? even when scrolling horizontally?

Patching a project loads the thumbnail 8x, and then uses 4 different api calls to reload the data:

- [x] - graphql project list
- [x] - 2x get project list endpoint
- [x] - 1x get project details

image


- [x] - With every folder deeper, thumbnail aspect ratio breaks more and more until it is just a vertical line. I'd say the thumbnail can maintain the aspct ratio specified by resolutionWidth and resoulutionHeight attributes (or the image itself)

image

Innders:
> Patching a project loads the thumbnail 8x, and then uses 4 different api calls to reload the data:
> 
> * [x]  - graphql project list
> * [x]  - 2x get project list endpoint
> * [x]  - 1x get project details

There might be other components using those queries so the data needs refreshing

src/pages/ProjectManagerPage/ProjectManagerPage.jsx
```diff
@@ -97,10 +97,6 @@ const ProjectManagerPage = () => {
     })
   }
 
-  const handleActivateProject = async (sel, active) => {
```
filipvnencak, line 100: Need to be defined or removed from line 209 - app crashing

```
 <ProjectManagerPageContainer
          selection={selectedProject}
          onSelect={setSelectedProject}
          onNoProject={(s) => setSelectedProject(s)}
          isUser={isUser}
          onNewProject={() => setShowNewProject(true)}
          onDeleteProject={handleDeleteProject}
          onActivateProject={handleActivateProject}
          customSort={projectSorter({ isLoadingUserPermissions, userPermissions, module })}
          isActiveCallable={isActiveDecider({ userPermissions, projectName, module })}
        >
```

filipvnencak (changes requested):
Clicking on manage project app crashed 

There is little glitch on updating label, the thumbnail blink with alt text
https://github.com/user-attachments/assets/fe61beea-133c-446f-b758-3a42bbec99f2

And I don't know it is related but when I right click on project which is in the folder an I want to create and move to new folder it creates root folder and not nested 

https://github.com/user-attachments/assets/48041571-47df-4313-93de-8c28b38578fc

The default Folder icon missing on list 
Screenshot 2026-05-22 at 10 20 52

Thumbnail after upload is visible only after reload
https://github.com/user-attachments/assets/ae54ffea-3fd5-446f-a21f-8543c299096c

filipvnencak:
The Archivation from context menu select different row, but when slicking on Active checkbox it doeasnt't




https://github.com/user-attachments/assets/cbfb1e1c-82b8-4f11-b570-5a11037c679c

filipvnencak:
Can we also update Create project dialog with label input?

Screenshot 2026-05-22 at 10 32 44

PR #2011, opened 2026-05-22
Overview: include empty folders in search results

filipvnencak:
## Description of changes 

Empty folders matching free-text search now appear in Overview results. Previously the folder lookup used `taskSearch` only, so a folder whose name matched the query but had no tasks was hidden.

## Technical details

- `useFetchOverviewData` now sends `folderSearch: folderFilters.search || taskFilters.search` to `/folders/search`, so the backend can resolve folders by their own name/path when the search is unscoped task text.

## Additional context

But beckend returning the empty array


Innders (changes requested):
The root cause appears to be that taskSearch and folderSearch are currently combined using AND logic instead of OR.

While removing taskSearch entirely allows the folder search to work, we can't do that because we still need to support searching by tasks. Instead, we need to update the logic so a folder is returned if it matches the folder search OR contains a matching task. This fix will likely need to be handled on the backend. @martastain

Innders:
@filipvnencak what's happening with this? Does a backend ticket need to be created for it?

PR #2025, opened 2026-06-03
Links optimisations and table column defaults

Innders:
## Links loading

Before this PR links would almost always load on all pages regardless if they were on the table tab or even a visible column.

Now we selectively only load links when we need them.

Skip loading links when
- Not on table tab
- All links columns are hidden
- All links columns are off the page and not in the view port

## Products page columns

The products page columns was not connected to views so any changes to columns like size etc would not be persisted.

Now changes are saved to the view.

## Default columns

The default columns for the different pages and entity types is way more restricted and keeps most columns (including links) hidden by default.

You can see the defaults here https://github.com/ynput/ayon-frontend/blob/5bff32ed1ffc183b008235ab6cde59a53f2955c2/src/pages/ProjectsPage/constants.ts


filipvnencak (changes requested):
Fresh view shows some columns but then when want to show more clcikng on 3 or 4 show all at once


https://github.com/user-attachments/assets/e27660ce-dff5-40ec-bef0-f57196c49ff0

PR #2026, opened 2026-06-04
Lists realtime updates fixes

Innders:
## Realtime updates

1. Removes query invalidation so that the lists are not refetcech on EVERY ws message.
2. Uses only the summary to update the list (with no additional API calls).
3. Handles create and delete events
4. Batches cache updates to happen one every 10 seconds in one update cache call.

## Creating new list fixes

1. Fixes issue where `n` shortcut would not work when list was already selected (which is always)
2. When a list is already selected inside a folder, the new list inherits the parent folder so the new list is put in the same folder.

### DIFF
--- a/shared/src/api/queries/entityLists/getLists.ts
+++ b/shared/src/api/queries/entityLists/getLists.ts
@@ -24,6 +24,40 @@ import {
   ListsPageParam,
 } from './types'
 
+// Helper function to batch pub/sub messages to avoid constant cache updates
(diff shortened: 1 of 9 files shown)


filipvnencak (changes requested):
when switching between projects there is no cache update on lists. I removed one list and in the other browser the Lists page still contains it, clicking it causes an 

https://github.com/user-attachments/assets/6c03cc2d-4b93-4255-af7c-66d3820e01fa

error because this list no longer exists.

PR #2040, opened 2026-06-09
Thumbnails hash

Innders:
## Description of changes 
Ensure all entity thumbnails use `?hash={thumbnailHash}`
image

## Testing

- Check all entity thumbnails have `?hash=` value
- Use backend branch https://github.com/ynput/ayon-backend/pull/954


filipvnencak (changes requested):
The thumbnail in Project detail panel broken.

https://github.com/user-attachments/assets/2fe4bddc-647e-4d93-b4ad-bef37383a5b4

martastain:
> The thumbnail in Project detail panel broken.
> 
I've fixed project updated at bump in the backend, but that only solves the problem in the table - i assume entity card uses thumbnailHash already, which we don't support yet on projects. .... could we inject updatedAt as hash there, @Innders ?

image

martastain (changes requested):
ThumbnailHash needs to be invalidated upon thumbnail upload - at least locally. The logic seems to work, but when a user uploads a new thumbnail, it is not propagated and page must be reloaded. 

QuickFix: inject random string as new thumbnail hash
IdealFix: listen to thumbnail.updated events and patch thumbnail hash to new actual hash from the event payload - the event summary contains entityType, entityId and thumbnailHash and the event comes in for every affected entity

martastain (approved):
Works great!

PR #2047, opened 2026-06-10
Activity Feed: Optional "version review" activities

jakubfiala:
## Description of changes 

Adds a new optional feature to the `Feed` component which displays activities of type `version.review`, and also allows  users to submit the version reviews directly from the CommentInput component. The last version review by the current user is displayed as a pill above the comment input.

## Technical details

- the feature can be toggled using the `versionReview` prop on the `Feed`
- adds a new `mine` argument to the `GetActivities` graphql query so the latest version review for a given user can be retrieved even if it's not already loaded in the feed (e.g. because it's too far back in the feed timeline)
- the `useLastVersionReview` hook ensures that if there is already a version review found in the loaded feed activities, no extra query is sent. Only once the feed has loaded and there's no version review from the current user found in it, is the additional query sent to check if there really isn't an older version review.

## Additional context

needs https://github.com/ynput/ayon-backend/pull/964 to work correctly

related to (but doesn't completely close) https://github.com/ynput/ayon-review/issues/213


Innders (changes requested):
- [x] Make the big review buttons the same width

Image

- [ ] Only show the tooltip if the labels are hidden.
- [x] Make review buttons height 32px so that it matches the Comment button

Image

- [x] If the tag button are present, collapsed the review buttons earlier. (I know the guests don't see the buttons but reviews might get opened up to regular users one, day just future proofing.)

Image

- [x] Drop to icons when there is not enough room here for the text

Image

- [x] If I create a review with some text and then click the button again (with empty text) then a n empty comment will be created.

Image

- [x] Improve the inbox message item. Use the icon and colour.

Image

Image

jakubfiala:
@Innders addressed all feedback (except the tooltips, as per discussion in Discord)

The inbox messages look like this (except the brackets, i removed those since the screenshot).

image

Innders (changes requested):
All those changes look good. Now just looking at it within review session.

- [x] Slight misalignment of category
- [x] Hover over the time and the activity goes crazy if it's too narrow. Replace the time hover effect with a tooltip that shows fill timestamp with 0 delay. Do this for all activities.
- [x] The guest should **not** see the category in the review activity.

Image

- [x] The comment box is floating at the top of the screen and needs to be at the bottom.

Image

- [x] (check) Normal users seeing buttons, but I might have done something wrong here.

Image

jakubfiala:
@Innders I haven't even opened a PR for including this into Review yet, so it would be worth reviewing the code once i've written it :) I will definitely not leave the layout the way it is in the screenshots, and of course I'll be sure to switch off the feature for non-guest users.

But the category misalignment + changing how the date gets rendered is good to know, that can be tweaked in this PR. I also need to change something about the GraphQL query anyway.

jakubfiala:
@Innders updated now. Please note the backend has also changed: https://github.com/ynput/ayon-backend/pull/975

As mentioned, I'll address the layout of the Review addon feed in the Review addon.

Innders (changes requested):
The padding of the comment input seems to be different when the review is enabled/disabled. The padding on enabled seems to be more and should always be the smaller padding when disabled.

(Not a big deal at all if it's a PITA)

Image

jakubfiala:
@Innders are you sure? to me it looks exactly the same. I'm not changing it, see here: https://github.com/ynput/ayon-frontend/pull/2047/changes#diff-baa42281680515bcff6c5a13441fcc0d6fe8d14b3701789590597ece5dad10cb

Innders (approved):

PR #2048, opened 2026-06-10
Slicer: persist selection per project, page and slice type

filipvnencak:
Slicer: persist selection per project, page and slice type

## Description of changes 

Fixes slicer selection leaking between projects. Selection is now saved per project, page and slice type, and switching slice types keeps each type's selection.

## Technical details

- `SlicerProvider` gets `scope={projectName}-{module}`, localStorage keys become `{project}-{page}-slicer-*` keyed by slice type
- slice type change no longer resets selection  the new storage key loads it
- new effect in `Slicer.tsx` rebuilds `rowSelectionData` once rows load, so restored selection actually filters (before it only highlighted)


martastain (approved):
works!

PR #2049, opened 2026-06-10
Representations: make columns resizable

antirotor:
## Description of changes 
Make columns in representation list resizable.

## Additional context

Some texts in representations were longer and therefore unreadable, this little fix make it more accessible.

### DIFF
--- a/shared/src/containers/RepresentationsList/RepresentationsList.tsx
+++ b/shared/src/containers/RepresentationsList/RepresentationsList.tsx
@@ -104,7 +104,7 @@ export const RepresentationsList = ({ entities = [] }: Props) => {
                 field={col.field}
                 header={col.header}
                 expander={col.expander}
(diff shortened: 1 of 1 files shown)


shared/src/containers/RepresentationsList/RepresentationsList.tsx
```diff
@@ -102,7 +104,7 @@ export const RepresentationsList = ({ entities = [] }: Props) => {
                 field={col.field}
                 header={col.header}
                 expander={col.expander}
-                style={{ width: col.width }}
+                style={{ width: `${col.width}px` }}
```
Innders: Don't think you need this. When using inline styles, integers are always treated as pixels by default.

Innders (changes requested):

Innders (approved):

PR #2052, opened 2026-06-11
Overview: do not include slicer folder in table

### LINKED ISSUE #2050: Overview: Do not include slicer folder in table
When filtering down the folders and tasks by the slicer. The selected folders should not be included.

You should be looking at the contents inside of the folder excluding the folder itself.

Image

This particularly apparent when you look at the count of folders.

Image

You can to be able to see that the folder has 4 folders (shots) inside of it.

filipvnencak:
## Description of changes 

Selecting folders in the slicer now shows only their contents in the table, the selected folder row itself is excluded, so folder counts reflect what is inside the folder.

## Technical details

- `useFetchOverviewData` gains an `excludeSelectedFolders` flag: `foldersMap` keeps only strict descendants of the selected paths, which promotes their children to root rows; the selected folders' direct tasks are fetched eagerly (there is no row left to expand) and their ids are re-included in the flat-list/grouped task query `folderIds`.
- `useBuildProjectDataTable` renders tasks of excluded selected folders as root rows (`createRootTaskRows`) — covers selecting a leaf folder that only contains tasks, in both hierarchy and flat-folder views.
- `ProjectTableProvider` threads an optional `selectedFolders` prop into the table build; entity-list slices keep showing the listed folders themselves (`sliceType !== 'entityList'` gate in `ProjectOverviewContext`, plus a `foldersMap.has` guard).
- Footer summaries (`ProjectOverviewTable`): folder stats use descendants only, task stats union in the selected folder ids so their direct tasks stay counted.

## Additional context

Closes #2050

### DIFF
--- a/shared/src/containers/ProjectTreeTable/context/ProjectTableProvider.tsx
+++ b/shared/src/containers/ProjectTreeTable/context/ProjectTableProvider.tsx
@@ -77,6 +77,7 @@ export interface ProjectTableProviderProps {
   entitiesMap: EntitiesMap
   tasksByFolderMap: TasksByFolderMap
   tableRows?: TableRow[] // any extra rows that we want to add to the table
+  selectedFolders?: string[]
 
   // grouping
   groups: EntityGroup[]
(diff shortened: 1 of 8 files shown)


Innders:
When creating new folders the slicer selection is used for the parent folders if there is no table selection.

The new entity dialog now has a dropdown that states what folders you are creating inside of with the ability to pick.

image

Innders (approved):

Issue #974, opened 2025-01-13
AY-7317_Password reset requirements hint


ynbot:
### Additional context:
link to discussion on [Discord](https://discord.com/channels/517362899170230292/1328329723952238674/1328329730117599263)
_(might be a private channel)_

_This issue was automatically created from Clickup ticket [AY-7317](https://app.clickup.com/t/86c1m4zxv)_

Innders:
- Password reset page
- Users password set dialog

![Image](https://github.com/user-attachments/assets/1fa628a2-d982-4297-80e1-a97647e3794d)

PR #2066, opened 2026-06-16
Views: Removed strongly typed View Settings

filipvnencak:
## Description of changes 

Frontend now owns its view settings types instead of generating them from the REST schema, so adding a view setting no longer needs a backend change. Pairs with backend #974.

## Technical details

- Settings types moved to `shared/src/api/viewSettings.ts`; all consumers repointed there.
- `views.ts` regenerated against ynput/ayon-backend#974 -> `GenericView*`.
- Codegen renamed the create/update body arg, so `updateViews.ts` pins a stable `payload` contract to keep saves working.

## Additional context

Pairs with ynput/ayon-backend#974

### DIFF
--- a/shared/src/api/queries/views/updateViews.ts
+++ b/shared/src/api/queries/views/updateViews.ts
@@ -178,7 +178,7 @@ const updateViewsApi = getViewsApi.enhanceEndpoints<TagTypes, UpdatedDefinitions
     },
     updateView: {
       // See createView note — keep the FE on `payload` despite codegen renaming the arg.
-      query: ({ viewType, projectName, viewId, payload }: any) => ({
+      query: ({ viewType, projectName, viewId, payload }: UpdateViewArg) => ({
         url: `/api/views/${viewType}/${viewId}`,
         method: 'PATCH',
         body: payload,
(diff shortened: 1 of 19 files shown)


kalisp (approved):
Ran this with BE PR, switched between stored Views, re-saved some of them.

Didn't explode.

shared/src/api/queries/views/updateViews.ts
```diff
@@ -1,10 +1,55 @@
...
+  updateView: OverrideMutationArg<Definitions['updateView'], UpdateViewArg>
+}
+
+const updateViewsApi = getViewsApi.enhanceEndpoints<TagTypes, UpdatedDefinitions>({
   endpoints: {
     createView: {
+      // Codegen names the request body arg after its model (genericViewPostModel).
+      // Override the query so the whole FE keeps passing `payload`, regardless of
+      // how codegen names the arg on the next regen.
+      query: ({ viewType, projectName, payload }: CreateViewArg) => ({
```
Innders, line 47: I'm not a fan of overriding the original query just so that it can use payload argument.

Can we please remove this and update the args to use `genericViewPatchModel` and convert all queries to use it.

OR get backend to change the arg name to something nicer (preferred).

Innders (changes requested):

PR #2086, opened 2026-06-18
Products: New and updated versions realtime events optimisations

### LINKED ISSUE #2084: Products: Real time updates performance issues
## Problem Description
During publishing, real-time (RT) updates trigger all connected clients to concurrently call the GetVersions GraphQL endpoint. Because GetVersions is the most complex query in the system, this thundering herd problem causes severe server degradation, leading to response times of up to 20 seconds.

Frustrated users frequently reload the page, generating a cascading wave of duplicate requests that worsens the overload.

## Technical Constraints & Root Cause
Dynamic Content: The query cannot simply be cached globally or run once because it dynamically handles unique user filters and distinct permission scopes per request.

Suspected Code Location: The heavy query logic likely originates from the getVersionsProducts logic in the frontend client:
shared/src/api/queries/versions/getVersionsProducts.ts (lines 928–1048)

##Proposed Remediation Plan
1. Immediate / Short-Term Fix (UX Patch)
Disable automatic RT refetches: Temporarily disable the automatic query trigger when a new version is created/updated.

Implement a "Refresh" banner/button: Instead of auto-fetching, show a UI notification informing the user that new versions are available with a manual "Refresh" button. This removes the automated thundering herd problem and improves the user experience during high-traffic publishing windows.

2. Long-Term Strategy
Stream data via message summary: Transition away from pulling the entire dataset via GraphQL on real-time triggers. Instead, stream the updated or new version data payload directly to clients using a message.summary event. The client can then update its local state incrementally without hitting the complex GetVersions endpoint.

Innders:
Completely rewrite the real time updating logic for versions.

## Before

Any change to a version would trigger a single request to `GetVersions`. 100 changes in a minute would result in 100 new requests to the server.

### Delete

Removes the version from the users cache, no api calls required.

### Updates

First checks if the version is even in the users cache and does nothing if it is not.

If the update is a change to `status` or `tags` the version data is patched using the `summary.value` value no api calls required.

If the update is anything else, like an attrib change, a new api request is made to a new simple graphql query `GetVersionsAttribs`.

```
query GetVersionsAttribs($projectName: String!, $versionIds: [String!]!) {
  project(name: $projectName) {
    versions(ids: $versionIds) {
      edges {
        node {
          id
          allAttrib
        }
      }
    }
  }
}
```

Note: In the future attrib change values will be in the message.summary so this api call won't be required.

### New versions

When a new version comes in we must, unfortunately, use the heavy `GetVersions` graphql query to:

1. Find out if the new version matches the current users filters.
2. Get all the version data required for the table.

Once we have the data, it is patched into the cache.

## Debouncing, Batches and Jitter

A big optimisation in the way lots of new messages are handled.

**Debouncing**

For updates we debounce new messages for 5 seconds.
For new versions we debounce for 20 seconds.
This means we updating the users caches far less, improving react performance and usability.

**Batching**

We store the messages in a pool and then handle them all in one query and patch. This means if there are 10 new versions over the space of 20 seconds, we make 1 graphql request for the 10 versions and then update the cache once with the 10 new versions. This should ensure the server does not get overwhelmed when lots of new versions are created and caps the max number of requests to 1 per 20 second.

**Jitter**
For all api requests there is a random jitter of 1500ms to ensure that all users are not making lots of requests at exactly the same time.

### DIFF
--- a/shared/src/api/queries/versions/gql/GetVersionsAttribs.graphql
+++ b/shared/src/api/queries/versions/gql/GetVersionsAttribs.graphql
@@ -1,0 +1,12 @@
+query GetVersionsAttribs($projectName: String!, $versionIds: [String!]!) {
+  project(name: $projectName) {
+    versions(ids: $versionIds) {
+      edges {
+        node {
+          id
+          allAttrib
+        }
+      }
+    }
+  }
+}
--- a/gen/graphql.schema.json
+++ b/gen/graphql.schema.json
@@ -8089,18 +8089,6 @@
                 "deprecationReason": null
               },
               {
-                "name": "filter",
-                "description": "Filter activities using QueryFilter",
-                "type": {
-                  "kind": "SCALAR",
-                  "name": "String",
-                  "ofType": null
-                },
-                "defaultValue": "null",
-                "isDeprecated": false,
-                "deprecationReason": null
-              },
-              {
                 "name": "first",
                 "description": "Pagination: first",
                 "type": {
--- a/shared/src/api/generated/graphql.ts
+++ b/shared/src/api/generated/graphql.ts
@@ -1798,38 +1798,38 @@ export type GetProductVersionsQueryVariables = Exact<{
 }>;
 
 
-export type GetProductVersionsQuery = { __typename?: 'Query', project: { __typename?: 'ProjectNode', product?: { __typename?: 'ProductNode', versionList: Array<{ __typename?: 'VersionListItem', id: string, name: string, version: number }> } | null } };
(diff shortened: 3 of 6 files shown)


shared/src/api/queries/versions/gql/GetVersionsAttribs.graphql
```diff
@@ -0,0 +1,12 @@
+query GetVersionsAttribs($projectName: String!, $versionIds: [String!]!) {
```
filipvnencak, line 1: This query is generated but never used, attrib changes still fetch the heavy `GetVersions`

filipvnencak (changes requested):

filipvnencak (approved):

PR #2091, opened 2026-06-20
Staging mode for actions

Innders:
## Description of changes 
Users can now have `Staging access` enabled to give them access to staging actions.

## Give access to Staging Mode

To allow a user enable staging mode they need to have the `Staging access` field enabled on their profile. This can only be done by and admin or manager.

## Enable Staging Mode

Once a user has access to Staging Mode they will see a switch appear at the top of their nav bar, where they can toggle staging mode on/off.

If the user has access to developer mode and staging mode then the switch will become a dropdown. The user can only be in one kind of mode at a time.

image

## Actions

When a mode is selected the actions menu will be highlighted with the mode color and the actions will be scoped to the bundle with the current mode.

### DIFF
--- a/shared/src/api/queries/users/getUsers.ts
+++ b/shared/src/api/queries/users/getUsers.ts
@@ -17,6 +17,7 @@ const USER_BY_NAME_QUERY = `
           isManager
           isService
           isDeveloper
+          isStagingAllowed
           isGuest
           active
           accessGroups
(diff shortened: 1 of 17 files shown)


filipvnencak (approved):
Tested staging + developer mode — switch, dropdown and action scoping all work. Adding/revoking access applies correctly after reload. LGTM 👍

BigRoy:
This is massive 🚀

PR #2095, opened 2026-06-24
Groups: show filter-aware count and percentage on group rows

### LINKED ISSUE #2083: Groups: Show summaries counts on groups
## Story

I want to group tasks by status and see how many tasks are in each status.
I want to be able to see how many tasks an assignee has.

### Problems

The count from the `/grouping` endpoint does not take into account filters and therefore cannot be used.

### Proposal

Use the new summaries endpoint to get this data for the attribute.

This is part of community AYON.

Image

## TODO

- [x] Write a query hook dedicated to getting breakdown summary for current group.
- [x] Use for tasks on overview page and versions on products page.
- [x] Show the data for each group on the group row header, including percentages.

filipvnencak:
## Description of changes
  
Group rows now show a filter-aware count and percentage of total when grouping by status, type, assignees, tags, or an enum attribute. No PowerPack license required.
  
## Technical details

  
- New `useGroupCounts` hook: one distribution stats call for the grouped field, reusing the footer cache (no duplicate fetch, no license).
- Counts overlaid in `useBuildGroupByTableData`; `GroupHeaderWidget` renders `count (pct%)`; Ungrouped uses `valueNotFilledCount` with remainder %.
- Wired on Overview + Versions/Products. Free-text fields (no distribution) show count only; folder/hierarchy out of scope.
  
## Additional context

Closes #2083

group_by_counts


Innders (changes requested):
The idea and design is correct.

There is a bug when the group has no items due to a filter it will show the total count and not 0.

Image

Innders (changes requested):
Good, the counts are looking better now. Found one more bug where the counts are wrong.

- [x] Counts are wrong for "ungrouped", like with tags, assignees or any `list_of_strings` attribute.

Image

Image

Innders (approved):
Looks great thanks

PR #2114, opened 2026-07-02
De-sync state between tabs

### LINKED ISSUE #2113: YN-0891: Desync Localstorage States Between Tabs

It's impossible to use multiple browser tabs as often the state is all synced together. 

_No response_

### Additional context:

_This issue was automatically created from AYON ticket [YN-0891]( https://support.ayon.app/projects/Active_Clients/overview?project=Active_Clients&type=folder&id=3961ea5075fd11f1935115a0b092c2d9)

Innders:
## Description of changes 
Many different states like slicer selection were using localstorage. This would mean all browser tabs would be in sync and it would make it impossible to work with multiple tabs.

For these kind of states we use session storage instead that is scope to the browser tab. Refreshing the page will keep the state but it won't be synced to different tabs.

### Pages

The biggest culprit was the slicer but other areas have also been fixed:

- Project page slicer
- Details panel tab selection
- Viewer panel configs like mute
- My tasks project selection
- Bundle variant selection
- Addon settings form expansion

### DIFF
--- a/shared/src/containers/Slicer/context/SlicerContext.tsx
+++ b/shared/src/containers/Slicer/context/SlicerContext.tsx
@@ -8,7 +8,7 @@ import {
   useSlicerRowSelection,
 } from '@shared/containers/Slicer'
 import { SimpleTableRow } from '@shared/containers/SimpleTable'
-import { useLocalStorage } from '@shared/hooks'
+import { useSessionStorage } from '@shared/hooks'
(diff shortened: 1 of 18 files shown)


BigRoy:
- With this change, when opening a new tab - what's the view it defaults to? :)
- What happens if you "Duplicate Tab" in chrome?

martastain:
Duplicate tab duplicates the state, but then both tabs are independent.
When you just copy a link and paste to new tab, it seems no previous state is used.

I like it.

PR #2138, opened 2026-07-07
Filters: When a corrupt filter causes, tasks, products, entityListItems not to load show reset button

### LINKED ISSUE #1993: Filters: When a corrupt filter causes, tasks, products, entityListItems not to load show reset button
For whatever reason sometimes corrupt filters can cause an entity to fail to load.

Instead of just showing a generic error message; detect the error is related to the filters and show an error message. "{Entities} were unable to load due to a corrupt filters".

Then show a button that allows you to easily reset your working view filters.

filipvnencak:
## Description of changes
  
When a corrupt filter stops tasks, products, or list items from loading, show a clear message and a reset button instead of a raw error dump.
  
## Technical details

  
- New `filterError.ts` helper: detects backend malformed-filter errors by message signature, falls back to "any active filter" when a query fails.
- Shared table (`ProjectTreeTable.tsx`) and lists (`ListItemsTable.tsx`) render a friendly placeholder + reset button on a filter error.
- `useFetchOverviewData` now surfaces the task-list/folder/grouped query errors into `ProjectOverviewContext` (was grouping-only), so a corrupt task filter shows in Overview.
- Products provider (`VPProjectTableProvider.tsx`) passes `queryFilters` so the same detection works there.

## Additional context
  
Screenshot 2026-07-08 at 12 42 56

  Closes #1993


shared/src/containers/ProjectTreeTable/ProjectTreeTable.tsx
```diff
@@ -1416,16 +1429,29 @@ const TableBody = ({
...
-                onClick={onResetView}
-              />
-            )}
-          </EmptyPlaceholder>
+          {filterError ? (
+            <EmptyPlaceholder message={filterErrorMessage} icon="filter_alt_off">
+              {onResetView && (
+                <Button
+                  variant="filled"
+                  label="Reset working view"
```
Innders: Should this be "reset filters"?

shared/src/containers/ProjectTreeTable/ProjectTreeTable.tsx
```diff
@@ -1416,16 +1429,29 @@ const TableBody = ({
...
+                  icon="restart_alt"
```
Innders: using different icon to other pages, pick one of them

Innders (changes requested):
Great start but I think we could enhance this more to make it more useful for us.

- [x] Add button "Report issue" that opens a featurebase message with details about the error (page etc) and the fill error message. The button should only be visible if featurebase has loaded in (message bubble at the top).
- [x] Add button "Copy error message" that copies error message to the clipboard. The button should also have the tooltip of the error message itself.

Innders (approved):
Hard to test but code LGTM

PR #2141, opened 2026-07-07
Lists: Replace items

### LINKED ISSUE #2115: Lists: Add "Replace" option
## Story

Be able to swap out one entry in the list quickly with another.

### Problems

I've made a list of versions, but want to swap one them with a newer version.

### Proposal

Add a "Replace" action to the list, available when right clicking an entity in the list.

Review lists already allow this in the Card and Playlist views (not in the Table view somehow)

## TODO

- [ ] Add "Replace" functionality.

### Attachments

[YN-0896](https://support.ayon.app/projects/Active_Clients/overview?project=Active_Clients&type=folder&id=30dde7f0764b11f1a043fd04b70c1311)

Innders:
## Description of changes 
- Add a context menu option to replace list items in place with different selected items.
- You can replace multiple items at once.

https://github.com/user-attachments/assets/28828b00-e113-4f5f-a8e7-bddb57ebc715

### DIFF
--- a/src/pages/ProjectListsPage/components/ReplaceListItemsDialog.tsx
+++ b/src/pages/ProjectListsPage/components/ReplaceListItemsDialog.tsx
@@ -1,0 +1,108 @@
+import { FC, useMemo, useState, useCallback } from 'react'
+import { EntityPickerDialog, PickerEntityType } from '@shared/containers'
+import { useProjectContext, useProjectFoldersContext } from '@shared/context'
+import { useListsContext } from '../context'
+import { useListItemsDataContext } from '../context/ListItemsDataContext'
+import { useUpdateEntityListItemsMutation } from '@shared/api'
+import { toast } from 'react-toastify'
+
+const ReplaceListItemsDialog: FC = () => {
+  const { projectName } = useProjectContext()
+  const { selectedList } = useListsContext()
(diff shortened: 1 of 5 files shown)


src/pages/ProjectListsPage/components/ReplaceListItemsDialog.tsx
```diff
@@ -0,0 +1,108 @@
...
+    async (newEntityIds: string[]) => {
+      if (!itemIdsToReplace || !selectedList?.id) return
+      setIsReplacing(true)
+      //   calculate the position for the new items based on the first item to replace
+      const firstItemId = itemIdsToReplace[0]
+      const firstItem = listItemsMap?.get(firstItemId)
+      const position = firstItem?.position || 1
+
+      try {
+        // 1. remove old items
```
filipvnencak, line 35: We can temporary cache the items before delete if the adding new fails, the temp cached items can be rolled back

filipvnencak (changes requested):
just one small thing if add new api call fails the old one items is alredy deleted and not cached

src/pages/ProjectListsPage/components/ReplaceListItemsDialog.tsx
Innders: Good catch

PR #2146, opened 2026-07-08
Sumaries: Do not collect stats for disabled columns and by default they should be off 

### LINKED ISSUE #2143: Sumaries: Do not collect stats for disabled columns and by default they should be off
Getting stats is slow and puts a lot of load on the server. To reduce this load we should not load the stats unless absolutely required.

- [ ] Do not load stats if the column summary is disabled
- [ ] Disable the columns by default

filipvnencak:
## Description of changes 

Column summary stats are no longer requested for hidden columns. This cuts a large amount of server load on Overview, Versions/Products, and Lists, where every default-hidden attribute column was being aggregated.

## Technical details

- `buildMetricTargets` now resolves visibility with `checkColumnVisibility` (opt-in) instead of `columnVisibility[id] !== false`, and takes a `defaultColumnVisibility` arg.
- Overview and VP tables pass `defaultColumnVisibility` from the column settings context into both target builds.
- VP `product_base_type` extra target is gated through `checkColumnVisibility`.
- Lists page drops the manual defaults/override merge and passes `defaultColumnVisibility` directly.
- Targets now match the table's rendered visible columns exactly; revealing a hidden column still refetches via the existing `hasNewTargetFields` path.

## Additional context

Closes #2143

changes also in ynput/ayon-power-pack#76

### DIFF
--- a/shared/src/api/queries/columnStats/metricTargets.ts
+++ b/shared/src/api/queries/columnStats/metricTargets.ts
@@ -1,6 +1,7 @@
 import type { VisibilityState } from '@tanstack/react-table'
 import { StatsOperation } from '@shared/api/generated'
 import { checkColumnVisibility } from '@shared/containers/ProjectTreeTable/utils/checkColumnVisibility'
+import type { SummaryCalc, RowScope } from '@shared/containers/ProjectTreeTable/types/summaryTypes'
 
 export type MetricTarget = {
   field: string // column or dot-path for JSONB, e.g. 'status' / 'attrib.fps'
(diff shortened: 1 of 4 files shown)


Innders (changes requested):
Yes we should disable stats for hidden columns but this more about if the summary is disabled or not.

For example here status is disabled for both task and folders and yet it is still requesting the data.

- [x] If the summary is disabled for a column and is showing "–" then it should not be included in query.

Image

The defaults refer to the summary state not the column visibility. Many users will never use the summaries and so it's a waste to have them enabled by default. The user needs to explicitly set that they want to see an summary of a column.

This is how it show look for a brand new user.

Image

- [x] All summaries (except count) should be disabled unless explicitly turned on.

kalisp:
I am now seeing decrease of query size from before.

Innders (approved):

PR #2149, opened 2026-07-08
Simple table: Rewrite to support more in built features

Innders:
## Description of changes 
The SimpleTable has been upgrade to include more inbuilt features:

- Context builder - Use the new prop `rowContextMenuBuilders` to create context menus on the row. This makes it easier to define custom context menu items with access to the row context and data. All components use SimpleTable now use this for their context menus.
- `actions` buttons - Now you can define actions to show on the row, either during hover or always. This standardised the styling of the buttons.
- `badge` - a string or JSX component that goes at the end of the row item. Useful for counts and meta data for a row. The string uses a default faded colour.
- Renaming - The renaming logic is now built into the SimpleTable making it easy to support renaming of rows. This consolidated two different renaming features on the Projects List and Lists list components.

## Testing

Extensive testing needs to be done on:

- Projects list (on the home page were all the projects are listed)
- Hierarchy Slicer
- Lists table (on the lists page)

## Future
This will make it far easier to add new badges like counts in the slicer and context menus like on the hierarchy slicer.


shared/src/containers/SimpleTable/SimpleTableRowTemplate.tsx
```diff
@@ -132,10 +166,63 @@ export const SimpleTableCellTemplate = forwardRef<HTMLDivElement, SimpleTableCel
...
+          <div className="text" {...pt?.text}>
+            {parents && <span className="path">{parents.join(' / ')} / </span>}
+            <span className="value">{value}</span>
+          </div>
+        )}
+
+        {!isRenaming && (
+          <>
+            <Spacer className="spacer" />
+            {badge && (
```
filipvnencak: when there is 0 (falsy value) then for example Folder has zero children and it didn't match the style, so there is badge but text is white.

Image

shared/src/containers/SimpleTable/hooks/useRowKeydown.ts
```diff
@@ -28,9 +33,11 @@ function useRowKeydown<T>({ handleRowSelect, handleArrowNavigation }: UseRowKeyd
         // navigate and select
         const direction = event.key === 'ArrowDown' ? 'down' : 'up'
         handleArrowNavigation?.(direction, row, event)
+      } else if (event.key.toLowerCase() === 'r') {
```
filipvnencak: This renaming shortcut not working on homepage and on lists page

filipvnencak (changes requested):
With no power license is possible to create Project folder

shared/src/containers/SimpleTable/SimpleTableRowTemplate.tsx
Innders: Good catch

shared/src/containers/SimpleTable/hooks/useRowKeydown.ts
Innders: It wasn't a bug inside of the Simple table but two different ones from the parent components.

- Lists: calling wrong function
- Projects: adding extra `folder-` prefix to folder id

src/containers/ProjectsList/ProjectsSimpleTable.tsx
```diff
@@ -30,27 +40,122 @@ export const ProjectsSimpleTable: FC<ProjectsSimpleTableProps> = ({
...
+        navigate('/manageProjects/projectSettings?project=' + id)
+      }
+    },
+    [onSettingsClick, navigate],
+  )
+
+  const handleRename = useCallback(
+    (id: string) => {
+      const isFolder = !!parseProjectFolderRowId(id)
+      console.log(isFolder)
```
filipvnencak: leftover console

src/containers/ProjectsList/hooks/useProjectFolderActions.ts
```diff
@@ -89,14 +89,16 @@ export const useProjectFolderActions = ({
...
         handleOpenFolderDialog({ label: folder.label, ...folder.data }, folderId)
-    }
+      }
     },
     [folders, handleOpenFolderDialog],
   )
 
   const openRenameFolder = useCallback(
     (rowId: string) => {
+      console.log('on rename folder', rowId)
```
filipvnencak: leftover console

filipvnencak (approved):

PR #2150, opened 2026-07-08
Developer actions

Innders:
## Description of changes 
Fixes an issue where developer actions would not load properly when going into developer mode. It seemed like developer mode worked correctly for actions before a recent update but in reality it was always using production.

Now we make sure to pass the specific dev bundle name to `variant` to actually load the correct dev actions.

## Technical details

- Required moving bundles queries to shared, so testing all bundles actions is important.


antirotor (approved):
I can confirm developer actions are now displayed properly, and when developer mode is turned off, I get production actions as expected. It also works in different contexts.

filipvnencak (approved):

PR #2154, opened 2026-07-09
Summaries: fix column-summary grouping and skip disabled fetches

### LINKED ISSUE #2153: Overview/Products grouping , column summaries: several bugs (empty group expansion, stats errors, disabled Products summary, wrong tooltip
Bugs around grouping + column summaries (Overview / Products). Spans 3 repos: ayon-frontend, power-pack summaries remote, backend.

**1. Overview — groups expand to empty.** Group by Days (date), Description, Director Notes, etc. Groups show, but expanding shows no children. "Ungrouped" won't expand. Group `filter` from grouping endpoint doesn't select tasks for date/free-text fields.

**2. Products tab — can't enable Products summary.** Parent/Products summary scope stays disabled even with product tree / group by Product. `parentScopeApplicable`

**3. Products tab — "Error loading stats" grouping by Product Type (and Task Type).** `column "product_type" does not exist` from `GetVersionsColumnStats`. 

**4. Wrong tooltip on Products tab.** Says summary only in "Hierarchy or Products".  Products tab has no Hierarchy mode. Reword. Lives in power-pack remote.

**5. Stats query fires with everything off.** Disable all summaries,  query still runs, always requesting `name`. `buildMetricTargets`

**6. Breakdown entries (Filled/Empty) stop being clickable.** Work first time; after disabling Folder + Task scopes, clicking does nothing.

filipvnencak:
## Description of changes 

Fixes three column-summary issues: the Products summary can now be enabled on the Products tab, grouping products by product/task type no longer errors the stats footer, and no column-stats query is sent when every summary is switched off.

## Technical details

- Add `anySummaryActive` in `metricTargets.ts`; skip the column-stats query when no summary is active (Overview, Products, Lists).
- `parentScopeApplicable` now also true when `scopes` includes `product`, so the Products (parent) scope is selectable on the Products tab (`ProjectTreeTable.tsx`).
- Drop `productType - product_type` from `VERSION_FIELDS` in `groupCounts.ts`; the versions stats endpoint has no `product_type` column.

### DIFF
--- a/gen/graphql.schema.json
+++ b/gen/graphql.schema.json
@@ -2121,6 +2121,30 @@
             "deprecationReason": null
           },
           {
+            "name": "activityCategories",
+            "description": null,
+            "args": [],
+            "type": {
+              "kind": "NON_NULL",
+              "name": null,
+              "ofType": {
+                "kind": "LIST",
+                "name": null,
+                "ofType": {
+                  "kind": "NON_NULL",
+                  "name": null,
+                  "ofType": {
+                    "kind": "SCALAR",
+                    "name": "String",
+                    "ofType": null
+                  }
+                }
+              }
+            },
+            "isDeprecated": false,
+            "deprecationReason": null
+          },
+          {
             "name": "allAttrib",
             "description": null,
             "args": [],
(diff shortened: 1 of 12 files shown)


Innders (changes requested):
- [x] When no columns are active it becomes impossible to enable name column.

Image

- [x] Is the name column on by default? It seems the whole query is being skipped, the query should only be skipped if the user explicitly turns off everything including name column.

Innders (changes requested):
There is an issue where loading the page for the first time can cause the name column to be disabled and unclickable when all other columns are disabled.

PR #2162, opened 2026-07-10
YN-0725: Copy Representation File Path via context menu

### LINKED ISSUE #2036: YN-0725: Copy Representation File Path via context menu

**Summary**

Add a frontend-only way to copy a representation file path from the UI via right-click context menu.

**Problem**

The current implementation was started as a backend/action-driven flow, but it was concluded that this can be handled on the frontend instead. A dialog-based approach was explored earlier, but the preferred UX is a context menu on the representation itself.

**Proposed UX**

When a user right-clicks a representation, show a **Copy path** submenu with:

-   **Copy Windows path**
    
-   **Copy Mac path**
    
-   **Copy Linux path**

**Technical implementation: constructing the representation path**

-   Read the representation’s **first file entry** and take its **rootless path**.
    
-   Determine which **root token** the representation/template uses (for example, `work` or another configured root key).
    
-   From the representation/anatomy data, look up the OS-specific root values for that same root token:
    
    -   Windows root
        
    -   Linux root
        
    -   Mac root
        
-   Construct the final path by prepending/replacing with the selected OS root value against the rootless path.
    
-   Expose three copy actions in the context menu:
    
    -   Copy Windows path
        
    -   Copy Linux path
        
    -   Copy Mac path
        

**Notes**

-   Do not use the published full path directly, because that may reflect only the publish platform path; use the **rootless** form as the base.
    
-   Do not include site overrides in this implementation; only generic OS variants are needed.
    
-   This should be doable entirely on the frontend.

### Additional context:

_This issue was automatically created from AYON ticket [YN-0725]( https://support.ayon.app/projects/Active_Clients/overview?project=Active_Clients&type=folder&id=d88d450fa6e740d7ac85aff3db9267af)

filipvnencak:
## Description of changes 

Adds a Copy path submenu to the representation right-click menu, with Copy Windows / Mac / Linux path actions that resolve the file path for each OS.

## Technical details

- New `replaceRoot` util swaps `{root[name]}` tokens for per-OS anatomy root values (`shared/src/util/replaceRoot.ts`).
- `RepresentationsList` builds per-OS root maps from `useGetProjectAnatomyQuery`, resolves the first file's rootless path, copies via `copyToClipboard`.
- Fragment `DetailsPanelRepresentationFragment` now fetches `files { path }`; carried through `versionsToRepresentations`.
- `WorkfileDetail` reuses the shared `replaceRoot` (removed its local copy).
- Uses generic OS roots only; no site overrides.

## Additional context

Closes #2036

Screenshot 2026-07-10 at 13 37 48

### DIFF
--- a/shared/src/api/generated/graphql.ts
+++ b/shared/src/api/generated/graphql.ts
@@ -1812,7 +1812,7 @@ export type GetDetailsPanelVersionQueryVariables = Exact<{
 }>;
 
 
-export type GetDetailsPanelVersionQuery = { project: { projectName: string, code: string, version: { id: string, version: number, name: string, author: string | null, status: string, tags: Array<string>, updatedAt: unknown, createdAt: unknown, thumbnailHash: string, thumbnailId: string | null, hasReviewables: boolean, parents: Array<string>, allAttrib: string, product: { id: string, name: string, productType: string, folder: { id: string, name: string, label: string | null, path: string | null, folderType: string }, latestVersion: { version: number } | null }, task: { id: string, name: string, label: string | null, assignees: Array<string>, taskType: string } | null, representations: { edges: Array<{ node: { id: string, name: string, fileCount: number } }> } } | null } };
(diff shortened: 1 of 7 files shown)


Innders (approved):
LGTM but I would look someone who knows what they are doing to review @BigRoy

BigRoy (approved):
Seems to work - it's a bit hidden away and we'll likely need to tell the client where it's at and then when rolled out confirm they are happy enough with it. I'd personally prefer if somehow we would be able to do this with fewer clicks, but at the same time - we don't want this to be in your face the majority of the time so hard to juggle where it fits.

Let's see what client says and knows about it for a while.

PR #2177, opened 2026-07-15
Details Panel: More menu delete option

### LINKED ISSUE #2172: Delete: Remove dangerous selection mechanic
Deleting entities from the table has a dangerous mechanic.

The deletion will use the **row** selection insead of the **cell** selection. This means you can right click on some cells to delete and end up deleting something else.

Image

- [ ] Completely ignore the row selection when deleting entities in the table. The only selection used should be the cell selection.
- [ ] Ignore the **row** selection for right click context menus (if there are any other).

### LINKED ISSUE #2173: Details Panel: More menu delete option
Add a red delete menu item to the bottom of the more menu for an entity. This should work for single and multiple selected entities.

Use the same confirmation dialog for deleting entities as the overview page.

Image

- [ ] Add menu item at the bottom with red danger styling
- [ ] Support multiple entities with label change
- [ ] Create a generic hook/component that can handle deleting and confirming any entity and can be used across the app.

Related to #2172

filipvnencak:
## Description of changes

Adds a red Delete option to the details panel more menu for single and multiple entities. All entity deletes (folder, task, product, version) now share one confirmation dialog and delete flow, and remove rows optimistically with rollback on failure.

## Technical details

- New `DeleteEntitiesProvider` + `useDeleteEntitiesContext` in `shared/src/context/DeleteEntitiesContext.tsx`, mounted app-wide in `app.tsx`.
- One batched `operations` request per project, ops sorted children-first; force fallback for folders with children.
- Confirmation dialog extracted to `shared/src/components/DeleteEntitiesConfirm`; `pluralize` moved to `shared/src/util`.
- Overview/lists table (`useDeleteEntities`) and VP menu (`useVPContextMenu`) migrated onto the shared flow.
- Optimistic delete: `updateOverview`/`patchProducts`/`patchVersions` remove deleted rows and undo on error; reconciliation moved to `invalidatesTags`.
- Delete target scoped per entry point: the details-panel more-menu deletes the panel's own displayed entity (never an outside selection, which can diverge from what the panel shows); the table right-click menu and toolbar delete the cell selection, falling back to the checkbox row selection only when no body cells are selected.

## Additional context

Closes #2173


shared/src/api/queries/overview/patchVersions.ts
```diff
@@ -194,7 +194,11 @@ export const patchVersions = (
...
-  // Step 3: Invalidate all affected caches to trigger refetching
-  // This will automatically refetch with filters and update calculated attributes
-  dispatch(injectedVersionsPageApi.util.invalidateTags(tags))
+  // Step 3: Invalidate affected caches to trigger refetching.
+  // Skip for deletes — the request is still in flight, an immediate refetch would
+  // resurrect the optimistically removed rows; reconciliation happens on success.
+  const nonDeleteOps = versions.filter((op) => op.type !== 'delete')
+  if (nonDeleteOps.length > 0) {
+    dispatch(injectedVersionsPageApi.util.invalidateTags(getVersionTags(nonDeleteOps)))
+  }
```
Innders: Not sure I totally understand these changes? Aren't all requests still in flight, why would it be specific to delete? If that's the case maybe we shouldn't invalidate at all.

shared/src/api/queries/overview/updateOverview.ts
```diff
@@ -703,17 +718,46 @@ const operationsApiEnhancedInjected = operationsEnhanced.injectEndpoints({
...
+
+        if (invalidateList.has('folder')) deletedEntityTags.push({ type: 'folder', id: 'LIST' })
+        if (invalidateList.has('task')) {
+          deletedEntityTags.push(
+            { type: 'overviewTask', id: 'LIST' },
+            { type: 'overviewTask', id: projectName },
+          )
+        }
+        if (invalidateList.has('product')) deletedEntityTags.push({ type: 'product', id: 'LIST' })
+        if (invalidateList.has('version')) deletedEntityTags.push({ type: 'version', id: 'LIST' })
```
Innders: I think we don't need this, if the delete doesn't work for some reason it's not the end of the world, we don't need to invalidate and refresh everything.

shared/src/containers/DetailsPanel/DetailsPanel.tsx
```diff
@@ -382,6 +383,7 @@ DetailsPanelProps) => {
       versionsInputRef={versionsInputRef}
       canUploadVersion={canUploadVersion}
     >
+      <DeleteEntitiesProvider>
```
Innders: The whole app is wrapped bu the provider, is this correct?

Innders (changes requested):
- [x] Label should include the entity type "Delete task" "Delete 3 tasks" "Delete version"

Image

shared/src/api/queries/overview/patchVersions.ts
filipvnencak: Hmm, yeah you're right,  it's fine for updates, but not needed for deleted entities. Optimistic removal already takes them out, so I'll skip the invalidate there.

shared/src/containers/DetailsPanel/DetailsPanel.tsx
filipvnencak: Good catch, it's redundant.

Innders (changes requested):
- [x] Deleting still uses row selection. Here I should only be deleting the cell selected tasks.

Image

Innders (changes requested):
It looks like the details panel deletion is based on selection. This is very dangerous because the details panel doesn't always match the selection.

- [x] Details panel should use current entity data for deletion, not outside selection

Image

filipvnencak:
Screenshot 2026-07-20 at 16 07 54

filipvnencak:
Details panel more-menu  -> row selection
 Table right-click -> cells (fallback to rows only if zero cells)
 Toolbar "Delete selected" -> cells (fallback to rows only if zero cells)

Innders (changes requested):
> Details panel more-menu -> row selection

The details panel should not use table/cell/card selection at all. There are times where the details panel does not match the current selection.

The entities to delete must come directly from inside `DetailsPanel.tsx` we know 100% that we are deleting those entities in the panel.

filipvnencak:
> The details panel should not use table/cell/card selection at all. There are times where the details panel does not match the current selection.


The more-menu delete targets entityDetailsData, so it always matches what's on screen, not the table selection.


Screenshot 2026-07-21 at 09 45 20

Innders (approved):

PR #2182, opened 2026-07-16
Reload data button

### LINKED ISSUE #2160: YN-0913: Proper reload button for all pages

There is a reload button on the overview page that makes it easy to just reload the data in the table.

&nbsp;

This should be universal across all pages.

&nbsp;

It should be a separate button outside of the search bar and it should be highlighted when there are new updates.

&nbsp;

New updates like created folders, tasks, products, versions should not be automatically streamed in the table and should just trigger the reload button to be highlighted. 

### Additional context:

_This issue was automatically created from AYON ticket [YN-0913]( https://support.ayon.app/projects/Active_Clients/overview?project=Active_Clients&type=folder&id=00da08807c4611f1ada54d6c1b976b20)

Innders:
# Description of Changes

- Adds a reload button next to the Search filter bar on most project pages.
- Reload button, reloads all data on the page, including things like the summaries.
- Rewrites the reloading logic of the pages to include unused caches.
- Sets more aggressive caching times for the queries (10 seconds vs default 60).
- In theory this button is useless right now because all changes should always be streamed in; but it is a good fallback and it lays the groundwork for future syncing updates.

# Testing

## Setup
There is now a new env variable `VITE_DISABLE_WS` that will disable realtime updates. This should be set to `true` to make it easier to make changes on a different account and then test reloading, to see if the changes show using the reload button.

## Steps
1. Open a page in two browser windows with different accounts **A** and **B**.
2. On **A**, do some different search queries and navigate around.
3. On **B**, make some changes to some entities.
5. On **A** click the reload button to see if the changes are shown.
6. On **A** go back to previous filters to check if the changes propagated across all caches.

### Overview
- Tasks only (no grouping)
- Hierarchy
- Group by folders
- Group by attribute/field

### Task progress
- Default view

### Products
- Versions only (no grouping)
- Group by Product
- Group by attribute/field

### Lists/Review
- Default view


martastain (approved):
I wasn't able to break it. Everything seems to work apart from out-of-scope issues mentioned in the chat

shared/src/api/queries/overview/getOverview.ts
```diff
@@ -255,7 +264,7 @@ const injectedApi = enhancedApi.injectEndpoints({
 
           const allNewTasks: EditorTaskNode[] = []
           const BATCH_SIZE = 100
-          const TASK_PER_FOLDER = 20
+          const TASK_PER_FOLDER = CACHE_TIME
```
filipvnencak: This TASK_PER_FOLDER is not used for cashing but for pagination, before it was 20 now is 10

shared/src/api/queries/actions/refreshActiveAndPurgeOthers.ts
```diff
@@ -0,0 +1,77 @@
...
+    currentArgs: unknown,
+    { refreshOtherActiveQueries = true }: RefreshOptions = {},
+  ) =>
+  (dispatch: ThunkDispatch<any, any, UnknownAction>, getState: () => any) => {
+    const queryApi = api as any
+    const state = getState()[api.reducerPath] as ApiState | undefined
+
+    try {
+      const primaryPromise = dispatch(
+        queryApi.endpoints[endpointName].initiate(currentArgs, { forceRefetch: true }),
```
filipvnencak: Missing `subsribe: false` in that case memory leaking

shared/src/api/queries/actions/refreshActiveAndPurgeOthers.ts
```diff
@@ -0,0 +1,77 @@
...
+
+export const refreshOtherActiveQueries =
+  (endpointName: string, currentArgs: unknown) =>
+  (dispatch: ThunkDispatch<any, any, UnknownAction>, getState: () => any) => {
+    const queryApi = api as any
+    const state = getState()[api.reducerPath] as ApiState | undefined
+
+    return Promise.all(
+      getOtherActiveQueryArgs(state, endpointName, currentArgs).map((args) =>
+        dispatch(queryApi.endpoints[endpointName].initiate(args, { forceRefetch: true })).unwrap(),
```
filipvnencak: `unwrap()` Didn't unsubscribe

filipvnencak (changes requested):

shared/src/api/queries/overview/getOverview.ts
Innders: Good spot

PR #2183, opened 2026-07-16
Lists: Add to lists dialog search

### LINKED ISSUE #2101: Lists: Add to lists dialog search
## Story

There are so many lists that it is hard to find the correct one when using the context menu to add to it.

If I have 100 lists and it's very hard and time consuming to find the right one.

### Problems

The context menu is not built for 100s of items.

### Proposal

- Completely remove sub-menu in the context
- Clicking add to list in the context - opens a dialog.
- The add to list dialog is the lists table with folders and search
- Select the lists you want to add to and the confirm
- Double click a list to auto confirm adding to it

filipvnencak:
Lists: Add-to-list dialog with search

## Description of changes 

Replaces the nested "Add to list" context submenu with a searchable dialog that reuses the Lists table, so it works with hundreds of lists. Select one or many lists and confirm, or double-click a list to add instantly. Same flow for Review -Add to review list.

## Technical details

- New `AddToListDialog` reuses `ListsTable` in a `picker` mode (local selection, no URL); opened via `openAddToListDialog` in `EntityListsContext`. `useBuildListMenuItems` now opens it instead of building a nested submenu.
- Picker hides create-folder/delete/filter, keeps searchable create-list (prepopulates the selected entities); row context menu is an allowlist.
- Lists scoped by entity type + editor access (`accessLevel >= 20`); picker ignores the page's saved list filters; current list excluded on the Lists item table.
- `SimpleTable`: ctrl/meta-click multi-select now toggles from live table state (was dropping other rows via a stale `rowSelection` closure). Shared change — affects all `SimpleTable` consumers; may need a shared version bump for the powerpack remote.
- Dialog menu id via `useId` so it doesn't collide with the sidepanel menu; dialog uses `block-shortcuts` to stop key leaks.

## Additional context

Screenshot 2026-07-16 at 11 47 30

Closes #2101


Innders (changes requested):
Working so much better already!

- [x] We have the room so make the dialog a lot bigger.

Image

- [x] More menu needs to use the new dialog as well

Image

- [x] Task progress uses old menu
- [x] OLD hierarchy.jsx uses old menu (replace menu with disabled version that is "deprecated")
- [x] OLD taskList.jsx uses old menu (replace menu with disabled version that is "deprecated")
- [x] OLD Products.jsx use old menu (replace menu with disabled version that is "deprecated")

- [x]  Once all of the old context menu refs are removed, cleanup and remove the "old" way of doing it. We do not need to build the menu and submenus anymore. The goal is to remove `useGetListsData` so that we do not need to load lists on page load. The list loading is now handled by the ListsTable when the picker opens.

Innders (changes requested):
- [x] Do not filter the table based on entity type, always show everything but disable the lists you cannot add it to.
- [x] `GetLists` and `listFolders` requests are made on page load when not on lists page because of the slicer. Only load these if the slicer type is lists. We should not see these requests unless we need the lists data.

src/pages/ProjectListsPage/hooks/useListContextMenu.ts
```diff
@@ -121,6 +128,18 @@ const useListContextMenu = (extraBuilders: ListRowContextMenuBuilder[] = []) =>
...
+      if (picker) {
+        return [
+          {
+            label: 'Select all lists',
+            icon: 'checklist',
+            hidden: !selectedFolderIds.length,
+            command: () => selectAllLists({ rowIds: selectedRows }),
+          },
+        ]
+      }
```
Innders: Remove the context menu completely for picker mode

Innders (changes requested):
Looking good, just some final finishing touches 💅 

- [x] Completely remove context menu for picker mode.
- [x] Remove the ~~strikethrough~~ on the disabled items
- [x] For disabled items replace the count badge with disabled message. `Type mismatch`

Innders (approved):

PR #2190, opened 2026-07-21
Products: Ensure table cells are focused properly

Innders:
## Problem
The products page table focus was buggy and often the focus would go to `body`. This would cause all sorts of issues like the shortcuts not working and the context menu not closing properly.

## Description of changes 
There were multiple factors that would cause this specifically on the products page.

**Keyboard Navigation & Focus Management**
- Explicit Cell Focus on Click: Forced body-cell focus on click after mousedown prevents native focus behavior (ProjectTreeTable.tsx).
- Scoped Keyboard Navigation: Filtered out headers, menus, and unrelated page elements from keyboard navigation sequences (useKeyboardNavigation.ts).
- DOM Focus Alignment: Synchronized actual DOM focus with logical selection state when navigating via Arrow keys and Tab (useKeyboardNavigation.ts).
- Header-Safe Focus Scoping: Restricted VP grid-to-table focus targets strictly to body cells, preventing draggable headers from being focused (VPFocusContext.tsx).

**Performance Optimization**
- Table Column Memoization: Stabilized the excluded-column list and memoized extra-column definitions to prevent unnecessary table-column reconstructions (VPTable.tsx).

### DIFF
--- a/shared/src/containers/ProjectTreeTable/hooks/useKeyboardNavigation.ts
+++ b/shared/src/containers/ProjectTreeTable/hooks/useKeyboardNavigation.ts
@@ -22,7 +22,8 @@ export default function useKeyboardNavigation() {
   const focusCellElement = useCallback(
     (cellId: string) => {
       focusCell(cellId)
-      requestAnimationFrame(() => document.getElementById(cellId)?.focus())
+      console.log(cellId)
+      requestAnimationFrame(() => document.getElementById(cellId)?.closest('td')?.focus())
     },
     [focusCell],
   )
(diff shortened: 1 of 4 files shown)


shared/src/containers/ProjectTreeTable/hooks/useKeyboardNavigation.ts
```diff
@@ -19,12 +19,18 @@ export default function useKeyboardNavigation() {
   const { setEditingCellId, editingCellId } = useCellEditing()
   const { setSelectedEntity } = useDetailsPanelEntityContext()
 
+  const focusCellElement = useCallback(
+    (cellId: string) => {
+      focusCell(cellId)
+      requestAnimationFrame(() => document.getElementById(cellId)?.focus())
```
filipvnencak: should be       

`requestAnimationFrame(() => document.getElementById(cellId)?.closest('td')?.focus())`

filipvnencak (changes requested):

shared/src/containers/ProjectTreeTable/hooks/useKeyboardNavigation.ts
Innders: Thanks

PR #2199, opened 2026-07-23
Overview: Table rendering unstable causing links picker to close on streaming updates

Innders:
## Description of changes 
Some object that went into the columns memo builder was unstable, causing the inner table cells to remount when there were data changes, like real time streamed updates.

This would cause render and state stability within the table like flickering dropdowns and dialogs closing.

This was particularly painful on busy servers where there were constant updates streamed in, making picking links a lot more difficult when the dialog would close often.

## Technical details

Uses stringify to stabilize object references for the useMemo deps. (a bit of hack).


Innders:
Branch included another branch by accident

PR #2205, opened 2026-07-27
Filters: text fields empty and not empty

### LINKED ISSUE #2159: YN-0912: Filters: Text fields empty and not empty

Sometimes you just want to find entities that has some text or no text in a field.

&nbsp;

For all text attributes across all entity filters, add an option to search for "empty" or "not empty".

&nbsp;

Changes to SearchFilter might need to br made. 

### Additional context:

_This issue was automatically created from AYON ticket [YN-0912]( https://support.ayon.app/projects/Active_Clients/overview?project=Active_Clients&type=folder&id=63a71d007c4511f1ada54d6c1b976b20)

filipvnencak:
## Description of changes 

Adds "Has {field}" / "No {field}" options to attribute filters, so entities can be found by whether a field is filled in at all. Available on Overview, Versions & Products and Lists.

## Technical details

- `useBuildFilterOptions.tsx`: enables the options for string, integer and float attributes.
- Options stay hidden for booleans ("No" already matches unset rows) and for attributes with an inheritable project default (fps, priority, resolution…), where the value always resolves and nullness would never match.
- `clientFilterToQueryFilter.ts`: text nullness compiles to a composite group (`isnull OR eq ''` / `notnull AND ne ''`) because cleared text attributes are stored as `""`, not null.
- `queryFilterToClientFilter.ts`: `tryMergeTextNullness` restores the chip so saved views round-trip.
- Tasks Progress is unaffected — it builds filter options without a config, so the options never render there.

## Additional context

Closes #2159

### DIFF
--- a/shared/package.json
+++ b/shared/package.json
@@ -132,7 +132,7 @@
   "dependencies": {
     "@tanstack/match-sorter-utils": "8.19.4",
     "@tanstack/react-virtual": "3.13.23",
-    "@ynput/ayon-frontend-shared": "^0.3.43",
+    "@ynput/ayon-frontend-shared": "^0.3.51",
     "@ynput/ayon-react-components": "^1.18.8",
     "axios": "1.16.1",
     "clsx": "2.1.1",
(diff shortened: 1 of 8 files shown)


Innders (changes requested):
Works really well! 

Found a small bug

- [x] When clicking date filter chip with no/has value, the date range picker opens.
https://github.com/user-attachments/assets/e5ebf6c2-ca72-493d-84e6-993678c60dd6

Innders (approved):

PR #2206, opened 2026-07-28
Delete: deleting entities should require typed confirmation

### LINKED ISSUE #2193: Delete: deleting entities should require typed confirmation
## Story
It's so easy to delete hundreds of folders and tasks within seconds.

### Proposal
When deleting an entity, prompt the user to first type something out that matches the deletion code before they can delete it. This should work the same way are deleting users.

Image

### Extra fixes

- [ ] When deleting hundreds of items the description pushes things off screen.

Image

filipvnencak:
## Description of changes 

Deleting entities now asks for typed confirmation. A single entity requires its name. Large multi-selections require typing the count per entity type. Also fixes the dialog overflowing off screen with many items.

## Technical details

- New `DeleteEntitiesConfirmDialog` replaces primereact `confirmDelete` in `DeleteEntitiesContext`; `deleteEntities()` API unchanged, no caller changes.
- Multi-selections over 1  show one numeric input per type; at or below, no typing.
- Counts include folder-cascade children from delete-info (`buildExpectedCounts`); falls back to selection only if the fetch fails.
- Toasts and force-delete fallback behave as before.

## Additional context

Closes #2193
Screenshot 2026-07-28 at 16 07 41
Screenshot 2026-07-28 at 16 07 13
Screenshot 2026-07-28 at 16 07 22


Innders (changes requested):
Love having to type the number in of each entity type!

- [x] When there are more than 1 entity being deleted, always show the numbers confirm even if it's a single selection

- [x] Replace the placeholders in the inputs with `Number of folders` etc. This will force the user actually think about what the numbers means rather than blindly just copying the placeholder. (You might need to increase the input width slightly).

Innders (changes requested):
Sorry I meant just remove the count from the input placeholder. Otherwise it's actually pretty hard to find the right counts 😅 

- [x] Add counts next to entity label

Image

Innders (changes requested):
Love the idea of typing out the different numbers.

I found one bug though.

When needing to force delete a folder (has products or versions in it) an error toast saying: "Unable to delete a folder with products or tasks." shows but we have actually deleted them correctly with force.

What seems to be happening is that the first attempt to delete throws the error because force wasn't used and this triggers a toast error that is now invalidate.

- [x] Ensure when force deleting that the previous delete error toast does not show.

filipvnencak:
Force confirmation also moved off `window.confirm` onto the styled dialog.

Screenshot 2026-07-31 at 11 53 18

Innders (approved):
All working now and thanks for moving the force to use proper confirm dialog.

PR #2209, opened 2026-07-29
Projects: fix code generation for regexes with underscores

### LINKED ISSUE #2187: Settings: Project code regex not working correctly
Codes are not generated properly from the project name using the project code regex.

https://discord.com/channels/517362899170230292/1528702389060767766

filipvnencak:
## Description of changes 

Fixes empty project code in the new project dialog when the configured code regex contains a literal underscore, e.g. `(?:\d{5}_)[A-Za-z]{3}` on `12345_Some_Project`.

## Technical details

- `createCode` replaced `_` with spaces before matching, so patterns with a literal `_` never matched (`NewProject.jsx`).
- Patterns containing `_` now run against the raw name; all others keep the spaced name, so `\b` initials and the default `^.{0,3}` behave exactly as before.
- Match processing extracted into a `matchCode` helper; capture-group handling and space stripping unchanged.

## Additional context

Closes #2187

### DIFF
--- a/src/pages/ProjectManagerPage/NewProjectDialog/NewProject.jsx
+++ b/src/pages/ProjectManagerPage/NewProjectDialog/NewProject.jsx
@@ -28,6 +28,26 @@ const PROJECT_STATES = [
   { value: 'production', label: 'Production' },
 ]
 
+// prefer unicode mode for \p{...} patterns, but some patterns legal without it (e.g. `\-` outside a class) only compile as non-unicode
+const buildCodeRegex = (pattern) => {
+  try {
+    return new RegExp(pattern, 'gu')
+  } catch (error) {
+    return new RegExp(pattern, 'g')
+  }
+}
+
+const matchProjectCode = (value, pattern) => {
+  const matches = [...value.matchAll(buildCodeRegex(pattern))]
+  if (!matches.length) return ''
+
+  return matches
+    .map((match) => (match.length > 1 ? match.slice(1).filter(Boolean).join('') : match[0]))
+    .join('')
+    .replaceAll(' ', '')
+    .replace(/^_+|_+$/g, '')
+}
+
 const DialogContent = styled.div`
   width: 100%;
   height: 100%;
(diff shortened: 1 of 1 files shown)


Innders (changes requested):
There are a few edge cases with createCode and project name sanitization. Here is a breakdown of the updates we need to make to ensure predictable regex matching across all server configurations.

## 1. Refactor createCode for Predictable Regex Execution
The current createCode implementation mutates project names under the hood (replacing _ with spaces and stripping spaces post-match). This causes unexpected bugs:

Length truncation: Hard length limits like ^.{5} fail because spaces get stripped after matching, returning fewer characters than requested.

Silent group drops: Multiple capture groups (e.g., ([A-Z]+)_(\d+)) silently throw away group 2 and only return group 1.

Unicode crashes: Lacks the u flag in RegExp, which causes errors when server patterns use modern features like \p{L}.

Proposed Fix: Update createCode to evaluate regexes cleanly against the raw name string and concatenate all capture groups if present:

```js
const createCode = (name, regexPattern) => {
  if (!regexPattern || !name) return ''

  try {
    // Add 'u' flag to support Unicode ranges & emojis properly
    const regex = new RegExp(regexPattern, 'gu')
    const matches = [...name.matchAll(regex)]

    if (!matches.length) return ''

    return matches
      .map((match) => {
        // If capture groups exist, combine group 1, group 2, etc.
        if (match.length > 1) {
          return match.slice(1).filter(Boolean).join('')
        }
        // Otherwise return full match
        return match[0]
      })
      .join('')
  } catch (error) {
    console.warn('Invalid regex pattern for project code', error)
    return ''
  }
}
```
## 2. Standardize Input Sanitization in handleNameChange
Currently, typing into the Project Name field directly only replaces spaces with underscores:

```js
// Current:
const handleNameChange = (newName) => {
  const nextName = newName.replace(/\s/g, '_')
  ...
}
If a user manually types special characters like hyphens (PRJ-99), createCode will attempt to parse hyphens before validate() eventually blocks submission.

Proposed Fix: Apply consistent sanitization so createCode always receives valid snake_case input:

JavaScript
const handleNameChange = (newName) => {
  // Keep input sanitized consistent with createName logic
  const nextName = newName
    .replace(/\s+/g, '_')
    .replace(/[^a-zA-Z0-9_]/g, '_')
  
  setName(nextName)
  if (!codeSet || code === '') {
    const newCode = createCode(nextName, codeRegex)
    setCode(newCode)
  }
}
```
## 3. Note on Server-Side Regexes (project_code_regex)
Since AYON project names strictly enforce snake_case (alphanumeric + underscores), server patterns must target underscores _ rather than hyphens - or spaces.

❌ (?<=-)\d+ (Fails because hyphens are sanitized out of name)

✅ (?<=_)\d+ (Works correctly on PRJ_99_Alpha -> 99)

filipvnencak:
Thanks, took most of it.

**Taken:** raw matching (dropped my `includes('_')` heuristic), all capture groups concatenated, `u` flag. One caveat on `u`: forcing it breaks patterns legal today (`\-{1}` outside a class throws-> empty code), so it's `try 'gu'`-> on `SyntaxError` retry `'g'`.

**Point 2 skipped — unreachable.** `handleNameChange` can't receive `PRJ-99`; `LabelWithNameField` normalizes via `createName` first (`LabelWithNameField.tsx:81` auto-derived, `:95` manual commit). A second weaker sanitizer would give two divergent rules.

**Point 3 agreed** - patterns should target `_`; worth putting in docs.

Verified (no unit runner, node script over both impls):

Only intentional diff vs `develop`: `^.{0,3}` on `A_B_C` → `A_B` (was `AB`), inner `_` now preserved.

Out of scope: `\p{Lu}` on `Žlty_Kôň_Ide`(Slovak diacritics) -> `ŽKI`, rejected by both FE and backend (ASCII-only). Diacritics stripping = separate change.

Innders (approved):
All the tests I did work now for it, thanks!

PR #2217, opened 2026-07-31
YN-0426: RFE: Inbox: Grouping & filtering

### LINKED ISSUE #1765: YN-0426: RFE: Inbox: Grouping & filtering
The inbox should have the search filter so you can filter the items by activity filters, like in the activity feed.

The inbox uses a special inbox resolver that can't have filtering on it. The compromise is to use the `activities` resolver on a per project basis. This will require the user to first select a project before they can filter the inbox.

Once a project is selected the `activities` resolver should be used instead of the `inbox`. Then the filtering and search arguments should be used.

- [x] Add the project list on the left of the inbox page so that you can filter inbox by project.
- [x] Once a project is selected, enable the filter bar, using `activities` in graphql instead of `inbox`.

filipvnencak:
## Description of changes 

Adds a projects panel to the inbox. Picking a project enables search and filtering for it, plus an unread-only toggle that works in both modes. Each tab keeps its own saved views.

## Technical details

- Project mode queries `project.activities` instead of `inbox`, rebuilding the inbox predicates as a QueryFilter in `util/inboxFilter.ts`. No backend change.
- `getProjectInbox.ts` is an infinite query. Its websocket top-up is debounced to one request per second and walks back by cursor when a burst outruns a page.
- Both list queries are cached per filter, so `updateInbox.ts` patches by walking the cache. The backend `all` flag is per-project, so the cross-project cache only loses that project's rows.
- Clear all sends the visible reference ids while a filter or the unread toggle is active, since `all` ignores filters server-side.
- The selected project lives in the view and mirrors to `?inboxProject=`. `project=` is left to the details panel, which rewrites it on every entity it opens.
- `ProjectsList` gains `allowEmptySelection` and `enableClickToDeselect`.

## Additional context

Closes #1765 · YN-0426

Screenshot 2026-08-13 at 14 49 53

Related: ynput/ayon-backend#1055
Also:  ynput/ayon-backend#1079


src/pages/InboxPage/components/InboxSearchFilter.tsx
```diff
@@ -0,0 +1,216 @@
...
+    dateRange.wrapFilterChange(newFilters, localFilters, setLocalFilters)
+
+  return (
+    <Wrapper
+      className={isDisabled || isLoading ? 'disabled' : undefined}
+      data-tooltip={
+        isDisabled
+          ? 'Searching and filtering works on one project at a time. Select a project on the left to enable it.'
+          : undefined
+      }
```
Innders:
```suggestion
      data-tooltip={isDisabled ? 'Select a project to enable filtering' : undefined}
```

src/pages/InboxPage/Inbox/Inbox.tsx
```diff
@@ -484,85 +584,107 @@ const Inbox = ({ filter }: InboxProps) => {
...
+            <ProjectsList
+              selection={selectedProject ? [selectedProject] : []}
+              onSelect={(ids: string[]) => {
+                // clicking the selected row clears it, which returns to the cross-project inbox
+                if (!ids.length) return setSelectedProject(null)
+                // folder rows are group headers, their ids are not project names
+                const projectName = ids.find((id) => !parseProjectFolderRowId(id))
+                if (projectName) setSelectedProject(projectName)
+              }}
+              allowEmptySelection
```
Innders, line 608: - Hide the "add project" shortcut button at the top.
- Hide the Archive and Delete menu items

src/pages/InboxPage/Inbox/Inbox.tsx
```diff
@@ -484,85 +584,107 @@ const Inbox = ({ filter }: InboxProps) => {
...
+          <SplitterPanel size={18} style={{ minWidth: 180 }}>
+            <ProjectsList
+              selection={selectedProject ? [selectedProject] : []}
+              onSelect={(ids: string[]) => {
+                // clicking the selected row clears it, which returns to the cross-project inbox
+                if (!ids.length) return setSelectedProject(null)
+                // folder rows are group headers, their ids are not project names
+                const projectName = ids.find((id) => !parseProjectFolderRowId(id))
+                if (projectName) setSelectedProject(projectName)
+              }}
```
Innders: This is enough code to have it's own handler function further up

src/pages/InboxPage/components/InboxSearchFilter.tsx
```diff
@@ -0,0 +1,216 @@
...
+    cursor: not-allowed;
+
+    & > div {
+      pointer-events: none;
+    }
+  }
+`
+
+const REASON_VALUES = [
+  { id: 'mention', label: 'Mentioned', icon: 'alternate_email' },
```
Innders, line 35: This is a cool idea! One small fix for it though, it's including assignment activities "You were added to task". It should only be comments.

src/services/inbox/getProjectInbox.ts
```diff
@@ -0,0 +1,119 @@
+import { gqlApi } from '@shared/api'
+import type { GetInboxMessagesQuery } from '@shared/api'
+import {
+  TransformedInboxMessages,
+  mergeInboxMessages,
+  transformInboxMessages,
+} from './inboxTransform'
+
+// Hand-written because codegen needs a server token; the endpoint name is camelCase
+// so a future generated `GetProjectInbox` operation cannot collide with it.
```
Innders: Still don't understand why this is handwritten, can it be done in .graphql file?

src/services/inbox/getProjectInbox.ts
```diff
@@ -0,0 +1,119 @@
...
+  important: boolean | null
+}
+
+type ProjectInboxResponse = {
+  project: { activities: GetInboxMessagesQuery['inbox'] }
+}
+
+export const projectInboxApi = gqlApi.injectEndpoints({
+  endpoints: (build) => ({
+    getProjectInbox: build.query<TransformedInboxMessages, GetProjectInboxArgs>({
```
Innders: You will want to use infiniteQuery here like with GetActivitiesInfinite query. Look at getActivities.ts for how to do this.

src/services/inbox/getProjectInbox.ts
```diff
@@ -0,0 +1,119 @@
...
+      query: ({ projectName, userName, referenceTypes, activityTypes, filter, last, cursor }) => ({
```
Innders: You will need to add onCacheEntry with pubsub so that new activities are added to the cache without the need to refresh. Again look at `getActivities.ts`

src/pages/InboxPage/components/InboxSearchFilter.tsx
```diff
@@ -0,0 +1,216 @@
...
+  { id: 'watching', label: 'Watching', icon: 'visibility' },
+  { id: 'relation', label: 'Related', icon: 'link' },
+]
+
+const ORIGIN_TYPE_VALUES = [
+  { id: 'task', label: 'Task', icon: 'check_circle' },
+  { id: 'version', label: 'Version', icon: 'layers' },
+  { id: 'folder', label: 'Folder', icon: 'folder' },
+  { id: 'product', label: 'Product', icon: 'inventory_2' },
```
Innders: Nice that you can filter by entity type! Remove product as you cannot comment on products anywhere yet

Innders (changes requested):
Great start, very usable and useful already. Made some comments on the code but also:

- [x] Add filter for the new activity type "review" `version.review`

Innders (changes requested):
- [ ] Add three options for review filter: All reviews, Approved, Requested changes

Image

- [ ] Fix review filter error

Image

Innders (changes requested):
Awesome

- [x] Projects list full height and the filters bar to the left of projects table.

Image

Innders (changes requested):
Great stuff!

- [x] Keep the icon as always the "unread items" one (when it's selected)

Image

- [x] Add views support to the page so that you can save your different filter set ups

src/pages/InboxPage/hooks/useInboxProject.ts
```diff
@@ -0,0 +1,56 @@
+import { useCallback, useEffect, useRef } from 'react'
+import { useQueryParam } from 'use-query-params'
+
+// not `project`: the details panel owns that one and rewrites it on every message opened
+export const INBOX_PROJECT_PARAM = 'inboxProject'
```
Innders: I would just use `project` here as we already in the inbox scope.

src/pages/InboxPage/hooks/useInboxViewSettings.ts
```diff
@@ -0,0 +1,117 @@
...
+      .catch((error) => {
+        // drop the mark so it can retry, otherwise nothing ever persists
+        bootstrapped.current.delete(viewType)
+        console.error('Failed to create inbox working view:', error)
+        toast.error('Inbox filters cannot be saved right now')
+      })
+  }, [viewType, viewSettings, isLoadingViews, createView, setSelectedView])
+
+  // wrapped: updateViewSettings clears local state with null, which a null project also means
+  const [localProjectName, setLocalProjectName] = useState<{ value: string | null } | null>(null)
```
Innders: The project shouldn't be saved to the view.

src/services/inbox/getInbox.ts
```diff
@@ -26,36 +30,16 @@ export const enhancedInboxGraphql = gqlApi.enhanceEndpoints<TagTypes, UpdatedDef
...
-          ...projectNames.filter((p: string) => !lastProjectNames.includes(p)),
-        ]
-
-        return {
-          messages: newMessages,
-          projectNames: newProjectNames,
-          pageInfo,
-        }
-      },
+      merge: mergeInboxMessages,
```
Innders: Convert this query to an infinite query so that pagination is done properly

src/services/inbox/getProjectInbox.ts
```diff
@@ -0,0 +1,188 @@
...
+  args: ProjectInboxInfiniteArgs,
+  page: { last: number; cursor?: string | null },
+) => {
+  const request = dispatch(
+    gqlApi.endpoints.GetProjectInbox.initiate(pageVariables(args, page), { forceRefetch: true }),
+  )
+  try {
+    return await request
+  } finally {
+    request.unsubscribe?.()
```
Innders: You can specifically not subscribe to the query like `{ forceRefetch: true, subscribe: false  }`

src/services/inbox/getProjectInbox.ts
```diff
@@ -0,0 +1,188 @@
...
+        // every arrival in a window asks for the same thing, so they collapse to one entry
+        const batcher = createRealtimeBatcher<unknown>(
+          async (_messages, isActive) => {
+            let cursor: string | null = ''
+            while (cursor !== null && isActive()) {
+              cursor = await topUp(cursor)
+            }
+          },
+          () => 'inbox.message',
+        )
```
Innders: I'm not convinced by this code. Can the query for the new message explicitly include all of the new activity ids. GetProjectInbox could accept optional ids argument that could be used to fetch new messages.

So when the batcher is finally called with all the messages you make 1 GetProjectInbox request with all the new ids, if you include the filters as well then you know which ones are filtered out or not. The same concept is used for tasks on the overview page.

That way we are only getting the data required.

Innders (changes requested):
Seems to all be working nicely, some cosmetic fixes and some code clean up.

- [x] Add gap between projects table and header

Image

- [x] Tab unread count should come before the views button

Image

src/services/inbox/getProjectInbox.ts
filipvnencak: The problem is that the ids aren't in the payload,  inbox.message only carries project and summary: {isImportant}. activity.created does have the id

Innders: > inbox.message only carries project and summary

@kalisp can `inbox.message` message have the `activityId` in the message summary?

@filipvnencak lets try to get backend to add it so that this can be done properly. But nice workaround though ha!

src/pages/InboxPage/hooks/useInboxProject.ts
```diff
@@ -0,0 +1,31 @@
+import { useCallback, useEffect, useState } from 'react'
+import { useQueryParam } from 'use-query-params'
+
+interface Options {
+  enabled?: boolean
+}
+
+// the details panel rewrites the same param for the entity it opens, so the selection is
```
Innders: So there is a clash with using `project` query param. Makes sense why the param was `inboxProject`. In this case then `inboxProject` should be used to prevent any weird behavior.

src/pages/InboxPage/hooks/useInboxViewSettings.ts
```diff
@@ -0,0 +1,92 @@
...
+}
+
+const useInboxViewSettings = (): Return => {
+  const { viewType, viewSettings, isLoadingViews, setSelectedView } = useViewsContext()
+  const { updateViewSettings } = useViewUpdateHelper()
+  const [createView] = useCreateViewMutation()
+
+  const settings = viewSettings as InboxViewSettings | undefined
+
+  // a tab's first visit has no view to write to, and updates abort without a baseline
```
Innders: I think I might have fixed this issue of the first view being rejected since writing this. Try merging develop and testing again when there is no view yet.

Innders (changes requested):
- [x] Code changes
- [x] Add activityId to backend so that realtime updates can be done properly

Innders (changes requested):
I think this hook could use `useQueryParams` instead?

https://github.com/ynput/ayon-frontend/blob/9c023ddecf27f87103a2b9bc0959d994203fb66c/src/pages/InboxPage/hooks/useInboxProject.ts

Innders (approved):

PR #2239, opened 2026-08-10
YN-0883: Enable word wrapping for column content

### LINKED ISSUE #2174: YN-0883: Enable word wrapping for column content

When the row height is bigger, content should wrap in the cell to make use of the space. Apply to all column types. 

&nbsp;

### Additional context:

_This issue was automatically created from AYON ticket [YN-0883]( https://support.ayon.app/projects/Active_Clients/overview?project=Active_Clients&type=folder&id=81ca64d7021b427a82ccac445c7dbc78)

filipvnencak:
## Description of changes 

Cell content now wraps onto multiple lines when the row height is raised, instead of staying on one truncated line. Applies to text, enum, chips, date and name columns on Overview, Lists and Versions/Products.

## Technical details

- Wrap mode is CSS-only: `Cell` and `TableCellContent` are `container-type: size` containers, widgets opt in via the new `wrapMode` mixin in `widgets/wrapMode.ts`.
- `constants.ts` holds both thresholds — `WRAP_MIN_CELL_HEIGHT` (container content box) and `WRAP_MIN_ROW_HEIGHT` (row height) — so every column flips at the same slider position.
- `Chips` now measures how many rows fit and packs chips with greedy line breaking, so `+N` reflects the multi-row space.
- `EnumCellValue` wraps values on a fixed 24px lattice, so overflowing rows clip whole instead of mid-chip.
- `TextWidget` overflow detection checks both axes, keeping the hover preview working for wrapped text.

## Additional context

closes #2174 
Links:
Screenshot 2026-08-10 at 08 55 55

Daterime:
Screenshot 2026-08-10 at 08 56 50

Folder Name
Screenshot 2026-08-10 at 08 57 55

Enum:
Screenshot 2026-08-10 at 08 59 18

Subtasks
Screenshot 2026-08-10 at 09 09 25


Innders (changes requested):
- [x] Do not wrap single select enum eg. `status` `folderType` etc. They should be vertically centered as well.
- [x] Alignment of the content in the cells should stay centered

Innders (changes requested):
Thanks for fixing those issues. Just this one more thing I noticed.

- [x] The avatar image gets pushed off the screen. It's useful to be able to resize it to just see the image.

Image

The way it currently works is the best
Image

Innders (approved):

PR #2241, opened 2026-08-11
Table: Make it easier to add columns with button and search

### LINKED ISSUE #2208: Table: Make it easier to add columns with button and search
## Story

I want to quickly add a new attribute column to the table

### Problems

Current I must
- Navigate to "customise" menu
- Click on columns
- Find the correct column through a long list

# Proposal

## Add button
Add a "Add" button after the label column that automatically opens the columns settings panel.

There is already a "Add list attribute" button on the list table and this should be hijacked to open the columns settings instead, on all the pages with tables.

Image

## Add columns menu

_Searching through the hidden columns is difficult._

- Remove the "Hidden columns" section
- Add an "Add" button the settings header.
- Use the menu component to list the columns to add, some with submenus like different entities and links.
- Add an "Add column" button at the bottom of "Visible Columns" that opens the top Add button dropdown menu

Image

Image

_The dropdown menu should work in a similar way to the "group by" button on the Review page_

## Search

_Often you know exactly which column you want to add._

Add a search button that opens a search input at the top of the columns settings, like in the Projects/Lists table.

When searching the columns become a flat list with no sorting. This becomes purely about adding/remove columns.

## TODO

- [ ] Hijack existing "Add list attribute" button to point to Columns settings
- [ ] Add "Add" button to all pages with the table
- [ ] Columns settings: Replace hidden columns with add dropdown menu
- [ ] Columns settings: Add search to

filipvnencak:
## Description of changes 

Adds a plus button at the end of the table header. It opens a menu with all columns and toggles them on click. The columns settings panel gets a search field too.

## Technical details

- `AddColumnButton` floats over the table. `useLastColumnOffset` keeps it on the last column edge and falls back to the right corner when it does not fit.
- `buildAddColumnsMenu` lists every column. Visible ones get a checkmark, clicking toggles them through `useAddColumnsMenu`.
- Attributes and links go into sub menus. Open sub menus now refresh when the items change, so checkmarks update while the menu stays open.
- `useProjectTableColumnItems` is the one source of columns for both the button and the settings panel.
- Search input moved out of Lists into a shared `TableSearch`. Lists, projects list and the entity picker use it now.
- Search results in the panel keep the default column order, with grouped columns at the bottom.

## Additional context

Closes #2208

Screenshot 2026-08-13 at 16 09 22
Screenshot 2026-08-13 at 16 09 14

https://github.com/user-attachments/assets/0d2f4430-9c94-49fa-b086-850df52391c9


Innders (changes requested):
- [x] Search results that in sub folders need show that path.
- [x] This path should also be part of the search results.

`Version attributes / FSP`

Image

Innders (changes requested):
## Bugs

- [x] Align the menu to the left of the button.

Image

- [x] Remove the min-width of the search input

Image

- [x] Search causes a unintended padding of the column items.

Image

What is the default sorting of the search items, seems kind of random.
- [x] The sorting should use the default order of columns BUT with grouped items at the bottom.

- [x] Remove rogue outline on the side

Image

## Menu Selection

After testing it, it feels weird that I can only see the columns that haven't been added yet.

- [x] Always show all the columns in a consistent order, not matter their visibility.
- [x] When a column is visible, highlight it and add a checkmark at the end of it.
- [x] You should be able to toggle the visibility by clicking on the column. (You can hie columns that are checked)

https://discord.com/channels/517362899170230292/1001811244535259207/1536740618896674868

Innders (changes requested):
Awesome, this is SO much better to use now! You are going to make a lot of people very happy.

Some small things to polish it more.

- [x] Rename "List attributes" with "Create list attribute" with a plus icon.

Image

- [x] On the overview page, group all the attributes into an "Attributes" sub menu

- [x] On the lists page, versions have "Version attributes" sub menu. Folders and Tasks lists need to have the same.

Innders (changes requested):
Could you please try adding two features that might help.

## Bulk select / deselect

When showing or hiding multiple columns, you can click and drag across the list instead of selecting each column individually.

- Click and hold on the first column to set the action.
- Drag over additional columns to apply the same action.
- If the first column is hidden, dragging will show every column you pass over.
- If the first column is shown, dragging will hide every column you pass over.
- Columns that are already in the correct state are left unchanged.

The state of the first column you click determines the action for the entire drag.

For example, if you click a hidden column, it becomes visible. As you drag over other columns, they are also made visible.

## Icons

Add the attribute icons to the column items.

Innders (changes requested):
The icons look great!

Did you try out the click drag mechanic?

- [x] Bulk select with click drag
- [x] The top gets cut off making it impossible to reach some items.

Image

Innders (changes requested):
The dragging on top of the table is a really awesome idea. It's going to look so good on demos and make it so much faster to use.

The biggest problem now is that it's not obvious you can do that.

There are two things that would help

- [x] When hovering the item, swap the item icon for the drag handle item.
- [x] When dragging the item the table should have a subtle highlight effect over it to show that it's a droppable zone.

Innders (changes requested):
Nice one, that's a lot more obvious.

- [x] Search items need to the same drag icon reveal

Image

- [x] There is this weird behaviour that the dragging only starts once my mouse goes outside, makes it feel like the dragging isn't working. Clicking the drag handle should start the drag immediately, otherwise it feels like the dragging isn't working. (Don't spend long on this)

- [x] Remove dashed outline from dropzone

- [x] The overlay seems doubled up?

https://github.com/user-attachments/assets/a322d38c-7724-49e7-b885-540e2675c226

shared/src/components/ProjectTableSettings/AddColumnButton.tsx
```diff
@@ -0,0 +1,72 @@
...
+    includeLinks,
+  })
+  const { menuItems, hasMenuItems, dragOverlay } = useAddColumnsMenu({
+    columns: visibleColumns,
+    scopes,
+    extraItems: extraMenuItems,
+  })
+  // state, not a ref: rendering null detaches the node, which has to be measured again
+  const [container, setContainer] = useState<HTMLDivElement | null>(null)
+  const offset = useLastColumnOffset(container)
```
Innders: I don't understand why there needs to be a position calculation here?

shared/src/components/ProjectTableSettings/AddColumnButton.tsx
```diff
@@ -0,0 +1,72 @@
...
+import { useProjectTableColumnItems } from './useProjectTableColumnItems'
+import { AddColumnMenu } from './AddColumnMenu'
+import type { MenuItemType } from '../Menu'
+
+const ADD_COLUMN_MENU_TABLE_ID = 'add-column-menu-table'
+
+const ButtonPosition = styled.div`
+  position: absolute;
+  top: 4px;
+  right: 4px;
```
Innders: Move the button further in so that it's not over the scrollbar.
```suggestion
  right: 24px;
```

shared/src/components/ProjectTableSettings/AddColumnButton.tsx
```diff
@@ -0,0 +1,72 @@
...
+    <ButtonPosition
+      ref={setContainer}
+      style={offset === null ? undefined : { left: offset, right: 'auto' }}
+    >
+      <Button
+        icon="add"
+        id={ADD_COLUMN_MENU_TABLE_ID}
+        data-tooltip="Add column"
+        disabled={!hasMenuItems}
+        onClick={() => toggleMenuOpen(ADD_COLUMN_MENU_TABLE_ID)}
```
Innders, line 60: Use secondary colour so that the button stands out slightly more

```suggestion
        variant="tonal"
```

shared/src/containers/ProjectTreeTable/buildTreeTableColumns.tsx
```diff
@@ -63,6 +63,32 @@ export const getColumnLabel = (columnId: string, scopes: string[] = []) => {
...
+  task_entity: getEntityTypeIcon('task'),
+  version: getEntityTypeIcon('version'),
+  version_entity: getEntityTypeIcon('version'),
+  product: getEntityTypeIcon('product'),
+  tags: 'local_offer',
+  createdAt: 'calendar_month',
+  updatedAt: 'calendar_month',
+  subtasks: 'checklist',
+  comments: 'comment',
+}
```
Innders, line 83: These icons are already defined in `getAttributeIcon`. Ensure there is one source of truth for the icons.  https://github.com/ynput/ayon-frontend/blob/11158e5009f8cc66a5ecacfc35394738ff8af904/shared/src/util/getAttributeIcon.ts#L6.

Innders (changes requested):
- [x] Merge conflicts
- [x] Code feedback
- [x] What is the order of the attributes, I needs to match the order set in the attribute settings.

Image

Image

shared/src/components/ProjectTableSettings/AddColumnButton.tsx
filipvnencak: It was so the + button sits at the edge of the last column instead of the far right when the columns don't fill the width. 
Removed

filipvnencak:
> What is the order of the attributes, I needs to match the order set in the attribute settings.

@Innders It was alphabetical, now uses server order so no changes in FE now

shared/src/components/ProjectTableSettings/AddColumnButton.tsx
Innders: Ah I see get it now, yeah lets keep it basic :)

Innders (approved):
This a great PR, I really like the dragging into the table feature!

PR #2251, opened 2026-08-13
YN-0997 - Overview: Fix folder count stats when searching for tasks

### LINKED ISSUE #2250: YN-0997: Overview: folder stats broken when using task search

When using a fuzzy search on the overview page, the folders stats become incorrect because they do not account for the task search filtering.

&nbsp;

Tasks are using 'filter' and 'search' but folders are only using 'taskFilter'.

&nbsp;

Add 'taskSearch' so that the folders are filtered by the task search.

&nbsp;

Then this can be added to stats to show the correct remaining number of folders. 

_No response_

### Additional context:

_This issue was automatically created from AYON ticket [YN-0997]( https://support.ayon.app/projects/Active_Clients/overview?project=Active_Clients&type=folder&id=bd50ac8096f911f1bfa13172523baf6f)

Innders:
## Description of changes 
Searching in the overview page is actually searching for tasks. This search on tasks was not passed to the folders resolver and so the count of folders was wrong.

Now the folders resolver has an `taskSearch` argument that can be passed to it to properly count the left over folders from the task search.

image

### DIFF
--- a/gen/graphql.schema.json
+++ b/gen/graphql.schema.json
@@ -9003,7 +9003,19 @@
               },
               {
                 "name": "taskFilter",
-                "description": "Fitler folders by tasks",
+                "description": "Filter folders by tasks",
+                "type": {
+                  "kind": "SCALAR",
+                  "name": "String",
+                  "ofType": null
+                },
+                "defaultValue": "null",
+                "isDeprecated": false,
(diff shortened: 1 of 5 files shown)


filipvnencak (approved):
Works great!!!

PR #2257, opened 2026-08-14
Slicer: Fix selection of redundant children being sent to server

Innders:
## Problem
There was an edge case where using shift select on parent folders would cause all child folders to be selected as well.

These child folders would then be included in tasks and versions queries along with `includeFolderChildren` even though their parent has been selected as well.

This would be very inefficient as every folder would scan it all it's children.

## Description of changes
1. When requesting data using folder ids and includeFolderChildren, filter out child folders so that only top level parent folders are left.
2. When using shift select, ensure child rows are not selected if their parent is collapsed.

### DIFF
--- a/shared/src/containers/ProjectTreeTable/hooks/useFetchOverviewData.ts
+++ b/shared/src/containers/ProjectTreeTable/hooks/useFetchOverviewData.ts
@@ -119,7 +119,7 @@ export const useFetchOverviewData = ({
     isLoading: isLoadingFolders,
     isUninitialized: isUninitializedFolders,
     getFolderById,
-    getParentFolderIds,
+    getFolderIdsWithoutChildren,
   } = useProjectFoldersContext()
 
   const expandedParentIds = Object.entries(expanded)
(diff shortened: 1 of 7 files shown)


shared/src/containers/ProjectTreeTable/hooks/useFetchOverviewData.ts
```diff
@@ -400,25 +401,40 @@ export const useFetchOverviewData = ({
...
   // In hierarchy mode with slicer-selected folders, use GetTasksList with folderIds to fetch
   // tasks directly under those folders in a single paginated request, rather than firing one
   // request per folder via getOverviewTasksByFolders.
-  const hierarchySlicerFolderIds =
-    showHierarchy && excludeSelectedFolders && selectedFolders.length ? selectedFolders : undefined
+  const getTasksDirectlyUnderFolder = showHierarchy && excludeSelectedFolders
+
+  const getTasksListFolderIds = () => {
+    if (taskIds?.length) {
+      // we are getting specific task ids so we do not need to tasks based on folder ids
```
BigRoy, line 416: need to tasks? 🤔  Grammar is off?

BigRoy (commented):
Works for me - much better!

Overview still seems to pass multiples of folder ids for GetTaskColumnStats and GetFolderColumnStats:

Image

I wonder if it's redundant because it may be including recursively the data as well? 🤔 Or maybe it doesn't 🤷

Innders:
> I wonder if it's redundant because it may be including recursively the data as well?

Yeah it does do it recursively so changing the stats queries matters a lot.

Updated the stats queries to use the parsed slicer selection so that only folder parents are included.

BigRoy (approved):
Works for me.

PR #2260, opened 2026-08-17
Products Folder Columns with Entity Table Columns Rewrite

### LINKED ISSUE #2158: YN-0911: Products: Add version parent folder columns

The ability to see folder fields in the products table, status, priority, custom attributes.

### Additional context:

_This issue was automatically created from AYON ticket [YN-0911]( https://support.ayon.app/projects/Active_Clients/overview?project=Active_Clients&type=folder&id=72e55ad07c4411f1ada54d6c1b976b20)

Innders:
## Shared `TableRow` Model

`TableRow` now separates the row’s main entity from related entities:

```ts
type TableRow = {
  id: string
  primary: EntityData
  parents?: Partial<Record<EntityType, EntityData>>
}
```

For a Versions/Products row, the structure is typically:

```ts
{
  primary: version,
  parents: {
    product,
    task,
    folder,
  }
}
```

This gives the table a consistent way to access fields regardless of where they come from:

```ts
getScopedValue(row, 'primary', 'version')
getScopedValue(row, 'product', 'productBaseType')
getScopedValue(row, 'folder', 'status')
getScopedValue(row, 'task', 'subType')
```

Columns no longer depend on page-specific flat fields such as `folderStatus`, `taskType`, or `productBaseType` being copied directly onto the row.

## Scoped Parent Columns

`ParentColumnDefinition` describes fields belonging to a related entity:

```ts
{
  id: 'folder_status',
  scope: 'folder',
  field: 'status',
  label: 'Folder status',
  optionKey: 'folderStatus',
}
```

The shared table builder uses these definitions to create:

- Accessors
- Headers
- Cell renderers
- Sorting behavior
- Editing behavior
- Column menu entries
- Settings-panel entries

This allows pages to declare their available parent fields rather than implementing custom column logic for every entity relationship.

## Versions/Products Page Changes

The Versions/Products page now uses explicit parent-column configuration through `VP_PARENT_COLUMNS`.

It adds folder-related columns to the version/product table, including:

- Folder status
- Folder type
- Folder tags
- Folder updated at
- Folder created at

It also exposes task-related fields such as:

- Task type
- Task status

And product-related fields such as:

- Product base type

The Products page also supports parent attributes through the shared attribute-column generation. Folder, task, and product attributes can be offered in the column settings/menu under their respective scopes, provided the relevant attribute definitions are available from the table context.

For example, the column menu can now distinguish between:

- Primary/version attributes
- Folder attributes
- Task attributes
- Product attributes

## Parent Attribute Control

Not every page has attribute data for every parent entity. `ParentColumnDefinition` therefore supports:

```ts
includeAttributes?: boolean
```

This lets each page decide whether attributes should be generated for a parent scope.

- Versions/Products keeps parent attributes enabled for its configured parent scopes.
- Lists disables them because list-item responses do not contain complete parent attribute metadata.

The flag is applied consistently to both:

- The actual table columns
- The column management menus

This prevents the UI from offering columns that the page cannot populate.

## Dynamic Lists Behavior

Lists use the same shared architecture, but their primary type is selected at runtime:

- Folder list: primary folder
- Task list: primary task, folder parent
- Version list: primary version, product/task/folder parents

The selected primary entity is not also exposed as a parent. This avoids duplicate menus such as “Task fields” on a task list when those fields already belong to the primary task entity.

## Data and Mutation Separation

The raw API list-item objects remain separate from `TableRow`.

Raw list-item data is still used for:

- RTK Query caching
- List-item mutations
- Entity mutations
- Filtering and pagination
- Optimistic updates

`TableRow` is the normalized view model used by the table and column system.

When a cell is edited, the table resolves the scoped entity and the update logic maps the change back to the correct API operation, whether it targets:

- The primary entity
- A parent entity
- A list-item-specific field

This keeps the shared table model clean without requiring the API mutation layer to understand the normalized row structure.

## Testing

- [ ] Overview page still works
- [ ] Products page with folder/tasks/product columns enabled.
- [ ] Lists page: `Folder`
- [ ] Lists page: `Task`
- [ ] Lists page: `Version`


filipvnencak (changes requested):
Two bugs here: one is Entity type folder while grouping by Status. Second one is parent folder is empty, I guess it's because the prop while grouping is pointing to parent folder in client and here is group - for example not ready
Image

Products page - Product group by, Thumbnail error

`{"code":404,"detail":"API endpoint /api/projects/.../products/.../thumbnail not found"}`
/products/{id}/thumbnail does not exist on server
Image


And there is the same error 2 times in one render in network, for example: Request validation error in [GET] /api/projects/demo_Big_Episodic/folders/loading-4/thumbnail



VPFolder fragment now pulls 6 extra fields on every version, with no gate on visible columns.

Measured on my 100 versions:
- develop fragments: 122780 B
- this PR: 144255 B so 21475 B more

That's +215 B per version row, fetched even when no folder column is shown.

PR #2261, opened 2026-08-17
Summaries: show column summaries for multi-select attributes

filipvnencak:
## Description of changes 

Fixes multi-select attributes always showing `—` in the column summaries footer. They were never sent to the stats query at all.

## Technical details

- `buildMetricTargets` now emits ENUM targets for `list_of_strings` and `list_of_integers` attribs (`metricTargets.ts`).
- Backend buckets the whole array as one serialized value; `normalizeDistribution` in `columnStats.ts` already unnests it.
- `datetime` and boolean attribs stay skipped — both still error on the backend, and one bad target fails the whole query.
- `list_of_integers` untested, no such attribute in any test project.


Innders (approved):
Works for me

PR #2264, opened 2026-08-19
Lists: Creating new review session from context menu does not show new list

### LINKED ISSUE #2147: Lists: Creating new review session from context menu does not show new list
You must refresh the page to see the new review session

filipvnencak:
## Description of changes 

Fixes a review session created from the folder/task context menu not appearing in the Reviews list until the page was reloaded.

## Technical details

- `EntityListsContext.createNewList` now invalidates `{ entityList, LIST }` after the `review-save-session-from-*` action. The previous tags used the newly created list id, which no cache entry provides.
- Per-list tags are still sent, but only when the action returns an id.
- Only affects the addon action path (folder/task). Version lists already invalidated correctly via `createEntityList`.

## Additional context

Closes #2147


Innders (approved):
Didn't test but code LGTM

PR #2268, opened 2026-08-21
Summaries: Name breakdown of folder and task types

### LINKED ISSUE #2247: Summaries: Name breakdown of folder and task types
## Story

I want to be able to see how many shots I have in a sequence, filtering out the sequence folder.

### Problems

Because the "parent" folder shows in the table #2246 the folders count of the children is off by one.

### Proposal

1. Add a new "Breakdown" option for the name column.
2. The footer value stays the same as before, counting all folders/tasks.
3. Hovering now shows a breakdown of the folder/task types.
Image

filipvnencak:
## Description of changes 

Adds a Breakdown option to the name column summary. The footer keeps showing the plain count, hovering now shows the type split.

## Technical details

- `breakdown` added to `SummaryCalc`, this is what gets persisted per view
- stats query asks for the subType distribution when breakdown is on, even when the subType column is hidden
- new `mainBreakdownFields` prop tells the addon what each side breaks down by: Overview folder + task types, Versions/Products product types
- versions have no type, so there only the products side shows up
- Lists doesn't set the prop, so the option stays hidden there

## Additional context

Closes #2247
Powerpack PR: ynput/ayon-power-pack#81

### DIFF
--- a/shared/src/api/queries/columnStats/metricTargets.ts
+++ b/shared/src/api/queries/columnStats/metricTargets.ts
@@ -153,8 +153,10 @@ export const buildMetricTargets = ({
     targets.push({ field: 'assignees', aggregations: ENUM })
   }
 
-  // name breakdown needs the type distribution even when the subType column is hidden
-  const nameBreakdown = columnSummaries?.['name'] === 'breakdown'
+  // no visibility check: it's opt-in, and a name column missing from both maps would kill this
+  const nameBreakdown =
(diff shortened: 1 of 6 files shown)


Innders (approved):
Nice one

Innders:
Note that a bug was found after merging -> https://github.com/ynput/ayon-frontend/issues/2270

PR #2269, opened 2026-08-21
Products: product type should not enum dropdown and string type

### LINKED ISSUE #2232: Products: product type should not enum dropdown and string type
Product types are immutable and they should not be a enum dropdown. They should be presented as a string type so they are read only and custom product types are **not** shown as invalid.

filipvnencak:
## Description of changes 

The type column no longer opens a dropdown on products and versions.
Custom product types now show as plain text instead of a red invalid chip.

## Technical details

- `subType` cell is read only for product/version rows, known types keep their icon chip (`buildTreeTableColumns.tsx`)
- field id for those rows was `taskType`, now `productType`
- paste into `subType` is skipped on product/version rows, it was validated against task types and cancelled the whole paste (`ClipboardContext.tsx`)

## Additional context

Closes #2232
Screenshot 2026-08-21 at 15 50 50


Innders (approved):
Didn't test but code looks good. Thanks!

PR #2273, opened 2026-08-24
2065 lists filter bar

### LINKED ISSUE #2065: Lists: Search filter bar for lists should be moved to top of lists table
Currently you must use the dropdown menu to open a dialog to edit the lists filtering.

This is confusing and unintuitive. It's even worse when you have filters set and you can't see them.

Move the SearchFilter bar to underneath the table header, just where the normal search input would go.

By default it is hidden but you can use the dropdown menu item to show it. Clicking the menu item again will remove any current search filters and hide the bar.

Use the compact version of the search filter.

Image

filipvnencak:
## Description of changes 

Lists filtering moves out of the fullscreen dialog into a compact search and filter bar under the lists table header. Active filters stay visible.

## Technical details

- New `ListsFiltersBar` (one compact `SearchFilter`) replaces `ListsFiltersDialog` and the dead `ListsFiltersButton`; options moved verbatim into `useListsFilterOptions`.
- `ListsDataContext` strips `--search--` chips before the server query, so search filters client-side and never refires the lists query.
- Main lists only, via `filtersBar={!picker && !isReview}`; picker dialogs and review sessions keep the plain search input. 

## Additional context

Screenshot 2026-08-25 at 11 47 34

Closes #2065


Innders (changes requested):
Keep the regular search and filters separate. If they are both open just stack them on top each other. The filters won't be used that much so the search needs to be fast and simple.

Innders (changes requested):
Better :)

- [x] Only have one filter icon

Image

- [x] Add add a close icon on the right. Same as the text search.
- [x] Reduce the gap between the search and filter as it's got double padding right now.

Innders (changes requested):
Sorry one last tiny thing!

- [x] Align close button center
- [x] Make close button smaller (less padding)

Image

Innders (approved):
Thanks

PR #2275, opened 2026-08-24
Summaries: Do not show main breakdown type when active

### LINKED ISSUE #2270: Summaries: Do not show main breakdown type when active
The main name breakdown should not show folder/task types if folder/task is not active.

Bug found as part of this PR: https://github.com/ynput/ayon-frontend/pull/2268

Image

filipvnencak:
## Description of changes
Fixes #2270 frontend half, rendering fix is in ynput/ayon-power-pack#83

Name breakdown no longer requests subType stats for a scope that is toggled off.

## Technical details

`buildMetricTargets` gates the `nameBreakdown` subType target on the name column's scope: folder/product entities need `primary`, task/version need `secondary`. Unset scope stays `all`, so nothing changes for existing users.

Screenshot 2026-08-24 at 16 22 47

### DIFF
--- a/shared/src/api/queries/columnStats/metricTargets.ts
+++ b/shared/src/api/queries/columnStats/metricTargets.ts
@@ -154,9 +154,15 @@ export const buildMetricTargets = ({
   }
 
   // no visibility check: it's opt-in, and a name column missing from both maps would kill this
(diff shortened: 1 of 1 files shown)


Innders (approved):
Works thanks

PR #2279, opened 2026-08-26
YN-0841: Entity Picker Thumbnails

### LINKED ISSUE #2060: YN-0841: Entity Picker Thumbnails

Add a thumbnail to the start of the row in the Entity Picker. Keep the icon paired with the label.

Hovering over the text or thumbnail of a row should show the EntityTooltip so that the user can see more information about the entity and a bigger thumbnail. 

Image

### Additional context:
link to discussion on [Discord](https://discord.com/channels/517362899170230292/1514607205880692826/1515025102201295051)
_(might be a private channel)_

_This issue was automatically created from AYON ticket [YN-0841]( https://support.ayon.app/projects/Active_Clients/overview?project=Active_Clients&type=folder&id=241510b0688711f1853f11d5b1f0cea1)

filipvnencak:
## Description of changes 

Adds thumbnails to entity picker rows, plus a hover tooltip that now also covers folders and workfiles.

## Technical details

- `getEntityThumbnailUrl` takes a `placeholder` param; picker rows pass `none` so a missing thumbnail 404s instead of returning a blank image.
- New `useImageStatus` hook preloads off-DOM and caches the result per URL, so no broken-image flash and no repeated 404s.
- `SimpleTableCellTemplate` keeps an empty (transparent) image slot when a thumbnail is missing, so rows stay aligned. Cell order is thumbnail, expander, icon, label.
- Product rows use `featuredVersion` instead of `latestVersion` for their thumbnail.
- `EntityTooltip` moved to `shared/components`, extended with folder and workfile queries, and opened from picker rows via `useRowTooltip`.

## Additional context

Closes #2060 
Screenshot 2026-09-09 at 09 35 38


Innders (commented):
Cool, just some notes.

- [x] Add small 2px border radius to the thumbnail.
- [x] Try putting the image after the indent and after the dropdown button.

Image

- [x] The tooltip hover should only be over the content of the row and it should be instant.

Image

filipvnencak:
@Innders  For me it looks very weird for if thumbnails is between indent and icon, because of empty/free-space if thumbnails isn't there

Screenshot 2026-08-27 at 14 19 51

Innders (changes requested):
Yep the thumbnail looks super weird there. I think the best place is at the end now.

- [x] Thumbnail at the end of the row

Image

shared/src/containers/EntityPickerDialog/hooks/useRowTooltip.tsx
```diff
@@ -0,0 +1,46 @@
...
+    return () => window.removeEventListener('scroll', close, true)
+  }, [hovered, close])
+
+  const onMouseOver = useCallback(
+    (event: React.MouseEvent<HTMLTableRowElement>) => {
+      if (!isEnabled) return
+      const id = event.currentTarget.id
+      const anchor = (event.target as HTMLElement).closest(
+        '.image, .value, .path',
+      ) as HTMLElement | null
```
Innders: Only show the tooltip when hovering the image.
```suggestion
      const anchor = (event.target as HTMLElement).closest('.image') as HTMLElement | null
```

shared/src/containers/SimpleTable/SimpleTableRowTemplate.tsx
```diff
@@ -112,13 +115,39 @@ export const SimpleTableCellTemplate = forwardRef<HTMLDivElement, SimpleTableCel
     ref,
   ) => {
     const [renameValue, setRenameValue] = useState(renameInitialValue ?? value)
+    const imgStatus = useImageStatus(img)
```
Innders: If there is no thumbnail for the entity then it should return empty image and not 404? This means you can just render empty image inside image container. Then this hook is not needed.

Innders (changes requested):
UI looks good just a couple things.

- [ ] Only show tooltip on thumbnail hover (suggested comment)
- [ ] Look into if useImageStatus is needed

Innders (approved):
Looks great

PR #2280, opened 2026-08-26
Wait for server full boot before fetching data

martastain:
Handles backend changes in https://github.com/ynput/ayon-backend/pull/1094

This pull request enhances the server restart handling logic in the `SocketProvider` component to ensure smoother state management when the backend server restarts. The main improvements involve debouncing the API state reset and making the logic more robust and less error-prone.

**Server restart handling improvements:**

* Introduced a debounced function `debouncedResetCall` to delay resetting the API state for 1 second after detecting a server restart, preventing unnecessary rapid resets and ensuring all replicas restart together.
* Updated the WebSocket message handler to trigger the debounced reset only when a `server.started` message with status `finished` is received and the server is marked as restarting, improving accuracy and reducing redundant resets.
* Removed the previous immediate API state reset logic from the `useEffect`, centralizing the reset logic within the WebSocket message handler for better control and predictability.

### DIFF
--- a/shared/src/context/WebsocketContext.tsx
+++ b/shared/src/context/WebsocketContext.tsx
@@ -117,6 +117,17 @@ export const SocketProvider = ({
   // Using useRef to persist the closure state across renders
   const messageStatsRef = useRef({ callCount: 0, lastCall: Date.now() })
 
+
+  const debouncedResetCall = debounce(() => {
+    // all replicas should start about the same time, 
+    // so we can reset the API state after a short delay
+    console.log('Resetting API state due to server restart')
+    setServerRestartingVisible(false)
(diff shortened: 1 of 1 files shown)


Innders (changes requested):
After a successful restart the frontend will be stuck in a refreshing loop that happens every 5s with console log messages.

- Server replica booted
- WebsocketContext.tsx:124 Resetting API state due to server restart

The loop is only broke by refreshing the page.

https://github.com/user-attachments/assets/10760f54-ccfb-4f49-aa15-7b3cd70a3585

martastain:
@Innders Thanks. it seems it only affected dev server and i tested it against a build. It seems to be fixed now

Innders (approved):

PR #2283, opened 2026-08-27
Attributes: Dynamic Enums

### LINKED ISSUE #2243: Attributes: Dynamic Enums
## Story

As an admin or manager, I want custom attributes to populate their option lists dynamically from system data (users, teams, planner resources) so that I don’t have to manually maintain large, constantly changing option lists.

- I want to be able to populate my custom attribute enum with dynamic data instead of manually having to add it.
- I want to have a custom attribute to select **users**.
- I want to have a custom attribute to select **teams**.
- I want to have a custom attribute to select **users & teams**.
- I want to have a custom attribute to select **resources** from the Planner.

### Problems

Currently, custom attribute options (select / multi-select) must be created and maintained manually. For large or fast-changing datasets like system users or teams:

Manual setup is tedious and error-prone.

Option lists do not automatically reflect real-time updates (e.g., new user added, name changes).

### Proposal

Leverage the existing backend and add-on support for dynamic enums to let admins source custom attribute dropdowns directly from system entities.

Built-in Supported Dynamic Sourcing:

- Users
- Teams
- Users and Teams

Users

Teams

Users & Teams

## TODO

1. Attribute Settings (Admin Setup)
[ ] Query the backend dynamic enums endpoint to fetch all registered dynamic enum types (built-in + add-ons).

[ ] Update the UI for select and multi-select custom attributes to include a source dropdown:

Custom (Default: manually defined options)

Dynamic Enums (Lists fetched dynamic enum options)

2. Attribute Renderers (Table & Detail Views)
[ ] Update attribute dropdowns in both Table and Detail views to detect when an attribute uses a dynamic enum.

[ ] Call the backend dynamic enum options endpoint to retrieve the latest option list for rendering.

[ ] Ensure option labels automatically reflect backend updates (e.g., user name changes).

Image

filipvnencak:
## Description of changes

An enum attribute ("Department", "Lead", "Priority") holds a list of options that someone types into the attribute editor by hand. When the list changes, someone has to edit it again.

An **enum resolver** is a named list the server builds on request  the studio's users, its teams, a project's statuses, or any list an addon provides. Pick a resolver instead of typing options and the attribute always offers the current list, which can also depend on where it is used: a "Lead" attribute can offer only the people on the project you are looking at.

## Technical details

- Attribute editor: `Enum source` picks a resolver, renders its settings form, and previews the first five options it returns.
- `Debug` dialog resolves the enum live with any context params and settings, so you can see exactly what a resolver returns. Admins can also resolve it as another user.
- Options are fetched once per resolver + settings through RTK Query and shared by columns, filters, slicer and the details panel. Nothing is websocket-invalidated.
- Lazy: a resolver is only called once a visible column, filter or slicer asks for it (`AttributeEnumsProvider`).
- Settings forms (`SimpleForm`) can drive their own selects from a resolver too.

## Additional context

Closes #2243

Summaries showed the avatar url as text for resolver-backed attributes fixed in powerpack: ynput/ayon-power-pack#92

https://github.com/user-attachments/assets/96893cd4-7313-4cc7-8864-cd5ba346f0df

1-attribute-editor-resolver-preview

2-debug-dialog

3-overview-column-dropdown

4-summaries-breakdown-avatars


Innders (commented):
After having a quick play. It would be really nice to see a preview of some of the enum items from the enum you have selected.

Maybe just five enums in a small list under the enum dropdown.

Image

Innders (changes requested):
The preview looks good and nice addition of the "+ x more"

## Settings
- [x] Preview icon does not take into icons that are actually images.

Image

- [x] Add a loading "skeleton" for when the preview enum items are loading. This also prevents the dialog height jumping around as the loading preview height and preview data height should stay the same.

Innders (changes requested):
What is happening with these action images here? The url doesn't seem to be properly resolved for the image, is this a backend issue?

Image

- [x] Add some spacing/gap between the preview items

Image

Innders (changes requested):
Hidden is hidden and we don't care about them at all, they should be completely ignored.
- [x] Do not show number of hidden message

Image

When there are no visible options returned we should return a message like 

"No preview items found. This could be because in context parameters (project_name) are required to get the items."

- [x] Add a message when no items are returned that **always** begins with "No preview items found."
- [x] Add an extra message if `project_name` is a parameter "This could be because in context parameters (project_name) are required to get the items."

kalisp:
> What is happening with these action images here

Resolved here https://github.com/ynput/ayon-backend/pull/1118

Innders (changes requested):
- [x] Add search to the enum resolvers dropdown for over 5 items
- [ ] Add support for in context `user` parameter
- [x] If the enum resolvers has a supported context param, it should be in a callout card at the top. This makes it clear that the results will depend on the context where the enum is used. Only have `project_name` and `user` for now.
- [x] [BONUS] Create an "Enum Playground Dialog". The idea is that you can fully interact with and test the enum resolver. You can see the full list of items, search for the items and set the context (project_name, user). It can be opened by clicking on "+ 25 more" or the context param callout. 
- [x] The simple form doesn't handle multi-select dropdowns properly. `type: "multiselect"`
- [x] When selecting an enum resolver, if the attribute label/name are not currently set, automatically set it to the enum resolver name.
- [x] When loading the table the enum resolver should only be called if there is at least one attribute using the enum visible in the table and visible in the viewport. The enum should be lazily loaded until it is actually needed in the table or filters dropdown.

Innders (changes requested):
- [x] Catch and show when there is an error with an enum

Image

shared/src/components/AttributeEditor/components/EnumSourceField.tsx
```diff
@@ -0,0 +1,315 @@
...
+  }
+
+  const userNote = scopes.includes('user')
+    ? ' User attributes have no project and get the options shown in the preview.'
+    : ' The preview is resolved without a project.'
+  return {
+    variant: 'info',
+    message: `Options depend on the project where the attribute is used.${userNote}`,
+  }
+}
```
Innders: This can be a lot simpler.

"The select options will change based on the project it is used in."
"The select options will change based on the user that is selected."

shared/src/components/AttributeEditor/components/EnumSourceField.tsx
```diff
@@ -0,0 +1,315 @@
...
+              <EnumResolverPreview
+                resolver={selectedResolver.name}
+                acceptedParams={selectedResolver.acceptedParams}
+                settings={settings}
+                onShowMore={openPlayground}
+              />
+              <Button
+                variant="text"
+                icon="science"
+                label="Open playground"
```
Innders:
```suggestion
                label="Debug"
```

Innders (changes requested):
Nice one, a couple extra things and this is going to be so good!

- [x] Simplify the context message to be easier to understand for regular users.

## Playground

This area can be used for better debugging. It would be really useful if it had a couple extra things.

- [x] Rename to "Debug"
- [x] Change the layout to have details on the left and results on the right

The details on the left should have three sections with titles: 

**Context Params** - Shows all context params that are **not**
 in the form. Supported params like `project_name` and `user` show a dropdown to pick an option. Unsupported ones show a note they are unsupported.

**Form** - Shows the form for the enum resolver

**Data** - Shows the raw json data for the enum resolver


Image

shared/src/components/AttributeEditor/components/EnumSourceField.tsx
filipvnencak: Simplified to your wording, one sentence per context param

Innders (changes requested):
- [x] The data needs to show the enum resolver data not the enum items.
- [x] The dialog show be bigger and the left columns should have the same widths.

Image

```json
{
    "name": "enumyProjectPalette",
    "acceptedParams": {
        "project_name": "string",
        "include_neutrals": "boolean",
        "show_hex": "boolean"
    },
    "settingsForm": [
        {
            "type": "boolean",
            "name": "include_neutrals",
            "label": "Include neutral swatches",
            "value": false
        },
        {
            "type": "boolean",
            "name": "show_hex",
            "label": "Show hex codes in descriptions",
            "value": true
        }
    ]
}
```

shared/src/components/AttributeEditor/components/EnumDebugDialog.tsx
```diff
@@ -0,0 +1,308 @@
...
+  flex: 1;
+  min-height: 0;
+`
+
+const Details = styled.div`
+  display: flex;
+  flex-direction: column;
+  gap: var(--base-gap-large);
+  width: 480px;
+  flex: none;
```
Innders, line 35: The details can be bigger
```suggestion
  gap: var(--base-gap-large);
  flex: 0.5;
```

shared/src/components/AttributeEditor/components/EnumDebugDialog.tsx
```diff
@@ -0,0 +1,308 @@
...
+            <SectionTitle>Form</SectionTitle>
+            {settingsFields.length ? (
+              <SimpleForm fields={settingsFields} values={settings} onChange={setFormValues} />
+            ) : (
+              <Message>This resolver has no settings.</Message>
+            )}
+          </Section>
+          <Section>
+            <SectionTitle>Enum Resolver Data</SectionTitle>
+            <RawData>{JSON.stringify(resolver, null, 2)}</RawData>
```
Innders, line 286: This should re-use the json editor from [DetailsDialog.tsx](https://github.com/ynput/ayon-frontend/blob/7d609826bd1703a0b2e94c5c52add0d47f167d27/shared/src/components/DetailsDialog/DetailsDialog.tsx#L16) so that it is properly formatted and have a background.

shared/src/components/AttributeEditor/components/EnumDebugDialog.tsx
```diff
@@ -0,0 +1,308 @@
...
+
+const PARAM_FIELDS: Record<string, FC<ParamFieldProps>> = {
+  project_name: ProjectParamField,
+  user: UserParamField,
+}
+
+const matchesSearch = (item: EnumItem, query: string) =>
+  [item.label, String(item.value), item.description, ...(item.fulltext || [])].some((text) =>
+    text?.toLowerCase().includes(query),
+  )
```
Innders, line 163: Use match-sorter for search matching

shared/src/components/AttributeEditor/components/EnumSourceField.tsx
```diff
@@ -0,0 +1,283 @@
...
+const CUSTOM_ENUM_SOURCE = '__custom__'
+const PREVIEW_LIMIT = 5
+const EMPTY_FIELDS: SimpleFormField[] = []
+const EMPTY_PREVIEW_MESSAGE = 'No preview items found.'
+const SEARCH_THRESHOLD = 5
+
+const CONTEXT_PARAM_MESSAGES: Record<string, string> = {
+  project_name: 'The select options will change based on the project it is used in.',
+  user: 'The select options will change based on the user that is selected.',
+}
```
Innders, line 27: Could you please define the supported context params and then generated a type from them. Then there is single source of truth for the supported context params. For example EnumDebugDialog is also using supported context params and is currently not type safe.

shared/src/components/AttributeEditor/components/EnumSourceField.tsx
```diff
@@ -0,0 +1,283 @@
...
+  cursor: pointer;
+
+  &:hover {
+    color: var(--md-sys-color-on-surface);
+    text-decoration: underline;
+  }
+`
+
+const ClickableInfoMessage = styled(InfoMessage)`
+  cursor: pointer;
```
Innders, line 71: Tighten up the padding and gap as it's currently quite big.


```suggestion
  padding: var(--padding-s) var(--padding-m);

  .content {
    gap: var(--base-gap-small);
  }
```

shared/src/components/SearchFilter/useBuildFilterOptions.tsx
```diff
@@ -23,7 +29,12 @@ import { customRangeOption, generateDatePresetOptions } from './filterDates'
 import { isEmpty, upperFirst } from 'lodash'
 import type { SliceFilter } from '@shared/containers/Slicer/types'
 import { FEATURED_VERSION_TYPES } from '../FeaturedVersionOrder'
+import { useContext } from 'react'
+import { useDispatch } from 'react-redux'
```
Innders, line 33: useDispatch cannot be used inside of components in the shared folder. Either pass it down through props/context or find a different way of doing it.

shared/src/hooks/useAttributeEnumOptions.tsx
```diff
@@ -0,0 +1,305 @@
...
+  try {
+    const result = await request.unwrap()
+    if (result.error) throw new Error(result.error)
+    return normalizeEnumItems(result.items)
+  } catch (error) {
+    throw toEnumRequestError(error)
+  } finally {
+    request.unsubscribe()
+  }
+}
```
Innders, line 61: Can `useLazyGetEnumOptionsQuery` be used here instead??

AI has a tendency to make RTK queries through the api object instead of using the proper query hooks like `enumOptionsQueries.endpoints.getEnumOptions.initiate`. This needs to be checked properly.

src/pages/SettingsPage/UsersSettings/UserAttribForm.jsx
```diff
@@ -124,6 +130,7 @@ const UserAttribForm = ({
 
   return (
     <>
+      {enumSubscriptions}
```
Innders, line 133: What is this, could you please explain it?

shared/src/containers/ProjectTreeTable/buildTreeTableColumns.tsx
```diff
@@ -1416,8 +1421,15 @@ const buildTreeTableColumns = ({
...
               columnId={column.id}
               value={value}
-              attributeData={{ type: attrib.data.type || 'string', widget: attrib.data.widget }}
+              attributeData={{
+                type: attrib.data.type || 'string',
+                widget: attrib.data.widget,
+                enumResolver: attrib.data.enumResolver,
+              }}
               options={attrib.data.enum || []}
+              enableCustomValues={!!attrib.enumIsLoading}
```
Innders, line 1430: I think this should always be false and doesn't have anything to do with enum loading?

shared/src/components/SimpleForm/useResolvedOptions.ts
```diff
@@ -0,0 +1,88 @@
+import { useEffect, useState } from 'react'
+import axios from 'axios'
```
Innders, line 2: This file needs to be cleaned up so that it is not using axios.

Innders (changes requested):
Great work. Some things to change in the Debug Dialog and some code tweaks. Specifically I don't understand the whole thing around `enumSubscriptions` and how that works, feels kind of hacky?

- [ ] Address code comments
- [ ] Improve PR description, explain how enums and enum resolvers work to a normal user who has never heard of them before, my grandma should understand it.
- [ ] Add PR description screenshots.
- [ ] Issue with summaries icons when the icon is an image.

Image

src/pages/SettingsPage/UsersSettings/UserAttribForm.jsx
filipvnencak: That was an element the hook returned that you had to remember to render, it's gone: useResolvedAttributeEnums is now <AttributeEnumsProvider> + useAttributeEnums()

shared/src/containers/ProjectTreeTable/buildTreeTableColumns.tsx
filipvnencak: Can't be plain false: while options are still loading options is empty, so every existing value rendered as a red warning chip. enableCustomValues now only allows custom values on change, new isLoadingOptions suppresses the error styling.

filipvnencak:
Backend bugs found while testing are tracked in ynput/ayon-backend#1125, neither blocks this PR.

Innders (changes requested):
**Request changes**

Please update enum loading to use one simple batch query.

**Architecture**

- Remove the enum provider, context, and subscription components.
- Add one hook that receives the attributes and their context.
- The hook should build all enum requests and call `getEnumOptionsBatch`.
- Pass the project name only to resolvers that accept `project_name`.
- Pass the user name only to resolvers that accept `user`.

**Batch query**

In `getEnums.ts`, `getEnumOptionsBatch.queryFn` should:

- Receive a list of enum requests.
- Create a stable key from the resolver name and its parameters.
- Sort nested parameters so the same request always has the same key.
- Remove duplicate requests.
- Call the existing `getEnumOptions` endpoint with `dispatch(...initiate())`.
- Use `subscribe: false`.
- Pass the `forced` value as `forceRefetch`.
- Return an object where each key is the stable request key.

The batch query should not make its own HTTP request. It should reuse the existing single-resolver query.

**Caching**

There should be two cache levels:

- The batch query caches the complete group of requests.
- The single-resolver query caches each individual resolver request.

This means:

- The same batch uses the batch cache.
- The same resolver request is reused in different batches.
- Duplicate requests result in one HTTP request.
- Changing the project or user creates a new cache key.
- Requests without project or user context can still reuse their existing cache.
- Existing enum tags must continue to invalidate both single and batch results.

shared/src/api/queries/enums/getEnums.ts
```diff
@@ -0,0 +1,165 @@
...
+        // without the registry every param is sent, which is what the app did before
+        const acceptedParams = new Map<string, Record<string, unknown>>()
+        try {
+          const resolvers = await api
+            .dispatch(enumsQueries.endpoints.listEnums.initiate(undefined, { subscribe: false }))
+            .unwrap()
+          resolvers.forEach((resolver) => acceptedParams.set(resolver.name, resolver.acceptedParams))
+        } catch {
+          // keep resolving: a missing registry only costs a wider cache key
+        }
```
Innders, line 122: listEnums query should not be inside of getEnumOptionsBatch. The main problem is that the provided requests arg is not complete and so you need to do extra work to figure out it's params. getEnumOptionsBatch should only be a wrapper around getEnumOptions so that you can requests multiple enum options in one RTK query. Each `request` in `requests` arg should be exactly what is passed to `getEnumOptions` args.

If the requests need to be mutated for the context params, this should be done outside of the query, in the hook.

Innders (changes requested):
There needs to be a small change in `getEnums.ts` otherwise the logic looks good.

Innders (approved):

PR #2289, opened 2026-08-31
Search filter: breadcrumb shows wrong parent group

filipvnencak:
## Description of changes 

Fixes the filter dropdown breadcrumb showing the same field twice, e.g. "Status / Status" instead of "Version / Status".

## Technical details

- Grouped options in `useBuildFilterOptions` carried their own label/icon in `group`, so the breadcrumb read the field name as its parent
- Group label and icon now come from `entityType`, in both `getOptionRoot` and `getAttributeFieldOptionRoot`
- Root group menu, chips and menu row icons are unaffected

## Additional context

Checked on Products (Version/Product groups), Overview (Folder attributes) and Task progress (single scope, no group segment).

### DIFF
--- a/shared/src/components/SearchFilter/useBuildFilterOptions.tsx
+++ b/shared/src/components/SearchFilter/useBuildFilterOptions.tsx
@@ -1049,8 +1049,8 @@ const getOptionRoot = (
 
     rootOption.group = {
       name: groupName,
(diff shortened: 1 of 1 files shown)


BigRoy (approved):
Can't confirm the code - but it works for me ✅

PR #2291, opened 2026-08-31
Access groups: fix saving permissions

### LINKED ISSUE #2290: accessGroups: unable to save access group.
## Story
Can't edit permissions in access groups.

### To reproduce

1. open studio setting
2. open permissions tab
3. try to save changes

### Expected behavior
the permissions should be updating

## Environment

- **1.16.5+202608311350**: 

### Additional context

In `AccessGroupDetail.jsx`, the form calls the API as follows:

```jsx
saveAccessGroup({
  accessGroupName,
  projectName: projectName || "-",
  data: formData,
});
```

However, the generated API client expects a `permissions` field and uses it as the request body:

```js
body: queryArg.permissions
```

As a result, `queryArg.permissions` is `undefined`, so the browser sends the `PUT` request without a body. The backend then returns:

```text
body: field required
```

The fix is a one-line change:

```diff
- data: formData,
+ permissions: formData,
```

filipvnencak:
## Description of changes 

Fixes saving access group permissions, which failed with a server validation error.

## Technical details

- `saveAccessGroup` now gets `permissions` instead of `data` in `AccessGroupDetail.jsx` — codegen renamed the body arg, so the PUT went out with no body
- Project permissions list back on `getAccessGroups({ projectName })`; the studio endpoint hardcodes `isProjectLevel: false`, which left "Clear project overrides" permanently disabled
- Added `providesTags` to `getAccessGroups` so the list refreshes after a save

## Additional context

Closes #2290

### DIFF
--- a/shared/src/api/queries/accessGroups/getAccessGroups.ts
+++ b/shared/src/api/queries/accessGroups/getAccessGroups.ts
@@ -73,12 +73,22 @@ enhancedApi.enhanceEndpoints({
         { type: 'accessGroup', id: 'LIST' },
       ],
     },
+    getAccessGroups: {
+      providesTags: (result) =>
+        result
+          ? [
+              ...result.map(({ name }) => ({ type: 'accessGroup', id: name })),
+              { type: 'accessGroup', id: 'LIST' },
+            ]
+          : [{ type: 'accessGroup', id: 'LIST' }],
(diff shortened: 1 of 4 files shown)


martastain (approved):
works. let's re-release tomorrow

PR #2293, opened 2026-08-31
Slicer: show all lists on every page

filipvnencak:
## Description of changes 

The Lists slicer no longer hides lists that don't match the page entity type. Any list filters any page.

## Technical details

- useEntityListsSlice no longer takes entityTypes. Pages keep their scopes, which still drive the attribute and status slices.
- Cross-entity resolution was already in place (`resolveEntityParents`): version lists resolve to parent folders and tasks, product and task lists to parent folders.
- Task progress page is unaffected, it has no `entityList` slice.


Innders (approved):
LGTM

PR #2300, opened 2026-09-04
Allow remote addon pages to disable global AYON shortcuts

jakubfiala:
## Description of changes 

Remote addon pages can now be configured to disable global AYON keyboard shortcuts (e.g. `b+b`). This can be useful if those shortcuts should be mapped to different things in the addon.

## Technical details

Doing this via a `useEffect` is not 100% optimal but it allows us to keep the shortcuts support on all pages and addons which don't explicitly switch it off. The alternative would be to move the `ShortcutsContext` into each page + addon, but this would require a larger refactor because even some global components such as the project menu use this context.

### DIFF
--- a/shared/src/context/AddonContext.tsx
+++ b/shared/src/context/AddonContext.tsx
@@ -10,6 +10,7 @@ export type RemoteAddon = {
   viewType?: string // if the addon is using views
   slicer?: { fields: string[] }
   projectList?: { multiSelect?: boolean; enabled: boolean }
+  disableGlobalShortcuts?: boolean
 }
 
 export interface RemoteAddonProps {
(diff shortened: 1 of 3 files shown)


Innders:
I'm not sure if disabling all global shortcuts is the best way as it can break non-conflicting shortcuts like `s+s` that users would expect to still work.

What if the global shortcuts context was passed to the addon so that it could decide to disable specific shortcuts it wants to overwrite?

jakubfiala:
@Innders fair point. The question would then be which ones we keep and which ones we disable - this request came from Milan and I think he specifically mentioned `b+b` but I don't know if it's the only one that could cause issues

Innders:
> The question would then be which ones we keep and which ones we disable

This should be up to the addon to decide. This PR should provide that control to the addon.

jakubfiala:
@Innders absolutely

PR #2313, opened 2026-09-10
Search filter: fix custom date range dialog

### LINKED ISSUE #2312: YN-1076:  Date Filter Popup Bug in Products Page

Intermittent bug on the Products page — when a "Created version" date filter is selected and the user then modifies any other filter (e.g. deselects/selects a value and hits Confirm), an unexpected popup appears. 

1. Add some date filter option to the filters
2. Add a new filter or click something in the filter bar
3. The date range selector pops up even if not selecting date filter.

### Additional context:

_This issue was automatically created from AYON ticket [YN-1076]( https://support.ayon.app/projects/Active_Clients/overview?project=Active_Clients&type=folder&id=ed9ea1b0acf411f185d705c26460fd26)

filipvnencak:
## Description of changes 

Fixes the custom date range dialog popping up while you edit an unrelated filter, and makes it edit the date chip you actually clicked.

## Technical details

- Dropped the auto-open block in `SearchFilterWrapper.handleFilterChange`; it ran on every `onChange`.
- `useDateRangeFilter` reads the clicked item's `data-parent` instead of a tracked filter id that goes stale after the query-filter round-trip.
- Both click-capture handlers match date chips by chip id, not label text, which carries a scope prefix on Overview, Lists and Products.

## Additional context

Closes #2312
Ticket: YN-1076

### DIFF
--- a/shared/src/components/SearchFilter/useDateRangeFilter.ts
+++ b/shared/src/components/SearchFilter/useDateRangeFilter.ts
@@ -123,12 +123,8 @@ export const useDateRangeFilter = (): UseDateRangeFilterReturn => {
     const chipEl = target.closest('.search-filter-item')
     if (!chipEl) return
 
-    const labelEl = chipEl.querySelector('.label')
-    const chipLabel = labelEl?.textContent?.replace(/:$/, '').trim()
-    if (!chipLabel) return
-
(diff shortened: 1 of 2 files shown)


Innders (approved):
Fixed it for me. Thanks.

PR #2317, opened 2026-09-15
Lists: Add Copy link button to Share tab

### LINKED ISSUE #1738: Sharing: Copy Link button
## Story

When sharing a list or review session, it would be good to have a "Copy link" button so that the user can easily copy the correct link. Often the correct link is the current URL but the user does not know this can they need some reassuring with the button they have the correct link.

#### Entity Panel
- Add a "Copy link" button in the more menu.
- Folder and Task links always navigate to Overview page.
- Version links always navigate to Product page.

#### Lists and Review Sessions
- Add a "Copy link button to the **Share** tab.
- URL should always be current URL but with the entity query params removed.
- The sharing tab should accept link button prop to be shown.
- The button should use `text` variant so that it does not have a background by default but shows a hover background.

`/projects/wing_it/lists?list=7d6774e6c39b11f0a3d68a82e56e4cd7`
`/projects/wing_it/reviews?review=fb8a1412a5ca11f0926216bcedd01d6a&listTab=access`
`/projects/wing_it/reviews/240e7e90775511f0a4d2726042497f69`

Image

## TODO

- [ ] Add copy link button to lists sharing tab.

filipvnencak:
## Description of changes 

Adds a Copy link button to the list and review session Share tab.

## Technical details

- `ListAccessForm` copies the current URL without the entity params (`uri`, `type`, `project`, `id`, `activity`).
- Passes it as `onCopyLink` to the powerpack `ListAccess` and to `ListAccessFallback`.
- Powerpack shows the button only after its release; older versions ignore the prop.

## Additional context

Related PRs:
- ynput/ayon-frontend-shared-internal#2 – adds `onCopyLink` to `ListAccess`
- ynput/ayon-power-pack#89 – bumps the submodule

Merge order:
1. ayon-frontend-shared-internal#2
2. ayon-power-pack#89 (after re-pointing the submodule to the merged commit on `main`)
3. Release powerpack

This PR can merge at any time. Without the new powerpack the button shows only in the non-powerpack fallback.

Closes #1738

Without Powerpack 
Screenshot 2026-09-15 at 16 29 29

With Powerpack
Screenshot 2026-09-15 at 16 44 30


src/pages/ProjectListsPage/components/ListAccessForm/ListAccessForm.tsx
```diff
@@ -51,6 +52,12 @@ export const ListAccessForm: FC<ListAccessFormProps> = ({
     GuestAccess,
   } = useListsModuleContext()
 
+  const handleCopyLink = () => {
+    const url = new URL(window.location.href)
+    ENTITY_PARAMS.forEach((param) => url.searchParams.delete(param))
+    copyToClipboard(url.toString())
+  }
```
Innders, line 60: Instead of using current url (which is probably okay) it would be better to construct the URL from scratch using projectName, list.id and isReview. This ensures that this button will work when we are only a completely different URL. Plus I think it's clearer to read; as removing the ENTITY_PARAMS is a bit confusing and unclear on the resulting url shape.

Innders: The link on the review tab is actually wrong as it takes you to the review tab when it needs to take you to the review session.

Innders (changes requested):
Button itself looks good.

- [x] fix issues with copy link logic

Innders (approved):
Looks good now thanks

PR #2318, opened 2026-09-16
Inbox: Day dividers, timestamps and tooltips

### LINKED ISSUE #2281: YN-0819: Inbox: QoL Design Improvements

Overview
--------

The current inbox presents a dense "wall of white text" that makes it difficult for supervisors and coordinators to quickly scan, prioritize, and manage heavy daily volumes of production messages. This issue aggregates small, high-impact design tweaks and quality-of-life improvements aimed at enhancing scannability, surfacing vital context, and providing basic bulk management controls.

&nbsp;

[ ] **Add "Clear Read" Action:** Replace or complement the absolute "Clear All" option with a safer "Clear Read" function so users don't accidentally wipe out unread messages requiring action.

&nbsp;

[ ] **Daily Line Breaks:** Add clear visual dividers or section breaks between days to easily separate today's incoming updates from the older backlog.

&nbsp;

[ ] **Timestamp Visibility:** Append the message receive **time** next to the date (e.g., 9 June, 9:15 AM instead of just 9 June). Full timestamp in tooltip.

&nbsp;

[ ] **Icon Tooltips:** Add hover tooltips to the various status and asset icons to clearly state what they mean.

&nbsp;

[ ] **Asset Path Hover:** Show the full path on hover (tooltip) when an asset name is truncated or uses a short prefix

&nbsp;

[ ] **User hover:** Hover over a user to get more information about them, like the teams they are currently in and role if they have one. (Extend UserTooltip)

&nbsp;

[ ] **Category Dots:** Display a small badge next to messages that have an assigned category, with a hover state revealing the category name. This allows critical action items to pop out from passive production updates. 

### Additional context:
link to discussion on [Discord](https://discord.com/channels/517362899170230292/1514206207446093855)
_(might be a private channel)_

_This issue was automatically created from AYON ticket [YN-0819]( https://support.ayon.app/projects/Active_Clients/overview?project=Active_Clients&type=folder&id=c00290b0656211f19f65018279137120)

filipvnencak:
## Description of changes

Makes the inbox easier to scan and clearing more precise.

- Day dividers split the list by date. The cleared tab has its own order, so it gets none.
- Every row shows when the message arrived. Today stays relative ("5 mins"), older rows show date and time, the exact timestamp is in the tooltip.
- Category dots sit next to the message body, stacked when a grouped row covers more than one category. Colours come from the project's activity categories.
- Tooltips on the entity icon (type and subtype) and on the author avatar. The avatar tooltip also lists the user's teams in that project.
- New "Clear read" button next to "Clear all". It clears read messages and leaves the unread ones.
- Clearing now stays inside the tab you are in. Until now the backend cleared every reference the user had in the project, including the other tab and rows that were already cleared.

## Technical details

- Day dividers render between rows, not as `groupedMessages` entries, so `rowIndex` and keyboard navigation are unchanged.
- Clear all and Clear read send an `itemFilter` (`active` + `important`, plus `read` for Clear read) instead of the removed `all` flag, and omit `ids` on that path. Requires ynput/ayon-backend#1122, the two have to merge together.
- Activity categories are fetched once per listed project in `Inbox` and passed down, so a list of 100 rows no longer mounts a query per row.
- The user tooltip queries teams on hover only, after a short delay, and `getTeams` keeps data for 10 minutes.
- `anatomy: false` on the inbox `getProjectsInfo` query drops one request per listed project. Nothing read it.

## Additional context

The path tooltip is dropped. The inbox payload only carries direct parents (task gives the folder, version gives folder and product) and `entityPath` is null for inbox rows, so it could only ever repeat what the cell already shows. A real path needs either a lookup per row or the folder path denormalised into the activity on the backend.

Part of #2281 · YN-0819

01-inbox-list
02-day-dividers-category-dots
03-hover-clear-and-avatar
04-user-tooltip
05-toolbar-clear-read


shared/src/containers/Feed/components/Tooltips/UserTooltip/UserTooltip.tsx
```diff
@@ -1,19 +1,65 @@
...
   pos: {
     top: number
     left: number
   }
 }
 
-const UserTooltip = ({ name, label, pos }: UserTooltipProps) => {
+const UserTooltip = ({ name, label, projectName, pos }: UserTooltipProps) => {
+  // only runs while the tooltip is mounted, so nothing is fetched until a hover
+  const { data: teams = [] } = teamsApi.useGetTeamsQuery(
```
Innders, line 22: Import useGetTeamsQuery properly and then use it like
```suggestion
  const { data: teams = [] } = useGetTeamsQuery(
```

shared/src/containers/Feed/components/Tooltips/UserTooltip/UserTooltip.tsx
```diff
@@ -1,19 +1,65 @@
...
+    const min = half + VIEWPORT_MARGIN
+    setLeft(Math.min(Math.max(pos.left, min), Math.max(min, max)))
+  }, [pos.left, memberships.length])
+
   return (
-    <Styled.Popup style={{ ...pos }}>
+    <Styled.Popup
+      ref={popupRef}
+      style={{ ...pos, left }}
+      className={memberships.length ? 'with-teams' : undefined}
```
Innders, line 49: use clsx
```suggestion
      className={clsx({ 'with-teams': memberships.length > 0 })}
```

src/pages/InboxPage/InboxMessage/InboxCategoryDots.tsx
```diff
@@ -0,0 +1,48 @@
...
+  projectName?: string
+  categories: string[]
+}
+
+const InboxCategoryDots = ({ projectName, categories }: InboxCategoryDotsProps) => {
+  // rendered only for rows that carry a category, so nothing is fetched for a plain inbox
+  const { data: projectCategories = [] } = useGetActivityCategoriesQuery(
+    { projectName: projectName as string },
+    { skip: !projectName },
+  )
```
Innders, line 27: It would probably be okay to have this here but there could be 100s of messages with a category and it will need to run this query every time. Even with the query cached, that's still overhead.

Move the query so that it only runs once in the inbox and then the categories as passed down to this component. This can be done in a hook like "useGetProjectsCategories" that will return only the categories for the projects you need.

src/pages/InboxPage/InboxMessage/InboxCategoryDots.tsx
```diff
@@ -0,0 +1,48 @@
...
+  align-items: center;
+  gap: 4px;
+`
+
+const Dot = styled.span`
+  width: 8px;
+  height: 8px;
+  min-width: 8px;
+  border-radius: 50%;
+`
```
Innders, line 15: When there are multiple categories then the dots should slightly overlap each other, like the stacked users.

src/pages/InboxPage/InboxMessage/InboxMessage.tsx
```diff
@@ -196,6 +283,44 @@ const InboxMessage = ({
...
+    userTooltipTimeout.current = setTimeout(
+      () => setUserTooltipPos({ top, left: left + width / 2 }),
+      USER_TOOLTIP_DELAY,
+    )
+  }
+
+  const handleUserMouseLeave = (): void => {
+    clearTimeout(userTooltipTimeout.current)
+    setUserTooltipPos(null)
+  }
```
Innders, line 312: Can all this code + the User image + the UserTooltip go in a separate component?

src/pages/InboxPage/components/InboxDateDivider.tsx
```diff
@@ -0,0 +1,55 @@
...
+  return dateObj.toDateString()
+}
+
+const getDayLabel = (date: string): string => {
+  const dateObj = new Date(date)
+  if (isToday(dateObj)) return 'Today'
+  if (isYesterday(dateObj)) return 'Yesterday'
+
+  return isThisYear(dateObj) ? dayFormat.format(dateObj) : dayWithYearFormat.format(dateObj)
+}
```
Innders, line 27: It's worth using date-fns `format` here to make it simpler and easier to read.
```suggestion
const getDayLabel = (date: string): string => {
  const dateObj = new Date(date)
  if (isToday(dateObj)) return 'Today'
  if (isYesterday(dateObj)) return 'Yesterday'

  return format(dateObj, isThisYear(dateObj) ? 'd MMMM' : 'd MMMM yyyy')
}
```

Innders (changes requested):
Some nice changes!

- [ ] Issue with user image and Clear button when hovering (they swamp) 

Image

- [ ] Path tooltip seems to be exactly the same as the path? If possible it should be fulll path, including the project name.

Image

- [ ] "Clear read" clears everything (is there BE branch needed??)

### PR Descriptions Improvments

- [ ] Expand on this line with more detail "Makes the inbox easier to scan: day dividers, message times, category dots, and hover tooltips for icons, paths and users."
- [ ] Add screenshot

Innders (changes requested):
Thanks for the changes and PR description update!

- [ ] Hovering causes the user image to appear after the button. Keep the same behaviour we already have in dev.

**Current (wrong)**

Image

**Dev (right)**

Image

- [ ] When multiple inbox items are selected, clicking the inline action should apply it to all items.

Image

- [ ] When multiple inbox items are selected, the Clear button should only appear when hovering and have label "Clear selection". 

Image

Innders (approved):

PR #2323, opened 2026-09-17
Entity picker: reviewables-only switch and task names

### LINKED ISSUE #2322: YN-1090: Picker: Add option to check switch to hide none reviewables

Sometimes you only care about finding products and versions that have reviewable content between departments.

&nbsp;

Show Reviewables
----------------

&nbsp;

Add a prop that when true, shows a switch at the top of the dialog called "Show reviewables only". By default this switch is turned on. It should hide products and versions that have "hasReviewables: false". Turning the switch off shows everything.

&nbsp;

Product Featured Version Task
-----------------------------

&nbsp;

Add a prop that will show:

- The task name of the product's featured version on the product
- The task name of the version on the version.

&nbsp;

Both props are default false and will be used in the review addon. 

### Additional context:

_This issue was automatically created from AYON ticket [YN-1090]( https://support.ayon.app/projects/Active_Clients/overview?project=Active_Clients&type=folder&id=cd59b9c0b1e611f1820e4dc3b428ba1b)

filipvnencak:
## Description of changes 

Adds opt-in props to hide products and versions without reviewables and to show task names.

## Technical details

- `showReviewablesSwitch` sends `hasReviewables` to the products and versions queries (`EntityPickerDialog.tsx`).
- A selected id is kept or dropped by a dedicated `getReviewableEntityIds` query, so searching and paging can't deselect a valid row (`getLinks.ts`).
- `showTaskNames` fetches `task { name }` via `@include` and shows it as the row badge (`buildEntityPickerTableData.ts`).

## Additional context

Closes #2322 · YN-1090

picker-reviewables-off
picker-reviewables-on


Innders (changes requested):
## UI

- [ ] Add a gap between the switch and table
- [ ] Move the switch to the right

## Code

I think the code is overly complicated now with those extra queries just to see if a version/product has a reviewable.

Please remove the dedicated getReviewableEntityIds queries.

Treat the currently loaded picker results as the source of truth for selection visibility.

When a search, pagination update, or “Show reviewables only” toggle causes a selected entity to disappear from the loaded results, remove that ID from rowSelection. 

This prevents hidden selections from remaining active in the background and means the submit action only uses visible selections.

Innders (changes requested):
That's a lot better, thanks!

After playing with it more, it would useful to have the switch when replacing a version in a list, but it should be default off.

- [ ] show reviewables switch in default off position. (visibility and default value should be separated)
- [ ] show the reviewables switch (default off) when replacing a version in a list.

Image

Innders (approved):

